Issue #2, opened 2022-03-28
Create provenance generation library


laurentsimon:
Provenance generation can be shared across ecosystems, except for some inputs like `buildConfig` and `parameters` that should be provided by ecosystem-specific builders.

We need to refactor the code we use for https://github.com/slsa-framework/slsa-github-generator-go to use code in this repo (library or CLI)

cc @[ianlewis](https://github.com/ianlewis)

laurentsimon:
I think this issue can be closed, right?

ianlewis:
I'll close it when I merge #47

PR #3, opened 2022-03-29
(WIP) Initial API for Github provenance generation.

ianlewis:
See https://github.com/slsa-framework/slsa-github-generator/issues/2


slsa/provenance.go
```diff
@@ -0,0 +1,128 @@
...
+	"fmt"
+
+	intoto "github.com/in-toto/in-toto-golang/in_toto"
+	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
+
+	"github.com/slsa-framework/slsa-github-generator/github"
+)
+
+// GithubHostedActionsBuilderID is the builder ID for Github hosted actions.
+const GithubHostedActionsBuilderID = "https://github.com/Attestations/GitHubHostedActions@v1"
```
laurentsimon, line 13: This will be set dynamically as per https://github.com/slsa-framework/slsa-github-generator-go/pull/17

cmd/slsa-github-generator/generate.go
```diff
@@ -0,0 +1,57 @@
...
+a Github Actions workflow.`,
+
+		Run: func(cmd *cobra.Command, args []string) {
+			ghContext, err := github.GetWorkflowContext()
+			check(err)
+
+			p, err := slsa.HostedActionsProvenance(slsa.WorkflowRun{
+				// TODO: Get subject names and digests
+				Subjects:      []intoto.Subject{},
+				BuildType:     "",
```
laurentsimon, line 33: @MarkLodato suggested we use https://github.com/slsa-framework/slsa-github-generator-go/blob/main/pkg/provenance.go#L135. I think this needs set be set dynamically for each ecosystem-specific builder.

laurentsimon (commented):
Thanks! 

I like the breakdown of functionalities.

In case it's useful, here's one design pattern we use in scorecard to share the same arguments for API and CLI https://github.com/ossf/scorecard/blob/main/cmd/root.go#L47 and https://github.com/ossf/scorecard/blob/main/options/options.go

Maybe not be needed, but pointing it out in case you find it useful.

slsa/provenance.go
ianlewis: My understanding was that the builder ID identifies the builder at the granularity of the build system (i.e. Github hosted runners, Github self-hosted, Cloud Build hosted, Cloud build custom workers etc) and not necessarily identify the reusable workflow that was used.

If we use a URI for the reusable workflow, how might one differentiate between hosted versus self-hosted Github Actions for example? I can imagine scenarios where folks may want to verify the builder id to make sure that builds are run from their self hosted runners.

cmd/slsa-github-generator/generate.go
ianlewis: Yeah, I wasn't sure what to do with this yet but it looks like we'll need it even if we don't set a buildConfig.

slsa/provenance.go
ianlewis: Does it not make sense for the language-specific reusable workflows to set a URI for the workflow into the invocation instead? that feels more like the right place.

ianlewis:
> Thanks!
> 
> I like the breakdown of functionalities.
> 
> In case it's useful, here's one design pattern we use in scorecard to share the same arguments for API and CLI https://github.com/ossf/scorecard/blob/main/cmd/root.go#L47 and https://github.com/ossf/scorecard/blob/main/options/options.go
> 
> Maybe not be needed, but pointing it out in case you find it useful.

Yeah, cosign and friends do this too. I may get there eventually. Thanks for pointing it out :)

cmd/slsa-github-generator/generate.go
ianlewis: Added a similar build type

cmd/slsa-github-generator/attest.go
```diff
@@ -0,0 +1,105 @@
...
+
+				_, err = f.Write(att.Bytes())
+				check(err)
+
+			}
+		},
+	}
+
+	c.Flags().StringVarP(&attPath, "signature", "g", "attestation.intoto.jsonl", "Path to write the signed attestation")
+	c.Flags().StringVarP(&subjects, "subjects", "s", "", "Comma separated list of subjects of the form NAME@SHA256")
```
ianlewis, line 102: It seems like artifact names can have characters like '@' and ',' so should use a character that can't be used.
https://github.com/actions/upload-artifact/issues/22

Maybe something like `name1:sha|name2:sha` ?

slsa/provenance.go
laurentsimon: > Does it not make sense for the language-specific reusable workflows to set a URI for the workflow into the invocation instead? that feels more like the right place.

definitely makes sense for the reusable workflow to set it, possibly via an API/CLI option?

cmd/slsa-github-generator/attest.go
laurentsimon: SGTM.

slsa/provenance.go
laurentsimon: > My understanding was that the builder ID identifies the builder at the granularity of the build system (i.e. Github hosted runners, Github self-hosted, Cloud Build hosted, Cloud build custom workers etc) and not necessarily identify the reusable workflow that was used.
> 
> If we use a URI for the reusable workflow, how might one differentiate between hosted versus self-hosted Github Actions for example? I can imagine scenarios where folks may want to verify the builder id to make sure that builds are run from their self hosted runners.

The idea is to have a mapping between reusable workflows and whether they use self-hosted vs GH-hosted runners, but I see your point. 

@MarkLodato further thoughts?

slsa/provenance.go
```diff
@@ -0,0 +1,123 @@
...
+	// GithubContext is the context for the workflow run.
+	GithubContext github.WorkflowContext
+}
+
+var (
+	parametersVersion = 1
+)
+
+// WorkflowParametersV1 contains parameters given to the workflow invocation in v1 format.
+type WorkflowParametersV1 struct {
```
MarkLodato, line 37: My inclination is to remove all of these fields except for `event.input` in the case of workflow_dispatch (or any others that are wide-open like that). The main purpose of this field is to enable the "parameterless" check of SLSA 4, which exists to prevent an actor from injecting unintended behavior into the build. I worry that adding all this here will just be noise that is impractical to validate, thus it will not add any value.

To be clear, it is an open question of how exactly to interpret this (https://github.com/slsa-framework/slsa/issues/278). My opinion is that we can safely omit anything that is unlikely to lead to an injection vulnerability. So in this case:

* Enums like `event_name` and `ref_type` only have fixed values, so they can't really be used to inject behavior except in contrived cases.
* Fields with a limited character set, such as `sha` and `actor`, are also unlikely to lead to a vulnerability.
* Fields with a more open character set, such as `head_ref`, _could_ potentially be used for shell injection attacks (e.g. `refs/heads/$(echo evil)`), but my inclination is to still omit it because validating the branch is basically impossible. One potential way to mitigate this is for this reusable workflow to only allow a safe character set.
* The full event payload could potentially have something in it that can lead to injection, but my guess is that in practice almost no builds are going to use those fields in such as way.

If desired, perhaps these can move to `environment` instead.

What do you think?

slsa/provenance.go
```diff
@@ -0,0 +1,123 @@
...
+					EntryPoint: w.GithubContext.Workflow,
+					URI:        w.GithubContext.RepositoryURI(),
+					Digest: slsa.DigestSet{
+						"sha1": w.GithubContext.SHA,
+					},
+				},
+				// Non user-controllable environment vars needed to reproduce the build.
+				Environment: map[string]interface{}{
+					// NOTE: Hosted runners are always amd64.
+					"arch":               runtime.GOARCH,
```
MarkLodato, line 93: nit: I would omit this since it's misleading - it tells us the architecture of the provenance generator, which may differ from the architecture that the actual build job took place on.

.github/workflows/provenance.yml
```diff
@@ -0,0 +1,152 @@
...
+      go-generator-sha256: ${{ steps.generator-gen.outputs.go-generator-sha256 }}
+      go-prov-sha256: ${{ steps.generator-gen.outputs.go-prov-sha256 }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.3.4
+        with:
+          fetch-depth: 0
+          repository: "${{ env.GENERATOR_REPOSITORY }}"
+          ref: "${{ env.GENERATOR_REF }}"
```
MarkLodato, line 59: Shouldn't this check out the exact version from the reusable workflow? For example, if someone calls `uses: slsa-framework/slsa-github-generator@abcd1234...`, shouldn't this check out `abcd1234...`?

But sadly I can't find a way to get it via the context. Any ideas?

.github/workflows/provenance.yml
```diff
@@ -0,0 +1,152 @@
...
+        description: "The artifact name of the signed provenance"
+        value: ${{ jobs.sign-prov.outputs.prov-name }}
+
+jobs:
+  ###################################################################
+  #                                                                 #
+  #                       Build the generator                       #
+  #                                                                 #
+  ###################################################################
+  generator:
```
MarkLodato, line 48: nit: Could you explain in a comment why we have two separate jobs, instead of just the build + run in a single job? The latter would avoid the need for upload/download artifact and the hash check. (I ask because I could easily see someone refactoring to merge into a single one, so if there's a reason to keep them separate, that should be documented.)

.github/workflows/provenance.yml
```diff
@@ -0,0 +1,152 @@
...
+
+      - name: Build slsa-github-generator
+        shell: bash
+        id: generator-gen
+        run: |
+          set -euo pipefail
+
+          # https://go.dev/ref/mod#build-commands.
+          go build -mod=vendor -o slsa-github-generator github.com/slsa-framework/slsa-github-generator/cmd/slsa-github-generator
+          GENERATOR_DIGEST=$(sha256sum slsa-github-generator | awk '{print $1}')
```
MarkLodato, line 83: Could you add a comment explaining why you do this hash check, i.e. to prevent tampering because we trust the output variable but not artifacts?

.github/workflows/provenance.yml
```diff
@@ -0,0 +1,152 @@
...
+          echo "$GENERATOR_HASH slsa-github-generator" | sha256sum --strict --check --status || exit -2
+
+          # Make the binary executable.
+          chmod a+x slsa-github-generator
+
+      - name: Create and sign provenance
+        id: sign-prov
+        shell: bash
+        env:
+          SUBJECTS: "${{ inputs.artifacts }}"
```
MarkLodato, line 133: Not needed? isn't this already passed in via `--subjects`?

cmd/slsa-github-generator/attest.go
```diff
@@ -0,0 +1,107 @@
...
+			return nil, errors.New("missing subject name")
+		}
+		if len(parts) == 1 {
+			return nil, fmt.Errorf("expected sha256 hash for subject %q", parts[0])
+		}
+
+		parsed = append(parsed, intoto.Subject{
+			Name: parts[0],
+			Digest: slsav02.DigestSet{
+				"sha256": parts[1],
```
MarkLodato, line 38: suggestion: do a sanity check to verify that it looks like a sha256 hash (`^[a-fA-F0-9]{64}$`), since that could be a common mistake.

Also force it to lowercase, as required by the [spec](https://github.com/in-toto/attestation/blob/main/spec/field_types.md#DigestSet).

slsa/provenance.go
```diff
@@ -0,0 +1,123 @@
...
+type WorkflowRun struct {
+	// Subjects is information about the generated artifacts.
+	Subjects []intoto.Subject
+
+	// BuildType indicates the type of build that was done. More importantly it
+	// also specifies the format of the BuildConfig.
+	BuildType string
+
+	// BuildConfig is metadata about the build.
+	BuildConfig interface{}
```
MarkLodato, line 26: nit:  Since this is always nil, why not omit this?

And if you do that, then you can hardcode BuildType within provenance.go, which further simplifies the interface.

MarkLodato (changes requested):

.github/workflows/provenance.yml
laurentsimon: I don't think it's available in the context and we need an OIDC request to get it, I think (That's how we're setting the `buildID` in the Go generator)

MarkLodato: Right, it's not in the context, but I do think it's an important problem to solve. For one, always checking out at head means that we can't properly test it without submitting.

Could a prior step pull out the repository and ref from the OIDC token?

laurentsimon: We should use OIDC in a different job, and use the results as input to the checkout - since workflow steps are not isolated from each others

.github/workflows/provenance.yml
laurentsimon: The calling workflow can tamper with the artifact that's uploaded by uploading a different binary

.github/workflows/provenance.yml
laurentsimon: For the Go builder, we needed to separate the "build of the builder" from the "build of the project". In provenance-only workflows, this does not seem necessary and could be removed.

.github/workflows/provenance.yml
laurentsimon: the input to the `--subject` is set by this `$SUBJECT` env variable. Maybe it's misleading, but the code does:
1. Set a `SUBJECTS` env variable for the script, with the value being the input `artifacts` passed to the reusable workflow
2. Run `... -- subjects $SUBJECT` to pass it as argument to the CLI

Does this clarify?

slsa/provenance.go
laurentsimon: Thanks for the feedback, Mark. This code is taken straight from the Go PoC, so let's make sure we agree on what to put inside it:

> My inclination is to remove all of these fields except for `event.input` in the case of workflow_dispatch (or any others that are wide-open like that). The main purpose of this field is to enable the "parameterless" check of SLSA 4, which exists to prevent an actor from injecting unintended behavior into the build. 

We can remove these and move them to `environment` if needed. I was hoping we would not have to handle special cases like `workflow_dispatch` in the generator logic to keep the code simple. My original reasoning for this was that by keeping the generator "dumb" (i.e., simply copying GitHub context), the logic can be moved to the verifier only. This way the verifier can be updated, new features added, new policies added, and still be used on binaries generated with an older generator's version. The more logic we add in the generator, the harder it will become to keep generator/verifier in sync and use verifier on builds generated with older generators... if that makes sense.

> 
> To be clear, it is an open question of how exactly to interpret this ([slsa-framework/slsa#278](https://github.com/slsa-framework/slsa/issues/278)). My opinion is that we can safely omit anything that is unlikely to lead to an injection vulnerability. So in this case:
> 
> * Enums like `event_name` and `ref_type` only have fixed values, so they can't really be used to inject behavior except in contrived cases.
> * Fields with a limited character set, such as `sha` and `actor`, are also unlikely to lead to a vulnerability.
> * Fields with a more open character set, such as `head_ref`, _could_ potentially be used for shell injection attacks (e.g. `refs/heads/$(echo evil)`), but my inclination is to still omit it because validating the branch is basically impossible. One potential way to mitigate this is for this reusable workflow to only allow a safe character set.
> * The full event payload could potentially have something in it that can lead to injection, but my guess is that in practice almost no builds are going to use those fields in such as way.
> 
> If desired, perhaps these can move to `environment` instead.
> 
> What do you think?

.github/workflows/provenance.yml
MarkLodato: Oh, I see. But why not use `${{ toJSON(github) }}` directly in the command? Is it due to shell quoting, such that using `env` avoids the issues? If so, that would be good to comment since it's not obvious to me as a reader.

laurentsimon: it's good practice to set these via env variables to avoid script injection. The way these scrips work is that GitHub replaces the content with the variable and create the script on disk with the result. So even escaping the variable as `"$UNTRUSTED_VARIABLE"` does not prevent injection. The only way to escape is to:
1. Use the env variable
2. Use `"$VAR"` as well

See https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections for more details. Fyi, scorecard has a check called [`Danegrous`](https://github.com/ossf/scorecard/blob/main/docs/checks.md#dangerous-workflow) to catch these too.

Let's maybe add a comment and a link to the GitHub doc?

slsa/provenance.go
laurentsimon: I think this is a placeholder and the caller (ecosystem reusable workflow) will pass data of their choosing here. This way we'll be able to remove the provenance generator across ecosystem generators.

slsa/provenance.go
ianlewis: Yeah, perhaps this is better taken up elsewhere. Reading the docs again, I see that you're right that since we are signing the provenance in the reusable workflow ourselves (and not Github Actions itself) that the builder id should reflect the workflow's trust base, but more examples would make it easier to understand.

cmd/slsa-github-generator/generate.go
ianlewis: The API allows you to specify it already in the WorkflowRun but I'll add a CLI option also.

.github/workflows/provenance.yml
ianlewis: Right, the intention was never to keep this as it is forever. Using the OIDC token gives the ref that the caller workflow used in their reference to the reusable workflow yaml I suppose?

slsa/provenance.go
ianlewis: The intention is also to allow folks to use these packages as a simple library for ecosystem/language-specific builders so allowing them to pass their own BuildConfig is desirable. BuildType needs to be configurable for the same reason.

cmd/slsa-github-generator/provenance.go
```diff
@@ -0,0 +1,58 @@
...
+	"errors"
+
+	"github.com/spf13/cobra"
+
+	"github.com/slsa-framework/slsa-github-generator/github"
+	"github.com/slsa-framework/slsa-github-generator/slsa"
+)
+
+// provenanceCmd returns the 'provenance' command.
+func provenanceCmd() *cobra.Command {
```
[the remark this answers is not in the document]
ianlewis: It's was a command I thought might be useful to generate provenance without going though the full signing process for those who want to test or play around with the project.

I can remove it if it's truly not necessary but I was inclined to add it as a separate command since it didn't really fit well as an option on the attest command.

ianlewis:
BTW, @laurentsimon are we using "The GOSST team" for the copyright headers in these projects?

.github/workflows/provenance.yml
laurentsimon: correct

.github/workflows/provenance.yml
laurentsimon: the slsa-framework is under the OpenSSF. Here's a discussion https://github.com/ossf/scorecard/issues/1534 on handling dates from folks at Linux Foundation, k8 and OpenSSF.

Note that I'm not against dates, but if they aren't truly needed I don't mind avoiding them.

slsa/provenance.go
ianlewis: Added code to generate the builder ID from the OIDC token

slsa/provenance.go
ianlewis: I don't think that will address mark's comment which is that the calling job could technically be a different architecture than the job running the provenance generation.

The only way I see that we can do that is to add it as an input.

I added issue #5 and we can have further conversation there. I removed the "arch" key for now.

.github/workflows/provenance.yml
ianlewis: I plan to do this in a follow up PR. Issue #4 will track this.

.github/workflows/provenance.yml
ianlewis: Added a better comment where the hash is checked.

slsa/provenance.go
ianlewis: I agree we should omit anything that wouldn't be considered a "parameter" from the perspective of "parameterless build" SLSA requirement. You should be able to check compliance with that by simply checking that parameters is empty.

"Includes all build parameters" requirement describes them as "build parameters under a user’s control" and I think most of these would not be considered under the "user's control".

note, The field we should include is `event.inputs` for clarity.

laurentsimon: > I agree we should omit anything that wouldn't be considered a "parameter" from the perspective of "parameterless build" SLSA requirement. You should be able to check compliance with that by simply checking that parameters is empty.

makes sense to me too.

> "Includes all build parameters" requirement describes them as "build parameters under a user’s control" and I think most of these would not be considered under the "user's control".
> 
> note, The field we should include is `event.inputs` for clarity.

So we'll only report `event.inputs` in the `invocation.parameters`, and move the rest to `environment`?

.github/workflows/provenance.yml
laurentsimon: How about `SLSA Authors` in the copyright?
@ianlewis feel free to keep the date in the copyright, we can re-visit this later. No need to block this PR on this.

slsa/provenance.go
laurentsimon: SG, thanks.

slsa/provenance.go
```diff
@@ -0,0 +1,128 @@
...
+	// Actor is the username of the user that initiated the workflow run.
+	Actor string `json:"actor,omitempty"`
+
+	// SHA1 is the commit SHA that triggered the workflow run.
+	SHA1 string `json:"sha1,omitempty"`
+}
+
+var (
+	parametersVersion = 1
+	audience          = "slsa-framework"
```
laurentsimon, line 66: let's make the scope even more specific. Can do in a follow-up PR once https://github.com/slsa-framework/slsa-github-generator-go/pull/25 is landed

slsa/provenance.go
MarkLodato: > My original reasoning for this was that by keeping the generator "dumb" (i.e., simply copying GitHub context), the logic can be moved to the verifier only. This way the verifier can be updated, new features added, new policies added, and still be used on binaries generated with an older generator's version. The more logic we add in the generator, the harder it will become to keep generator/verifier in sync and use verifier on builds generated with older generators... if that makes sense.

Hmm, that argument definitely has merit.

That said, my main worry is about the which party has the knowledge and control to ensure that the builds are parameterless. If the verifier is responsible for this (i.e. generator "dumb"), the verifier may not have complete knowledge or ability to actually do this, and the builder/generator may make changes over time that invalidate assumptions. But if the generator is responsible, then it becomes a bit more clear that they are fulfilling a contract of having no other parameters besides these.

slsa/provenance.go
```diff
@@ -0,0 +1,128 @@
...
+			Type:          intoto.StatementInTotoV01,
+			PredicateType: slsa.PredicateSLSAProvenance,
+			Subject:       w.Subjects,
+		},
+		Predicate: slsa.ProvenancePredicate{
+			BuildType: w.BuildType,
+			Builder: slsa.ProvenanceBuilder{
+				ID: fmt.Sprintf("https://github.com/%s", t.JobWorkflowRef),
+			},
+			Invocation: slsa.ProvenanceInvocation{
```
MarkLodato, line 88: Can you also fill out the `metadata` field? https://slsa.dev/provenance/v0.2#metadata

At a minimum, `metadata.completeness.parameters: true`.

MarkLodato (changes requested):

slsa/provenance.go
laurentsimon: That makes sense too. 

Since there seems to be a consensus that only `event.inputs` are deemed actual parameters, I suppose it's fine to move everything else (payload name, payload, etc) into the `invocation.environment` then?

This would mean that the generator is "dumb" for the information it copies into `invocation.environment`, so I think this alleviates my original concern.

.github/workflows/provenance.yml
ianlewis: SGTM

cmd/slsa-github-generator/provenance.go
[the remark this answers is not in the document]
ianlewis: Yeah, it is maybe confusing. I think we can just remove this and give some extra instructions for debugging. 

It's a bit difficult to set up and run because you need a github context and OIDC provider so it's only really practical to run if running in a github action, but you have to kind of fake the github context for this command anyway so it's probably not really worth it.

I'll track that in #6

slsa/provenance.go
ianlewis: #7

slsa/provenance.go
ianlewis: I'll add `metadata.completeness.parameters` and follow up on other fields in another PR

Added issue #8

slsa/provenance.go
ianlewis: SGTM

ianlewis: I updated the environment though I'm still not super happy with it. I'm thinking now that if we want the environment to be as complete as possible that we'll want to just include the whole github context in there rather than specific fields like we've added in the parameters. Though some things like maybe the `action` field in the `github` context are perhaps not relevant.

That said, it seems that `environment` is only supposed to be what's necessary to reproduce the build and nothing more so I'm of two minds. The idea of keeping the generator "dumb" does have merit I think, but that seems to go against how many of the fields are described in the spec.

cmd/slsa-github-generator/provenance.go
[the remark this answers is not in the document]
ianlewis: Removed.

slsa/provenance.go
ianlewis: Added `metadata.completeness.parameters'

.github/workflows/provenance.yml
ianlewis: Added license headers and merged the jobs into one.

slsa/provenance.go
ianlewis: I'm going to resolve this for now, but feel free to add new comments on the environment if you have more concerns.

ianlewis:
Made a couple of new minor(ish) API decisions
- Remove the LogEntry return from Upload so that the signing API is not cosign/rekor specific and will allow us to add interfaces and perhaps support other signers
- Added fields determined by buildType to the WorkflowRun so that users of the API can override SLSA fields that are dependent on buildType if they choose to customize it.

laurentsimon:
> Made a couple of new (hopefully) minor API decisions
> 
> * Remove the LogEntry return from Upload so that the signing API is not cosign/rekor specific and will allow us to add interfaces and perhaps support other signers
> * Added fields determined by buildType to the WorkflowRun so that users of the API can override SLSA fields that are dependent on buildType if they choose to customize it.

"override" means there will be defaults, what will they be?

.github/workflows/provenance.yml
```diff
@@ -0,0 +1,110 @@
...
+jobs:
+  ###################################################################
+  #                                                                 #
+  #                       Build the generator                       #
+  #                                                                 #
+  ###################################################################
+  generator:
+    outputs:
+      go-generator-sha256: ${{ steps.generator-gen.outputs.go-generator-sha256 }}
+      go-prov-sha256: ${{ steps.generator-gen.outputs.go-prov-sha256 }}
```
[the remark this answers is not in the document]
ianlewis: Yeah, this should be removed. I thought I had done that.

ianlewis:
> > Made a couple of new (hopefully) minor API decisions
> > 
> > * Remove the LogEntry return from Upload so that the signing API is not cosign/rekor specific and will allow us to add interfaces and perhaps support other signers
> > * Added fields determined by buildType to the WorkflowRun so that users of the API can override SLSA fields that are dependent on buildType if they choose to customize it.
> 
> "override" means there will be defaults, what will they be?

Override was maybe the wrong word. "Defaults" are set by NewWorkflowRun which creates a generic WorkflowRun with a proveance-only buildType (`https://github.com/slsa-framework/slsa-github-generator@v1`). It basically defines the buildType and the dependent fields. Language specific builders would probably want to set their own buildType and either modify fields on the returned WorkflowRun from NewWorkflowRun, create their own version of the function, or otherwise create the WorkflowRun themselves since they would be specifying the buildType and the fields that are determined by the buildType (invocation, buildConfig, materials) because they have more specialized knowledge about the build. 
`HostedActionsProvenance` won't have default values.

https://github.com/slsa-framework/slsa-github-generator/pull/3/files#diff-e8bae1665df5da39dc685be0a7688695ceb442878e6b76345ff46bb34924e3b3R62

It would be nice to have better extension points for other builders to set their own buildType and override specific bits like `environment`, or `materials` but I feels like it's already complicated enough.

At least, that was my thinking.

.github/workflows/provenance.yml
```diff
@@ -0,0 +1,110 @@
...
+
+###################################################################
+#                                                                 #
+#            Input and output argument definitions                #
+#                                                                 #
+###################################################################
+on:
+  workflow_call:
+    inputs:
+      artifacts:
```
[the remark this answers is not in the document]
ianlewis: Yeah, originally my intention was to have users upload the artifacts and put the names here but it's not really required since we don't actually use the artifacts.

`subject` or `subjects` might be a better name here.

.github/workflows/provenance.yml
```diff
@@ -0,0 +1,110 @@
...
+# limitations under the License.
+
+name: SLSA provenance generator
+
+permissions:
+  contents: read
+
+env:
+  # This repo.
+  GENERATOR_REPOSITORY: ianlewis/slsa-github-generator
```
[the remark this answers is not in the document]
ianlewis: Ack. Bad merge. will fix.

slsa/provenance.go
```diff
@@ -0,0 +1,65 @@
...
+const (
+	// GithubHostedActionsBuilderID is the builder ID for Github hosted actions.
+	GithubHostedActionsBuilderID = "https://github.com/Attestations/GitHubHostedActions@v1"
+	audience                     = "slsa-framework"
+)
+
+// HostedActionsProvenance generates an in-toto provenance statement in the SLSA
+// v0.2 format for a workflow run on a Github actions hosted runner.
+func HostedActionsProvenance(w WorkflowRun) (*intoto.ProvenanceStatement, error) {
+	t, err := github.RequestOIDCToken(audience)
```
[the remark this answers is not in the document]
ianlewis: Is it ok if I add verification in a follow up PR? I think this is necessary to add but I think I'd like to focus on the golang API and github workflow here in order to scope the PR.

Created issue #9 to track.

.github/workflows/provenance.yml
```diff
@@ -0,0 +1,110 @@
...
+  workflow_call:
+    inputs:
+      artifacts:
+        description: "Formatted list of artifacts of the form NAME:SHA256[|NAME:SHA256[|...]]"
+        required: true
+        type: string
+    outputs:
+      attestation-name:
+        description: "The artifact name of the signed provenance"
+        value: ${{ jobs.sign-prov.outputs.prov-name }}
```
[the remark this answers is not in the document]
ianlewis: Yes, Will fix.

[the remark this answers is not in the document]
ianlewis: This actually needs a bit more since I think I'll need to define the job output for `generator` and assign that the output of the `sign-prov` step. This is a remnant from removing the `sign-prov` job that was in the original slsa-go repo.

laurentsimon:
> slsa-github-generator



> 



> > > Made a couple of new (hopefully) minor API decisions
> > > 
> > > * Remove the LogEntry return from Upload so that the signing API is not cosign/rekor specific and will allow us to add interfaces and perhaps support other signers
> > > * Added fields determined by buildType to the WorkflowRun so that users of the API can override SLSA fields that are dependent on buildType if they choose to customize it.
> > 
> > 
> > "override" means there will be defaults, what will they be?
> 
> Override was maybe the wrong word. "Defaults" are set by NewWorkflowRun which creates a generic WorkflowRun with a proveance-only buildType (`https://github.com/slsa-framework/slsa-github-generator@v1`). It basically defines the buildType and the dependent fields. Language specific builders would probably want to set their own buildType and either modify fields on the returned WorkflowRun from NewWorkflowRun, create their own version of the function, or otherwise create the WorkflowRun themselves since they would be specifying the buildType and the format of fields that are determined by the buildType (invocation, buildConfig, materials). That is because they have more specialized knowledge about the build. `HostedActionsProvenance` won't have default values.
> 
> https://github.com/slsa-framework/slsa-github-generator/pull/3/files#diff-e8bae1665df5da39dc685be0a7688695ceb442878e6b76345ff46bb34924e3b3R62
> 
> It would be nice to have better extension points for other builders to set their own buildType and override specific bits like `environment`, or `materials` but I feels like it's already complicated enough.
> 
> At least, that was my thinking.
> 
> The simplest way to use it would be something like:
> 
> ```go
> r := slsa.NewWorkflowRun(subjects, githubContext)
> r.BuildType = golangSpecificBuildType
> r.BuildConfig = /* custom Go builder config (steps etc.) */
> p, _ := slsa.HostedActionsProvenance(r)
> /* sign and upload */
> ```

Let's start with this, it's simple and is flexible to test in various scenarios/use cases.

slsa/provenance.go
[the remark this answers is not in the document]
laurentsimon: +1, let's focus on API in this PR and address verification in a follow-up PR.

.github/workflows/provenance.yml
```diff
@@ -0,0 +1,110 @@
...
+###################################################################
+#                                                                 #
+#            Input and output argument definitions                #
+#                                                                 #
+###################################################################
+on:
+  workflow_call:
+    inputs:
+      artifacts:
+        description: "Formatted list of artifacts of the form NAME:SHA256[|NAME:SHA256[|...]]"
```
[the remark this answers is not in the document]
ianlewis: Unfortunately, github actions set-output truncates multi-line output unless you escape it which seems like an annoying thing that anyone wanting to specify multiple artifacts will run into.
https://github.community/t/set-output-truncates-multiline-strings/16852

They would have to do something like this in order for it to work right....
```bash
DIGEST=$(sha256sum binary1 binary2)
DIGEST="${DIGEST//'%'/'%25'}"
DIGEST="${DIGEST//$'\n'/'%0A'}"
DIGEST="${DIGEST//$'\r'/'%0D'}"
```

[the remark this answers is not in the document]
ianlewis: I went ahead and made the change, though I worry a bit about this UX

cmd/slsa-github-generator/attest.go
```diff
@@ -0,0 +1,126 @@
...
+	// shaCheck verifies a hash is has only hexidecimal digits and is 64
+	// characters long.
+	shaCheck = regexp.MustCompile(`^[a-fA-F0-9]{64}$`)
+)
+
+func parseSubjects(subjectsStr string) ([]intoto.Subject, error) {
+	var parsed []intoto.Subject
+
+	subjects := strings.Split(subjectsStr, "\n")
+	for _, s := range subjects {
```
MarkLodato, line 45: nit: skip over blank lines. There might be trailing newlines, and we might as well ignore inner blank lines as well.

cmd/slsa-github-generator/attest.go
```diff
@@ -0,0 +1,126 @@
...
+		parts := strings.SplitN(s, " ", 2)
+		if len(parts) == 0 {
```
MarkLodato, line 47: nit: `SplitN` will never return 0 results since the separator is non-empty. Optionally, `Cut()` may be more readable?

```go
shaDigest, name, found := strings.Cut(s, " ")
name = strings.TrimSpace(name)
```

cmd/slsa-github-generator/attest.go
```diff
@@ -0,0 +1,126 @@
...
+		name := strings.TrimSpace(parts[1])
+		parsed = append(parsed, intoto.Subject{
+			Name: name,
+			Digest: slsav02.DigestSet{
+				// Lowercase the sha digest to comply with the SLSA spec.
+				"sha256": strings.ToLower(shaDigest),
+			},
+		})
+	}
+	return parsed, nil
```
MarkLodato, line 67: Check that each name is unique, as required by the spec. (OK to do in a follow-up PR.)

MarkLodato (approved):

laurentsimon (approved):

ianlewis:
> > > At least, that was my thinking.
> > > The simplest way to use it would be something like:
> > > ```go
> > > r := slsa.NewWorkflowRun(subjects, githubContext)
> > > r.BuildType = golangSpecificBuildType
> > > r.BuildConfig = /* custom Go builder config (steps etc.) */
> > > p, _ := slsa.HostedActionsProvenance(r)
> > > /* sign and upload */
> > > ```
> > 
> > 
> > Let's start with this, it's simple and is flexible to test in various scenarios/use cases.
> 
> Just a last-minute thought. We could also have `WithXXX()` functions, which may be simple too:
> 
> ```go
> r := slsa.NewWorkflowRun(subjects, githubContext)
>        .WithBuildType("some_string")
>        .WithBuildConfig(some_struct)
> ```
> 
> ```go
> func (self * WorkflowRun) WithBuildConfig(arg interface{}) *WorkflowRun{
>   self.BuildConfig = arg
>   return self
> }
> ```
> 
> Can do that in a follow-up PR if we want

BuildConfig is empty so this works but I think the use cases for others will often be to alter individual fields rather than set the full value (`Materials`, `Environment`, etc.) so I'm not sure this pattern really holds. I think we need some more time to see how it will actually be used.

cmd/slsa-github-generator/attest.go
ianlewis: #12

laurentsimon:
> >   
> > ```go
> > func (self * WorkflowRun) WithBuildConfig(arg interface{}) *WorkflowRun{
> >   self.BuildConfig = arg
> >   return self
> > }
> > ```

> >   
> > Can do that in a follow-up PR if we want
> 
> BuildConfig is empty so this works but I think the use cases for others will often be to alter individual fields rather than set the full value (`Materials`, `Environment`, etc.) so I'm not sure this pattern really holds. I think we need some more time to see how it will actually be used.

Maybe `AddMaterial()` and `AddEnvironment()` would work. Let's just file an issue for later discussion

cmd/slsa-github-generator/attest.go
ianlewis: Done.

cmd/slsa-github-generator/attest.go
ianlewis: `strings.Fields` seems to be better for our use case so I went with that.

ianlewis:
I think we're good to merge from my end.

ianlewis:
> Maybe `AddMaterial()` and `AddEnvironment()` would work. Let's just file an issue for later discussion

Filed #13

laurentsimon:
Merging, thanks @ianlewis !

PR #16, opened 2022-04-06
Pre submit

### LINKED ISSUE #10: Add pre-submit and tests

ianlewis:
Fixes #10

### DIFF
--- a/.github/workflows/pre-submit.yml
+++ b/.github/workflows/pre-submit.yml
@@ -1,0 +1,25 @@
+name: Pre submits
+on: [pull_request, workflow_dispatch]
+
+permissions: read-all
+
+jobs:
+  pre-submit:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.3.4
+
+      - name: setup-go
+        uses: actions/setup-go@bfdd3570ce990073878bf10f6b2d79082de49492 # v2.2.0
+        with:
+          go-version: "1.18"
+
+      - name: Run tests for builder
+        run: |
+          set -euo pipefail
+
+          # Download dependencies.
+          go mod vendor
+          # Test.
+          go test -mod=vendor -v ./...
--- a/.github/workflows/pre-submit.yml
+++ b/.github/workflows/pre-submit.yml
@@ -15,7 +15,7 @@ jobs:
         with:
           go-version: "1.18"
 
-      - name: Run tests for builder
+      - name: Run tests
         run: |
           set -euo pipefail
 
--- a/github/oidc_test.go
+++ b/github/oidc_test.go
@@ -1,0 +1,20 @@
+package github
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestRequestOIDCToken(t *testing.T) {
+	wantToken := &OIDCToken{JobWorkflowRef: "hoge"}
+	_, stop := NewTestOIDCServer(wantToken)
+	defer stop()
+
+	gotToken, err := RequestOIDCToken("fuga")
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if want, got := wantToken, gotToken; !reflect.DeepEqual(want, got) {
+		t.Errorf("unexpected workflow ref, want: %#v, got: %#v", want, got)
+	}
+}
--- a/github/oidc_test.go
+++ b/github/oidc_test.go
@@ -1,20 +1,71 @@
 package github
 
 import (
-	"reflect"
+	"encoding/base64"
+	"errors"
+	"fmt"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
 func TestRequestOIDCToken(t *testing.T) {
-	wantToken := &OIDCToken{JobWorkflowRef: "hoge"}
-	_, stop := NewTestOIDCServer(wantToken)
-	defer stop()
-
-	gotToken, err := RequestOIDCToken("fuga")
-	if err != nil {
-		t.Fatalf("unexpected error: %v", err)
+	testCases := []struct {
+		name     string
+		audience string
+		expected *OIDCToken
+		raw      string
+		err      error
+	}{
+		{
+			name:     "basic token",
+			audience: "hoge",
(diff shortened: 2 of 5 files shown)


.github/workflows/pre-submit.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.3.4
+
+      - name: setup-go
+        uses: actions/setup-go@bfdd3570ce990073878bf10f6b2d79082de49492 # v2.2.0
+        with:
+          go-version: "1.18"
+
+      - name: Run tests
```
asraa, line 18: Maybe let's also include a job for a test build for the binary? May catch some compilation bugs https://github.com/slsa-framework/slsa-github-generator/blob/main/cmd/slsa-github-generator/main.go

asraa (approved):

.github/workflows/pre-submit.yml
laurentsimon: see https://github.com/slsa-framework/slsa-github-generator-go/blob/main/.github/workflows/pre-submit.yml, we have an additional `go build -mod=vendor`. Maybe this is implicitly done when running `go test`..? Let's just add it to be on the safe side

github/oidc_test.go
```diff
@@ -0,0 +1,20 @@
...
+func TestRequestOIDCToken(t *testing.T) {
+	wantToken := &OIDCToken{JobWorkflowRef: "hoge"}
+	_, stop := NewTestOIDCServer(wantToken)
+	defer stop()
+
+	gotToken, err := RequestOIDCToken("fuga")
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if want, got := wantToken, gotToken; !reflect.DeepEqual(want, got) {
```
laurentsimon, line 17: Maybe use github.com/google/go-cmp/cmp, you can then use `cmp.Diff` to show diff.
See example https://github.com/slsa-framework/slsa-github-generator-go/blob/main/pkg/build_test.go

slsa/provenance_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+	"github.com/slsa-framework/slsa-github-generator/github"
+)
+
+func TestHostedActionsProvenance(t *testing.T) {
+	testCases := []struct {
+		name     string
+		r        WorkflowRun
+		expected *intoto.ProvenanceStatement
+		err      bool
```
laurentsimon, line 18: how about making this an `error` type? Then you can test the error is of the expected type, see https://github.com/slsa-framework/slsa-verifier/blob/main/main_test.go#L306
This has been useful for me in the past, I caught some non-expected errors :-)

laurentsimon (approved):
LGTM overall, left a few comments to improve error checking

.github/workflows/pre-submit.yml
ianlewis: Yeah, go test ./... will tell you if a package isn't buildable even if there are no tests.

slsa/provenance_test.go
ianlewis: I can do that though I don't think I'm using any particular types at the moment (just whatever type happens to be returned by fmt.Errorf). I suppose I could return different types but didn't necessarily think it was worth it as it really wouldn't be used outside of tests. Maybe I'll just return some unexported types for now.

.github/workflows/pre-submit.yml
ianlewis: Also, building the binary won't necessarily use all packages (only those referenced), so we would need to do `go build ./...` which `go test ./...` does anyway.

slsa/provenance_test.go
laurentsimon: I think it's still useful to test that the internal function work as expected. I tend do do this:
```
var errSomeName = errors.New("some msg)

fun() {
  return fmt.Errorf("%w", errSomeName)
}
```

Would this work?

ianlewis: Yeah, I'll make this a bit better. At least for this file it's mute since it's not really used. I'll add some more tests over on RequestOIDCToken since that's really where the errors are produced.

In general, it's a bit annoying in Go to add custom error messages with your own types.

One can do:
```golang
if err := someFunc(); err != nil { 
  return fmt.Errorf("%w: %v", errSomeName, err)
}
```
and get the property that `errors.Is(err, errSomeName) == true` but then you loose the type of the error returned from someFunc and I'd not really consider that a good practice.

The "right" way to do it is to go full bore with your own wrapping error type but it gets cumbersome quick as you need to do that for every type you want to test for. We could have our own internal base wrapping type that can compare types and allow custom messages but it's maybe a bit of a pain to use: https://go.dev/play/p/6f1bDFwr8vj

ianlewis: I think is to define our own types and use `errors.As` to compare is probably what we want so I'll take a stab at it.

ianlewis: Something like this is what I came up with, but I'll hold off on adding this and just use static errors for now.
https://go.dev/play/p/H6mrl41DzAX

.github/workflows/pre-submit.yml
ianlewis: So I think this is unnecessary since `go test` is designed to catch build errors as well and `go test ./...` will catch build and test errors in all packages, not just those imported transitively from `main.go` but feel free to veto again if you really think it's necessary.

slsa/provenance_test.go
laurentsimon: Does  the following work?

```go
fmt.Errorf("%w: %w", errSomeName, err)
```

.github/workflows/pre-submit.yml
laurentsimon: LGTM then. If there's a link to back up our claims, even better.

slsa/provenance_test.go
ianlewis: Errorf only supports one wrapped error AFAICT
https://cs.opensource.google/go/go/+/refs/tags/go1.18:src/fmt/errors.go;l=17

But incidentally we could do something like wrap twice maybe.
```golang
fmt.Errorf("%w: %v", fmt.Errorf("%w", errSomeName), err)
```

The real trick comes when checking with either `errors.Is` which checks strict equality or `errors.As` which checks type equality. In any case, we can keep this in mind for future PRs. I think we should be good for this one.

laurentsimon (approved):
Thanks

PR #19, opened 2022-04-06
Allow whitespace in subject name.

### LINKED ISSUE #18: Support `name` containing whitespace
The current implementation disallows whitespace in the subject `name` because it uses `strings.Fields`. The spec does not disallow whitespace, and spaces in filenames are not uncommon, so it would be best if the implementation allowed that.

To fix, using `strings.Cut` (or `strings.SplitN` or a regexp) would avoid the issue.

ianlewis:
Fixes #18

### DIFF
--- a/cmd/slsa-github-generator/attest.go
+++ b/cmd/slsa-github-generator/attest.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"bufio"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


MarkLodato (approved):

ianlewis:
BTW, I don't have commit access so someone has to merge these PRs for me once they are approved or give me access so I can do it myself.

MarkLodato:
Oops! Just gave you access now.

PR #24, opened 2022-04-08
Remove .git from repo URI

### LINKED ISSUE #17: Drop `.git` in URIs
The generated URIs append `.git` to the end, but this is wrong.

```
        "uri": "git+https://github.com/ianlewis/actions-test@refs/heads/main.git",
```

I suggest removing this altogether. (Another option would be to move it before the `@`, but that doesn't seem to have any value.)

ianlewis:
Fixes #17


asraa (approved):
thanks!

PR #25, opened 2022-04-08
WIP Verify OIDC Tokens

### LINKED ISSUE #9: Verify requested OIDC tokens
@asraa said:
> we should probably verify that the client ID is set to the requested audience and that the issuer was github This library would be useful! https://github.com/coreos/go-oidc/blob/v2.2.1/verify.go#L73
> 
> obviously we are generating this internally and not expecting this to be passed in thru user args, but it is probably good to verify
> 
> also see where this is used for usage https://github.com/sigstore/fulcio/blob/b2186c01da1ddf807bde3ea8c450226d8e001d88/pkg/config/config.go#L195-L199

ianlewis:
Fixes #9

### DIFF
--- a/cmd/slsa-github-generator/attest.go
+++ b/cmd/slsa-github-generator/attest.go
@@ -109,12 +109,12 @@ run in the context of a Github Actions workflow.`,
 				check(errors.New("expected at least one subject"))
 			}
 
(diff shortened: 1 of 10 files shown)


ianlewis:
Will recreate PR

PR #30, opened 2022-04-14
Add a basic CODEOWNERS file.

ianlewis:


laurentsimon (approved):
Shall we create a maintainer's team and set that in CODEOWNERS in the future? This would let GitHub round robin for reviewers. Wdut?

ianlewis:
Yeah, it's a good idea. It might be nice to get other folks to review if there is a slsa-framework org-wide team for this kind of project.

PR #31, opened 2022-04-15
Add --predicate cli flag

### LINKED ISSUE #21: Support outputting unsigned provenance
In case a user wants to create GitHub workflow provenance and then attach that to a container image using cosign CLI. By doing this, the attached container image would include rekor bundle and certificate in the cosign simple signing schema.

e.g.

```
$ ./slsa-github-generator attest --unsigned --output predicate.json

$ ./cosign attest --predicate predicate.json --type intoto gcr.io/asra-ali/hello-ko
```

The generic provenance builder should only output the `predicate`

cc @laurentsimon

ianlewis:
Fixes #21

### DIFF
--- a/cmd/slsa-github-generator/attest.go
+++ b/cmd/slsa-github-generator/attest.go
@@ -17,6 +17,7 @@ package main
 import (
 	"bufio"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -97,6 +98,7 @@ func getFile(path string) (io.Writer, error) {
 
 // attestCmd returns the 'attest' command.
 func attestCmd() *cobra.Command {
+	var predicatePath string
 	var attPath string
 	var subjects string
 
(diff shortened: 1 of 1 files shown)


laurentsimon (approved):
Thanks!

Do we plan to expose this in the API too?

I've built a reusable workflow for https://github.com/google/ko and I need exactly this feature. But I think I'd rather use an API

ianlewis:
> Thanks!
> 
> Do we plan to expose this in the API too?
> 
> I've built a reusable workflow for https://github.com/google/ko and I need exactly this feature. But I think I'd rather use an API

Do you think it's not appropriate to do something like what we've done here and called `HostedActionsProvenance` and just use the `Predicate` field?

```
p, _ := slsa.HostedActionsProvenance(someWorkflowRun)
fmt.Println(p.Predicate)
```

laurentsimon:
this seems good enough to me. We can do the same to set the material sections (I was mostly curious if you thought it should be treated differently - I don't see why :-))

Issue #8, opened 2022-04-03
Support more metadata fields


ianlewis:
Add support for more metadata fields:
```
    "metadata": {
      "buildInvocationId": "<STRING>",
      "buildStartedOn": "<TIMESTAMP>",
      "buildFinishedOn": "<TIMESTAMP>",
      "completeness": {
        "parameters": true/false,
        "environment": true/false,
        "materials": true/false
      },
      "reproducible": true/false
    },
```

ianlewis:
`buildInvocationId` could be set to `github.run_id` but that doesn't change if the workflow is re-run. The only thing that changes in that case is `github.run_attempt` which gets incremented each re-run but reset to 1 each new run. So it seems we'll need to combine both these values in order to get a unique ID.

`github.run_number` could also potentially identify the run along with the `github.workflow` but I think `run_id` and `run_attempt` is closer to what we want.

ianlewis:
This issue could also be thought to be handled by #47 as ecosystem builders can extend the invocation with the build information like start/end time and reproducibility.

PR #32, opened 2022-04-15
Add build invocation ID

ianlewis:
Updates #8

### DIFF
--- a/slsa/provenance.go
+++ b/slsa/provenance.go
@@ -42,6 +42,12 @@ func HostedActionsProvenance(w WorkflowRun) (*intoto.ProvenanceStatement, error)
 		builderID = fmt.Sprintf("https://github.com/%s", t.JobWorkflowRef)
 	}
 
+	buildInvocationID := w.GithubContext.RunID
+	if w.GithubContext.RunAttempt != "" {
+		// NOTE: RunID does not get updated on re-runs so we need to include RunAttempt.
+		buildInvocationID = fmt.Sprintf("%s-%s", w.GithubContext.RunID, w.GithubContext.RunAttempt)
+	}
+
 	return &intoto.ProvenanceStatement{
 		StatementHeader: intoto.StatementHeader{
 			Type:          intoto.StatementInTotoV01,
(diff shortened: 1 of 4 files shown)


slsa/provenance.go
```diff
@@ -58,7 +64,8 @@ func HostedActionsProvenance(w WorkflowRun) (*intoto.ProvenanceStatement, error)
 			Materials:   w.Materials,
 			// TODO(https://github.com/slsa-framework/slsa-github-generator/issues/8): support more metadata fields.
 			Metadata: &slsa.ProvenanceMetadata{
-				Completeness: w.Completeness,
+				BuildInvocationID: buildInvocationID,
+				Completeness:      w.Completeness,
```
asraa, line 68: i'm just curious about this, but is a workflowrun materials really completeness true (from `NewWorkflowRun`)?

For generic generator, I'm not sure

slsa/provenance.go
```diff
@@ -42,6 +42,12 @@ func HostedActionsProvenance(w WorkflowRun) (*intoto.ProvenanceStatement, error)
 		builderID = fmt.Sprintf("https://github.com/%s", t.JobWorkflowRef)
 	}
 
+	buildInvocationID := w.GithubContext.RunID
+	if w.GithubContext.RunAttempt != "" {
+		// NOTE: RunID does not get updated on re-runs so we need to include RunAttempt.
+		buildInvocationID = fmt.Sprintf("%s-%s", w.GithubContext.RunID, w.GithubContext.RunAttempt)
```
asraa, line 48: Could you please add some documentation on the format of this field? 

(Maybe add here https://github.com/slsa-framework/slsa-github-generator#provenance-example and add a note)

asraa (commented):
Thanks!

slsa/provenance.go
ianlewis: Completeness gets taken from the `WorkflowRun`. The generic generator only supplies `completeness.parameters = true` based on a suggestion from @MarkLodato. All the rest of the fields are false.

https://github.com/slsa-framework/slsa-github-generator/blob/main/slsa/github.go#L117-L119

See https://github.com/slsa-framework/slsa-github-generator/pull/3#discussion_r840747519 for context when it was added.

slsa/provenance.go
ianlewis: I will add something related to the invocation ID. I also added an issue to better document the fields determined by the `buildType` where we have made some judgements about the data there. 
https://github.com/slsa-framework/slsa-github-generator/issues/33

ianlewis: Done.

slsa/provenance.go
ianlewis: I'm going to resolve as I think there isn't anything to do here but feel free to reopen if necessary.

laurentsimon (approved):

laurentsimon:
Ready to merge?

ianlewis:
> Ready to merge?

Yep. Just waiting for an approval.

Issue #4, opened 2022-03-31
Feature: use OICD for dynamic checkout of reusable workflow


laurentsimon:
See https://github.com/slsa-framework/slsa-github-generator-go/pull/25

laurentsimon:
cc/ [ianlewis](https://github.com/ianlewis/slsa-github-generator/tree/slsa-gen-lib)

ianlewis:
We added the action but we still don't use it in the re-usable workflow.

Testing it out, AFAICT you need to check out the repo before calling even actions defined locally in the same repo, so we have a chicken-and-egg scenario where we want to run the detect-workflow action but we don't now the repo and ref to check out in order to get access to run it.

Maybe @asraa's original inline bash is actually better since it's self-containerd in the reusable workflow's yaml?

laurentsimon:
I'm fine with a bash action if it works.. and if it's the only solution :-)

ianlewis:
I think it's ok for now. I think it might be good as it's own action somewhere but it's perhaps not top priority.

PR #35, opened 2022-04-22
Feature: add workflow detection action

asraa:

Fixes https://github.com/slsa-framework/slsa-github-generator/issues/4


asraa:
I'm not sure I can test this out since the `ACTIONS_ID_TOKEN_REQUEST_TOKEN` var doesn't seem to be set if I test the local action inside the repository.

However, this works:
https://github.com/asraa/slsa-on-github-test/runs/6134846300?check_suite_focus=true

.github/actions/detect-workflow/action.yml
```diff
@@ -0,0 +1,31 @@
...
+    description: The current workflow reference
+    value: ${{ steps.detect.outputs.ref }}
+runs:
+  using: 'composite'
+  steps:
+    - run: | 
+        set -euo pipefail
+        
+        status_code=$(curl -sS -H "Authorization: Bearer $ACTIONS_ID_TOKEN_REQUEST_TOKEN" "$ACTIONS_ID_TOKEN_REQUEST_URL&audience=slsa-framework/slsa-github-generator-go/builder" -o jwt.json -w '%{http_code}')
+        if [[ $status_code -ge 400 ]]; then
```
laurentsimon, line 17: let's check `$status_code <200 || $status_code >= 300`, since all 2xx should be acceptable.
(the 400 is my fault, I copied this code without checking what it was doing)

laurentsimon (approved):

.github/actions/detect-workflow/action.yml
```diff
@@ -0,0 +1,31 @@
...
+    description: The current workflow repository, format org/repository
+    value: ${{ steps.detect.outputs.repository }}
+  ref:
+    description: The current workflow reference
+    value: ${{ steps.detect.outputs.ref }}
+runs:
+  using: 'composite'
+  steps:
+    - run: | 
+        set -euo pipefail
```
ianlewis, line 14: I thought that maybe we could use the Go code to do this since we have retrieval, parsing, and verification in the [github](https://github.com/slsa-framework/slsa-github-generator/tree/main/github) package. Though I realize it would be a bit more involved to compile and run it and shell is convenient.

ianlewis (approved):
Thanks!

.github/actions/detect-workflow/action.yml
asraa: updated this with Ian's go code, which has this check

.github/actions/detect-workflow/action.yml
asraa: updated to use this! was extremely easy PTAL

.github/actions/detect-workflow/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.18 as builder
```
laurentsimon, line 1: maybe pin by hash

.github/actions/detect-workflow/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+COPY . /app
+
+RUN go get -d -v
+
+# Statically compile our app for use in a distroless container
+RUN CGO_ENABLED=0 go build -ldflags="-w -s" -v -o app .
+
+# A distroless container image with some basics like SSL certificates
+# https://github.com/GoogleContainerTools/distroless
+FROM gcr.io/distroless/static
```
laurentsimon, line 13: maybe pin by hash

.github/actions/detect-workflow/main.go
```diff
@@ -0,0 +1,40 @@
...
+
+func main() {
+	ctx := context.Background()
+
+	c, err := github.NewOIDCClient()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	audience := "slsa-framework/slsa-github-generator/detect-workflow"
```
laurentsimon, line 20: I suspect this may fail when use from a different repo? Can we pass audience as an option?

asraa: done

laurentsimon (approved):

.github/actions/detect-workflow/action.yml
ianlewis: Good to hear! Kudos for doing this.

.github/actions/detect-workflow/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang@sha256:3b7fa1bed1510cfdadbbd839ab1c2c4ca5ee86bc259d2cb71142f6bd4e70b59f as builder
+
```
laurentsimon, line 2: a little late, but I was wondering if we could avoid dockerfile by using:
```
- uses: action/setup-go@xxx
- uses: action/checkout@xxx
- runs: go build
- runs: run the program
```

I ask because docker build/run take longer, and because it would avoid maintaining the hashes in the docker file.

I don't know if this code works, though.. especially the part on checkout. I feel that it might, the same way that the dockerfile is able to reference its own code...

PR #38, opened 2022-04-27
Use workflow detect

ianlewis:
Adds repository and ref detection to the reusable workflow.

Uses bash code to get the OIDC token and detect the repository/ref. This is because the reusable workflow yaml is retrieved and used on it's own and has no context for the rest of the reusable workflow's repo.

We would like to use the lovely code that @asraa wrote to [detect the workflow repository and ref](https://github.com/slsa-framework/slsa-github-generator/tree/main/.github/actions/detect-workflow), but we have a chicken and egg scenario because we need to [check out the slsa-github-generator repo](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-an-action-in-the-same-repository-as-the-workflow) before using actions defined in the same repo.

I'm not particularly happy with any of the options so I'm open to suggestions for more alternatives.

## This PR's current implementation: Bash code to get and parse the OIDC token
Bash code to get the OIDC token with curl and parse with jq

Pros:
- Code is inline in the workflow so no need to checkout the rest of the repo.
- No need to use a static repo/ref

Cons:
- Need to copy this code and maintain in all reusable workflows
- No OIDC token verification

## Alternative 1: Set a static repo/ref to a well known version when we release one in order to bootstrap the detection process.
  1. Checkout static repo/ref
  1. Run detect-workflow
  1. Check out repo/ref detected by the detect-workflow step.

Pros:
- Able to reuse OIDC token retrieval code from library
- Able to reuse the action code
- Includes OIDC token verification

Cons:
- Need to bootstrap with a static repo/ref

### DIFF
--- a/.github/workflows/slsa2_provenance.yml
+++ b/.github/workflows/slsa2_provenance.yml
@@ -48,6 +48,11 @@ jobs:
       id-token: write # Needed for keyless.
       contents: read
     steps:
+      # Check out the base repository to get the detect-workflow action.
+      - name: Checkout the repository
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.3.4
+        with:
+          fetch-depth: 0
       - name: Run local detect action
         id: detect
         uses: ./.github/actions/detect-workflow
--- a/.github/workflows/slsa2_provenance.yml
+++ b/.github/workflows/slsa2_provenance.yml
@@ -17,11 +17,6 @@ name: SLSA provenance generator
 permissions:
   contents: read
 
-env:
-  # This repo.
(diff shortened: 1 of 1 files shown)


ianlewis:
Since the oidc code and detection action is generic, we could potentially put it in its own repo as well. Like under github.com/google. That way it wouldn't be such a big deal to reference it via a specific version/hash.

laurentsimon:
I think I'm in for a new repo. Once we merge the ecosystem-specific repos into this one, we won't be able to use the action at all. Not sure where the action should live: google, openssf, or slsa-framework?

laurentsimon (approved):
Thanks!

PR #39, opened 2022-04-29
Return log entry from rekor upload call

lumjjb:
Return log entry object so that the caller can output information about log ID.

Signed-off-by: Brandon Lum <lumjjb@gmail.com>

### DIFF
--- a/signing/sigstore/signer.go
+++ b/signing/sigstore/signer.go
@@ -24,6 +24,7 @@ import (
 	"github.com/sigstore/cosign/cmd/cosign/cli/rekor"
 	"github.com/sigstore/cosign/pkg/cosign"
 	"github.com/sigstore/cosign/pkg/providers"
+	"github.com/sigstore/rekor/pkg/generated/models"
 	"github.com/sigstore/sigstore/pkg/signature/dsse"
 
 	intoto "github.com/in-toto/in-toto-golang/in_toto"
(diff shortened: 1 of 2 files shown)


MarkLodato:
Sorry, I don't have enough context to review.

asraa:
@lumjjb Could you please fix the test return val?

lumjjb:
@asraa need another workflow approval ! Tks

signing/sigstore/signer.go
```diff
@@ -115,15 +116,16 @@ func (s *Signer) Sign(ctx context.Context, p *intoto.ProvenanceStatement) (*Atte
 }
 
 // Upload uploads the signed attestation to the rekor transparency log.
-func (s *Signer) Upload(ctx context.Context, att *Attestation) error {
+func (s *Signer) Upload(ctx context.Context, att *Attestation) (*models.LogEntryAnon, error) {
```
ianlewis, line 119: @lumjjb I originally thought this would become part of a generic interface that could support other signers and transparency logs besides cosign/rekor. Is there a use case you had for this?

lumjjb: yea agreed - was thinking about that as well when i was making this change, but wasn't sure what the interface would look like so exposed it for now since we wanted to expose the log ID so that it can be retrieved more easily (instead of having to do a search on rekor)

ianlewis: Yeah, on the other side of things I didn't have a good idea what the use cases would be so I left it off.

Where did you want to expose the ID? Maybe a more generic struct or interface with a string ID would work? (similar to the Attestation struct)

PR #41, opened 2022-05-03
✨ Update cosign to v17

laurentsimon:
There seems to be an API breaking between v1.6 and v1.7, see https://github.com/slsa-framework/slsa-github-generator-go/runs/6265968724?check_suite_focus=true

We need to provide an OIDC redirect URL. I think in our case there's none because we don't have a user in the loop.


ianlewis (approved):
LGTM

PR #42, opened 2022-05-03
Make Sign take generic intoto.Statement

lumjjb:
Make signing library more generic for intoto statements.

Signed-off-by: Brandon Lum <lumjjb@gmail.com>

### DIFF
--- a/cmd/slsa-github-generator/attest.go
+++ b/cmd/slsa-github-generator/attest.go
@@ -130,7 +130,10 @@ run in the context of a Github Actions workflow.`,
 
 			if attPath != "" {
 				s := sigstore.NewDefaultSigner()
-				att, err := s.Sign(ctx, p)
+				att, err := s.Sign(ctx, &intoto.Statement{
+					StatementHeader: p.StatementHeader,
+					Predicate:       p.Predicate,
+				})
 				check(err)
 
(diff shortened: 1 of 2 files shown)


asraa (approved):

cmd/slsa-github-generator/attest.go
```diff
@@ -130,7 +130,10 @@ run in the context of a Github Actions workflow.`,
 
 			if attPath != "" {
 				s := sigstore.NewDefaultSigner()
-				att, err := s.Sign(ctx, p)
+				att, err := s.Sign(ctx, &intoto.Statement{
```
laurentsimon, line 133: I think if we want to use cosign's `attest` API, we still need a way to pass a predicate only. Probably fine to leave as-is for now, and wait till we actually need it. (I use cosign attest via the CLI for the ko containers atm)
@asraa any thought?

asraa: I don't think there's any way to write `AttestCmd` with outputs for the attestation or log entry right now -- (btw are you holding on to the uploaded log entry anywhere?)

I think we will need to wait for a refactor of that command to support Blobs, or a new `AttestBlobCmd`.

For the blob case, how would `Attest` compute the subject digest without either the binary blob as a second argument or the entire Statement?

PR #43, opened 2022-05-09
Workflow name

### LINKED ISSUE #11: Entrypoint is the workflow name
At it stands the provenance `invocation.entryPoint` is set to the `workflow` field from the Github Actions workflow. This is the name of the calling workflow or the path if the name is not set.

This seems wrong to me. I think the path to the calling workflow yaml file would be more useful but I'm not sure. We could get from `event.workflow` in the github context.

ianlewis:
Fixes #11

Before the change the `predicate.invocation.configSource.entryPoint` has the name of the user workflow (not the reusable workflow) specified in the workflow yaml:
```yaml
"predicate": {
    "invocation": {
        "entryPoint": "SLSA provenance"
        ...
      },
...
```

After this change it's a path to the workflow:
```yaml
"predicate": {
    "invocation": {
      "configSource": {
        "entryPoint": "https://github.com/ianlewis/actions-test/.github/workflows/provenance-only.yml"
        ...
      },
...
```


joshuagl (approved):
Code looks good. Before and after examples of the provenance would make PRs like this easier to review.

MarkLodato:
Agreed. I can't really review because I'd need to see an example to be sure that it's correct.

Speaking of which, what is the plan for unit tests? That would make these PRs much easier to review.

ianlewis:
Sure I can add tests. It was smallish change but maybe tests are warranted. I'll also remember to add before/after notes to the PR description for future PRs.

ianlewis:
@joshuagl @MarkLodato I added a before/after example and a simple unit test for the entrypoint. PTAL

slsa/github.go
```diff
@@ -63,12 +66,25 @@ type WorkflowParameters struct {
...
+	// Create the entrypoint from the repository URI @ workflow path in the
+	// event. `workflow` is not used from the github context because it includes
+	// the workflow name rather than the path.
+	// NOTE: ServerURL does not have a trailing slash and Repository does not
+	// have a leading slash.
+	entryPoint := fmt.Sprintf("%s/%s",
+		c.ServerURL,
+		path.Join(
+			c.Repository,
+			fmt.Sprintf("%s", c.Event["workflow"]),
```
ianlewis, line 78: @MarkLodato @joshuagl Now that I look at this again, the entrypoint could maybe just be `c.Event["workflow"]` which is the the path to the user workflow within the repository pointed to by URI (line 88). We can maybe leave off the ServerURL and Repository part rather than a full URI. Any thoughts?

joshuagl: Agreed. I like `entryPoint` being the path within the repo at `uri`

MarkLodato: Yes, being relative to the repo was the intention.

ianlewis: Done.

laurentsimon:
Feel free to merge. If all works well, the e2e test will pick the change and fail

PR #45, opened 2022-05-12
Add Signer and TransparencyLog interfaces.

ianlewis:
Adds interfaces which will allow the generator to different signers in the future.

### DIFF
--- a/signing/signer.go
+++ b/signing/signer.go
@@ -1,0 +1,35 @@
+package signing
+
+import (
+	"context"
+
+	intoto "github.com/in-toto/in-toto-golang/in_toto"
+)
+
+// Attestation is a signed attestation.
+type Attestation interface {
+	// Cert returns the certificate used to sign the attestation.
+	Cert() []byte
+
+	// Bytes returns the signed attestation as an encoded DSSE JSON envelope.
+	Bytes() []byte
+}
+
+// Signer is used to sign provenance statements.
+type Signer interface {
+	// Sign signs the given provenance statement and returns the signed
+	// attestation.
+	Sign(context.Context, *intoto.Statement) (Attestation, error)
+}
+
+// LogEntry represents a transparency log entry.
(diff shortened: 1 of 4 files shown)


ianlewis:
@lumjjb Can you take a look at this too? I updated the transparency log Upload to return a generic log entry with ID. Would that support your use case?

signing/signer.go
```diff
@@ -0,0 +1,35 @@
...
+type Signer interface {
+	// Sign signs the given provenance statement and returns the signed
+	// attestation.
+	Sign(context.Context, *intoto.Statement) (Attestation, error)
+}
+
+// LogEntry represents a transparency log entry.
+type LogEntry interface {
+	// ID returns the ID of the transparency log entry.
+	ID() string
```
lumjjb, line 28: I think there is some discussion that the UUID should be returned instead but there's a limitation of the cosign interface that doesn't return the UUID. (@asraa). Since this for now is pretty implementation specific, it is likely going to change in the future. 

I think a nice addition here is to add the rekor address here as well.

lumjjb (approved):
LGTM. I think there is a caveat that these would still generally be interfaces that are subject to quite a bit of change. It's likely that if we introduce an alternative tlog or attestation type that we need to encode a bit more information to handle the data.

At the same time, because there aren't any alternatives in the near future, i am in agreement with simplicity as it is now. Small note on UUID that I discussed with @asraa

laurentsimon (approved):

signing/signer.go
ianlewis: The ID here is meant to be fairly generic as this will likely be a different value for different tlog implementations.

We could maybe add a break-glass `Metadata() map[string]interface{}` method or something to return arbitrary info if we have more things that are rekor specific. Or we could make the rekor Log entry type public on the package and allow type conversion back to it.

Happy to entertain different options.

PR #46, opened 2022-05-13
Revert back to using github.workflow as entrypoint

ianlewis:
`workflow` is not present in all event payloads so we'll need to go back to using `workflow` from the `github`context for now.
https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads

Updates #11

### DIFF
--- a/slsa/github.go
+++ b/slsa/github.go
@@ -15,8 +15,6 @@
 package slsa
 
 import (
-	"fmt"
-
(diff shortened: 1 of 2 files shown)


ianlewis:
This should fix https://github.com/slsa-framework/slsa-github-generator-go/issues/108 once that project updates.

PR #47, opened 2022-05-16
Refactor and add BuildType interface

### LINKED ISSUE #2: Create provenance generation library

### LINKED ISSUE #13: Improve WorkflowRun customization
Consider adding utility functions on the WorkflowRun to customize fields. This will allow generators that use the API to set their own build type and fields easier.

Some earlier suggestions from @laurentsimon 

> Just a last-minute thought. We could also have `WithXXX()` functions, which may be simple too:
> 
> ```go
> r := slsa.NewWorkflowRun(subjects, githubContext)
>        .WithBuildType("some_string")
>        .WithBuildConfig(some_struct)
> ```
> 
> ```go
> func (self * WorkflowRun) WithBuildConfig(arg interface{}) *WorkflowRun{
>   self.BuildConfig = arg
>   return self
> }
> ```
> 
> Can do that in a follow-up PR if we want

> Maybe `AddMaterial()` and `AddEnvironment()` would work. Let's just file an issue for later discussion.

ianlewis:
Depends on #69 
Fixes #13 
Fixes #2 

Improve the API by adding the `BuildType` interface that builders can extend to implement their own build types. The intention is to improve the API and not rely so much on function parameters and allow the API to scale easier with minimal backwards incompatible changes.

The Github API client was added in order to get the workflow path via the API and solve #11 in a later PR.

The `BuildType` interface represents a SLSA build type and the `GithubActionsBuild` implements it with a basic Github Actions build type that can be extended by ecosystem builders.

A typical ecosystem builder might look like:

```go
type GoBuild struct {
    GithubActionsBuild
}

func (b *GoBuild) URI() string {
	return "https://github.com/slsa-framework/slsa-github-generator-go@v1"
}

func (b *GoBuild) BuildConfig(context.Context) (interface{}, error) {
	return buildConfig{
		Version: buildConfigVersion,
		Steps: []step{
			// Single step.
			{
				Command: com,
				Env:     env,
			},
		},
	}
}
```

You can then create it and generate SLSA from it:
```go
b := GoBuild{
	GithubActionsBuild: NewGithubActionsBuild(subjects, githubContext)
}
g := NewHostedActionsGenerator(&b)
p, _ := g.Generate(context.Background())
```

Github API and OIDC clients are a bit annoying to deal with in tests so the generator will support not using them by setting a NilClientProvider. In that case the generator might generate invalid SLSA but is good enough for automated or e2e tests.

```go
if isTest() {
	g = g.WithClients(&NilClientProvider{})
}
```


github/workflow.go
```diff
@@ -78,3 +75,18 @@ func GetWorkflowContext() (WorkflowContext, error) {
...
 	return w, err
 }
+
+// GetToken gets the Github Actions token.
+// See: https://docs.github.com/en/actions/security-guides/automatic-token-authentication
+func GetToken() (string, error) {
+	var w struct {
+		Token string `json:"token,omitempty"`
+	}
+	ghContext, ok := os.LookupEnv(githubContextEnvKey)
```
laurentsimon, line 85: I think you can simply use `os.LookupEnv("GITHUB_TOKEN")`

ianlewis: My understanding is that you still need to add at it to you env like `GITHUB_TOKEN: {{ secrets.GITHUB_TOKEN }}` or `GITHUB_TOKEN: {{ github.token }}` but we're already adding the full `github` context anyway.

I don't see `GITHUB_TOKEN` in the default env vars: https://docs.github.com/en/actions/learn-github-actions/environment-variables

slsa/buildtype.go
```diff
@@ -0,0 +1,209 @@
...
+	return &GithubActionsBuild{
+		subject: s,
+		Context: c,
+		Clients: &DefaultClientProvider{},
+	}
+
+}
+
+// URI implements BuildType.URI.
+func (b *GithubActionsBuild) URI() string {
```
ianlewis, line 78: I think I want to make the base `GithubActionsBuild` not implement `URI` so that it's clear that this needs to be set by ecosystem builders.

I can use a private type for the provenance-only generator.

slsa/provenance.go
```diff
@@ -21,28 +21,49 @@ import (
...
+// NewHostedActionsGenerator returns a SLSA provenance generator for the given build type.
+func NewHostedActionsGenerator(bt BuildType) *HostedActionsGenerator {
+	return &HostedActionsGenerator{
+		buildType: bt,
+		clients:   &DefaultClientProvider{},
+	}
+}
+
+// Generate generates an in-toto provenance statement in SLSA v0.2 format.
+func (g *HostedActionsGenerator) Generate(ctx context.Context) (*intoto.ProvenanceStatement, error) {
```
ianlewis, line 52: I'm thinking of maybe serializing to json and returning the `[]byte` here so it's more clear that you shouldn't be modifying the `ProvenanceStatement` after it's been generated. Fields should be modified by implementing your own build type.

github/workflow.go
laurentsimon: It should be available in the re-usable workflow. Interesting that it's not listed in the default env variables, though. SO maybe you're right.

ianlewis: Yeah, I've seen some places online that refer to it without explicitly adding it to env so 🤷 

In any case getting it from the `github` context works.

slsa/provenance.go
ianlewis: I think I'm going to keep this as is until I know a bit more about whether we're going to want to use the `attest` API from cosign and what that's going to look like.

ianlewis:
I'll finish this up and send for review after #50 gets merged.

ianlewis:
@laurentsimon Could you take a look at this? would like to merge at some point soon.

laurentsimon (approved):
LGTM, thanks!

PR #48, opened 2022-05-17
Fix sha for distroless

ianlewis:
The previously used distroless sha seems to have had missing layers as I got a `ApplyLayer exit status 1 unexpected EOF` when attempting to execute the action.


laurentsimon (approved):
Thanks!

Issue #36, opened 2022-04-26
Merge all ecosystem builders into this repo


laurentsimon:
I was discussing with @MarkLodato about the cost of maintenance of the (possibly) many builder repos we will eventually have. 

The main reason we opted to have separate ecosystem repos was to avoid asking users to update their workflow hash/version when another ecosystem's builder was updated. However, we're starting to wonder if this is the right thing to do or not.

An alternative would be to have all builders in this repository, which *could* reduce maintenance for us. I imagine it would look like this:
- Ecosystem builders: .github/workflows/slsa3_go.yml, .github/workflows/slsa3_ko.yml, etc
- Each builder's specific code could live under ./go, ./ko, etc. Maybe overtime we would find better ways to share code between them, e.g. thru a standard codebase with "interface/template" for each ecosystem: dry-run command, download dependency command, compilation command, etc. which would allow us to abstract away the underlying compiler

Thoughts?

cc @asraa @ianlewis @joshuagl

joshuagl:
I suspect it will be easier to unbundle the ecosystem builders into separate repos at a later date, rather than bundle them into a consolidated repo once they become established, so I am in favour of having all builders in this repo in the near-term and trying to identify ways to code share.

+1 from me.

ianlewis:
Yeah, I think it will be simpler to consolidate docs etc. so there's only one place to look for info/references as well and will make it easier to consume if we have a single repo.

Docs could be something like:
We provide the following workflows:
- Generic provenance-only
- Go
- Dockerfile/Docker image
- etc.

Here's how to create your own ecosystem builder:
...

ianlewis:
Would it also make sense to include the verifier here as well? Might make it easier to manage potential breaking changes if they are both in one repo with one set of pre-submits etc.

MarkLodato:
> Would it also make sense to include the verifier here as well? Might make it easier to manage potential breaking changes if they are both in one repo with one set of pre-submits etc.

My inclination is no, that is a different use case with different code. Plus, the verifier will hopefully support more than thus this generator in the future.

laurentsimon:
I'm fine with that too. If we were to merge the verifier into this repo, I suppose we'd rename this repo to `slsa-tools` and have multiple commands like: `verify`, `generate`, etc.

laurentsimon:
[...]

PR #50, opened 2022-05-17
✨ Merge Go builder code into this repo

laurentsimon:
See https://github.com/slsa-framework/slsa-github-generator/issues/36


ianlewis (approved):
I did a quick pass but I assume this is mostly a just a copy of the existing repo.

Though, it seems like one dependency is missing from go.sum?
https://github.com/slsa-framework/slsa-github-generator/runs/6476273227?check_suite_focus=true

laurentsimon:
correct, just a copy. I'll fix the dependency problem before merging (https://github.com/slsa-framework/slsa-github-generator-go/pull/129).

ianlewis:
> correct, just a copy. I'll fix the dependency problem before merging ([slsa-framework/slsa-github-generator-go#129](https://github.com/slsa-framework/slsa-github-generator-go/pull/129)).

Ok, thanks. I did see some changes like merging the pre-submit yaml, but I'm comfortable that I'm not missing anything then.

PR #51, opened 2022-05-18
Get entryPoint for the invocation via the Github API

### LINKED ISSUE #11: Entrypoint is the workflow name

ianlewis:
Depends on #47 
Fixes #11 

Before the change the `predicate.invocation.configSource.entryPoint` can have name of the user workflow (not the reusable workflow) specified in the workflow yaml:
```yaml
"predicate": {
    "invocation": {
        "entryPoint": "SLSA provenance"
        ...
      },
...
```

After this change it's a path to the workflow:
```yaml
"predicate": {
    "invocation": {
      "configSource": {
        "entryPoint": ".github/workflows/provenance-only.yml"
        ...
      },
...
```


joshuagl (approved):
This is great, thanks. Thanks also for showing before and after fragments of the provenance.

PR #61, opened 2022-05-18
✨ [DO NOT REVIEW] Test e2e pre-submits

laurentsimon:


ianlewis:
I think pre-submits run on Drafts too so you could do that for testing in the future too :)

laurentsimon:
All seems to be working. Closing.

PR #118, opened 2022-05-27
Move builders together under internal/builders

### LINKED ISSUE #68: Refactor directory structure
I was thinking about where was best to put builders long term. A couple of options were
- internal/builders/* : builder code is generally not meant to be imported so this might make the most sense?
- cmd/* : Where provenance-only is currently. This seems like a common pattern but seems arbitrary. Not sure I like it.
- builders/* : Where go is currently. This is ok, but code is importable.

I kind of like the idea of putting main packages under a .../cmd/ structure so maybe this is also an option but maybe I'm overthinking it.
- internal/cmd/{go-builder,python-builder,slsa-builder,...}: commands
- internal/builders/{go,python,slsa,...}: supporting build code

I think the most important thing is just to be consistent with our structure.

ianlewis:
Fixes #68 

This PR should be a simple move for builders to under `internal/builders` per #68. We move them under `internal/` to make it clear and enforceable that builder code isn't imported as a library.


README.md
```diff
@@ -11,15 +11,16 @@ This repository contains the code, examples and technical design for the blog po
 - [Technical design](#technical-design)
   - [Blog post](#blog-post)
   - [Specifications](#specifications)
+
 ---
 
 ## Golang projects
 
-To generate SLSA provenance for your Golang project, follow [builders/go/README.md](builders/go/README.md)
+To generate SLSA provenance for your Golang project, follow [internal/builders/go/README.md](builders/go/README.md)
```
laurentsimon, line 19: forgot to update the link?

README.md
```diff
@@ -11,15 +11,16 @@ This repository contains the code, examples and technical design for the blog po
...
 
 ## Other projects
 
-To generate SLSA provenance for other programming languages, follow [builders/generic/README.md](builders/generic/README.md)
+To generate SLSA provenance for other programming languages, follow [internal/builders/generic/README.md](builders/generic/README.md)
```
laurentsimon, line 23: forgot to update the link?

laurentsimon (approved):
Thanks!

Issue #120, opened 2022-05-27
Update doc for Golang builder


laurentsimon:
We want to document:
- each field of the config file
- the variables supported
- how to migrate or try our builder with a co-existing gorelease workflow

PR #121, opened 2022-05-27
✨ Add section about Goreleaser in doc

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/120

### DIFF
--- a/internal/builders/go/README.md
+++ b/internal/builders/go/README.md
@@ -67,9 +67,9 @@ ldflags:
 
 If you are already using Goreleaser, you may be able to migrate to our builder using multiple config files for each build. However, this is cumbersome and we are working on supporting multiple builds in a single config file for future releases. 
 
-In the meantime, you can use both Goreleaser and this builder in the same repository. For example, you can pick one build you would like to start generating provenance for. Goreleaser and this builder can co-exist without interfering with one another. We think gradual adoption is good for project to get used to SLSA.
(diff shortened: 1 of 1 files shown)


builders/go/README.md
```diff
@@ -22,6 +23,112 @@ This document explains how to use the builder for Golang projects.
...
+binary: binary-{{ .Os }}-{{ .Arch }}
+
+# (Optional) ldflags generated dynamically in the workflow, and set as the `env` input variables in the workflow.
+ldflags:
+  - "{{ .Env.VERSION_LDFLAGS }}"
+```
+
+### Migration from goreleaser
+
+If you are already using Goreleaser, you may be able to migrate to our builder using multiple config files for each build. However, this is cumbersome and we are working on supporting multiple builds in single config file for future releases.
```
ianlewis, line 68: This seems a bit confusing to me. As a goreleaser user I'm not exactly sure what I should do. Is it worth mentioning that `.slsa-goreleaser.yml` supports a subset of goreleaser's configuration file?

ianlewis (approved):

builders/go/README.md
laurentsimon: I've updated. PTAL. Do you think I should provide an example?

ianlewis: I think this is ok for now. Thanks.

Issue #28, opened 2022-04-14
e2e testing for provenance-only workflow


ianlewis:
Related #76 

e2e tests:
- [x] Different triggers (push, workflow_dispatch, schedule, new-tag)
  - [x] push, schedule
    - https://github.com/slsa-framework/example-package/pull/27
    - https://github.com/slsa-framework/example-package/pull/24
  - [x] workflow_dispatch
    - https://github.com/slsa-framework/example-package/pull/41
  - [x] tag
    - https://github.com/slsa-framework/example-package/pull/33
- [x] Different branches (for each trigger above)
  - https://github.com/slsa-framework/example-package/pull/50
- [x] Pre-submit e2e tests
  - #122
- [x] Different arguments: vary the env variables, make sure the final provenance contains the right information
  - https://github.com/slsa-framework/example-package/pull/58 
- [x] Builder information, such as runner env (except hash)

General Tasks:
- [x] Auto-close issues when e2e tests pass
- [x] Update documentation as necessary

ianlewis:
e2e tests for the generic workflow have good coverage for now. Will add new ones as we add new features.

PR #122, opened 2022-05-30
Add pre-submit test workflow for provenance-only workflow

ianlewis:
Updates #28
Depends on #126
Blocked by #125 #158


.github/workflows/pre-submit.e2e.generic.workflow.yml
```diff
@@ -0,0 +1,29 @@
+name: pre-submit e2e generic workflow
```
laurentsimon, line 1: Note: the last part of the filename is typically for config option

internal/builders/generic/attest.go
```diff
@@ -180,3 +193,9 @@ run in the context of a Github Actions workflow.`,
 
 	return c
 }
+
+// isPresubmitTests returns true if running in pre-submit tests.
+func isPreSubmitTests() bool {
```
laurentsimon, line 198: shall we move this into the library? (It's going to be duplicated in all builders)

laurentsimon (approved):

internal/builders/generic/attest.go
ianlewis: I'd rather that it not be a public API so maybe it can go under `internal/` somewhere.

.github/workflows/pre-submit.e2e.generic.workflow.yml
ianlewis: Do you think I should leave it off? Or maybe give it another name? I guess I just put something there to have the same number of file name parts.

.github/workflows/scripts/pre-submit.e2e.generic.workflow.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# TODO(github.com/slsa-framework/slsa-github-generator/issues/129): Address base64 output format.
+# disable pipefail and use cat in order to swallow base64 errors.
+# set +o pipefail
+# ATTESTATION=$(base64 -d "$PROVENANCE" | cat)
+# set -o pipefail
+# NOTE: Provenance is not-signed and is in simple in-toto format, not DSSE.
+ATTESTATION=$(cat "$PROVENANCE")
+
+e2e_verify_predicate_subject_name "$ATTESTATION" "binary-name"
```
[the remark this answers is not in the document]
ianlewis: Yeah, SG

.github/workflows/pre-submit.e2e.generic.workflow.yml
ianlewis: If you think it should be changed I'll send a follow up PR.

PR #127, opened 2022-05-30
Return an error when OIDC server not available

ianlewis:
`net/url.Parse` does not return an error when the url is empty so `OIDCClient` gets all the way to fetching a token with a URL like `?audience=...` before failing.

This PR changes it so that creating the OIDC client fails when the `ACTIONS_ID_TOKEN_REQUEST_URL` environment variable is empty with a (hopefully) better error message.


laurentsimon (approved):
Thanks!

Issue #141, opened 2022-05-31
Rename go builder type


laurentsimon:
We're still using [slsa-github-generator-go](https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/go/pkg/provenance.go#L33): shall we change it to `slsa-github-generator`? 

Wdut @ianlewis @asraa

laurentsimon:
@MarkLodato what would you suggest?

ianlewis:
I think adding `/go` at the end would work? We probably want to differentiate the different builders.
so `https://github.com/slsa-framework/slsa-github-generator/go@v1

laurentsimon:
I'm struggling to see the difference between this type and the `path_to_reusable_workflow.yml@v1.2.3`, for example, which identifies the builder uniquely.

laurentsimon:
I'll make the change to `https://github.com/slsa-framework/slsa-github-generator/go@v1`

laurentsimon:
Repo: https://github.com/slsa-framework/example-package/tree/main
Run: https://github.com/slsa-framework/example-package/actions/runs/2419091764
Workflow name: go schedule main SLSA3 adversarial build
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/
Trigger: schedule
Branch: main
Date: Wed Jun  1 03:32:29 UTC 2022

**Tests are passing now. Closing this issue.**

Issue #140, opened 2022-05-31
Add repository_id and actor_id to provenance


laurentsimon:
There's a recent attack that manages to re-name repositories https://sockpuppets.medium.com/how-i-hacked-ctx-and-phpass-modules-656638c6ec5e, which would bypass the current provenance verification. 

We can protect agains this by recording the repository_id, see https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/about-security-hardening-with-openid-connect#understanding-the-oidc-token

This could also be added to Fulcio's certificate generation. @asraa shall we start a threat about this on sigstore repos?

For the time being, let's record it as part of the `invocation`, and provide some examples in the README how to extract it, e.g. via `ja`

laurentsimon:
Repo: https://github.com/slsa-framework/example-package/tree/main
Run: https://github.com/slsa-framework/example-package/actions/runs/2419171690
Workflow name: go schedule main SLSA3 config-noldflags
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/
Trigger: schedule
Branch: main
Date: Wed Jun  1 04:02:50 UTC 2022

laurentsimon:
Repo: https://github.com/slsa-framework/example-package/tree/main
Run: https://github.com/slsa-framework/example-package/actions/runs/2420591002
Workflow name: go schedule main SLSA3 adversarial build provenance
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/
Trigger: schedule
Branch: main
Date: Wed Jun  1 09:30:17 UTC 2022

**Tests are passing now. Closing this issue.**

Issue #130, opened 2022-05-30
Update Go builder to latest version of library


ianlewis:
Related #67

ianlewis:
@laurentsimon We should probably discuss whether it's worth it to continue to have a separate go.mod for the Go builder. I think we addressed privacy by adding it to `internal` and I'm doubtful the Go builder will have significantly more dependencies than the base library, but I'd like your thoughts.

In the future, I can definitely see us having a separate go.mod for a Docker builder (e.g. #23). For example, if we need to import Docker client that would import a huge number of transient dependencies.

laurentsimon:
Let's merge the go.mod, there does not seem to be a simple solution to this at the moment. We can re-visit later if this becomes an issue.
Thanks @ianlewis and @asraa for bringing this up

laurentsimon:
I needed this as part of https://github.com/slsa-framework/slsa-github-generator/pull/142, so I made the changes in the PR

Issue #67, opened 2022-05-19
Feature: automated update of builder's API


laurentsimon:
Our builders need to use the API of https://github.com/slsa-framework/slsa-github-generator at HEAD.
This means every time we make an update to the generic API, we need to update the builder's go.sum/mod

I don't think we can check for it in pre-submits, because the hash to update to is unknown until the commit is pushed to main.

I think we have the following options:
1. Figure out how to configure dependabot to send us PR
2. Write our own schedule-trigger workflows once every X days to create an PR.

Note that this may quickly become an endless cycle if we're not careful: each new commit will create a new HEAD, which would trigger a new PR, etc. We may want to send a PR *only* if the code of the generic API has changed.

ianlewis:
I added #66 to consider using Go workspaces. That should allow us to not have to deal with constantly updating as long as it's in the same workspace, but I haven't tried to use it much yet.

laurentsimon:
If workspaces work, certainly a better way that what I'm proposing here. Thanks for the link!

ianlewis:
BTW, is it necessary to have a separate `go.mod` for each builder? what benefit does it actually give us if we're in the same repo? Currently we don't have one for provenance-only and I'm wondering if we should add one, or if maybe removing the one for the go builder is possible way to solve the issue.

laurentsimon:
I think the original idea was to have each builder separate so that they have their own dependencies, and not the cumulative sum of all builder's dependencies. I'm not sure the dependencies for each builder will be different in practice, so it may be ok to share the the same go.mod?

Another reason was to keep them private. I think we can achieve via `internal/`, right?

ianlewis:
`internal/` should keep folks from importing it directly but it'll still get included.

PR #142, opened 2022-05-31
✨ Record repo, actor and owner IDs

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/140
closes https://github.com/slsa-framework/slsa-github-generator/issues/141
closes https://github.com/slsa-framework/slsa-github-generator/issues/130
closes https://github.com/slsa-framework/slsa-github-generator/issues/67


internal/builders/go/README.md
```diff
@@ -22,6 +22,114 @@ This document explains how to use the builder for Golang projects.
 
 To generate provenance for a golang binary, follow the steps below:
 
+### Configuration file
+
+Define a configuration file called `.slsa-goreleaser.yml` in the root of your project.
```
laurentsimon, line 27: please ignore these doc updates

github/oidc.go
```diff
@@ -50,6 +50,15 @@ type OIDCToken struct {
 
 	// JobWorkflowRef is a reference to the current job workflow.
 	JobWorkflowRef string `json:"job_workflow_ref"`
+
+	// RepositoryID is the unique repository ID.
```
laurentsimon, line 54: Note: these values may be present in the payload, but the payload's format depends on the trigger and the payload is not always present. So I think it's useful to extract it in a simple way and expose it.

internal/builders/go/pkg/provenance.go
```diff
@@ -33,7 +33,7 @@ const (
 	buildType              = "https://github.com/slsa-framework/slsa-github-generator-go@v1"
```
laurentsimon, line 33: shall we update this string? The builder's fullpath contains both the URI and the version/tag.

ianlewis: It's ok for this PR.

internal/builders/go/pkg/provenance.go
```diff
@@ -71,21 +80,35 @@ func GenerateProvenance(name, digest, command, envs, workingDir string) ([]byte,
...
+	p, err := g.Generate(ctx)
+	if err != nil {
+		return nil, err
+	}
 
-	// Identifies that this is a slsa-framework's slsa-github-generator-go' build.
-	wr.BuildType = buildType
 	// Sets the builder specific build config.
-	wr.BuildConfig = buildConfig{
+	p.Predicate.BuildConfig = buildConfig{
```
ianlewis, line 111: The idea is that BuildConfig would get set by overriding `BuildConfig` on `goProvenanceBuild`
https://github.com/slsa-framework/slsa-github-generator/blob/main/slsa/buildtype.go#L39-L40

slsa/buildtype.go
```diff
@@ -179,6 +181,41 @@ func (b *GithubActionsBuild) Invocation(ctx context.Context) (slsa.ProvenanceInv
...
+		return i, fmt.Errorf("oidc client: %w", err)
+	}
+
+	if oidcClient != nil {
+		t, err := oidcClient.Token(ctx, []string{b.Context.Repository})
+		if err != nil {
+			return i, err
+		}
+
+		if t.RepositoryID == "" {
```
ianlewis, line 198: Are these checks generic enough that they could go in `verifyToken` on the `OIDCClient`?

https://github.com/laurentsimon/slsa-github-generator/blob/22cc114346c587b0751eb61baf01ca52d8cca197/github/oidc.go#L172-L190

ianlewis (changes requested):

internal/builders/go/pkg/provenance.go
laurentsimon: gotcha. I'll create a `BuildConfig()` function.

slsa/buildtype.go
laurentsimon: You're right. Let me move them there. Thanks!

laurentsimon: done. PTAL

internal/builders/go/pkg/provenance.go
laurentsimon: done, PTAL

slsa/buildtype.go
laurentsimon: Sorry, I mis-spoke. These fields are not generic OIDC fields, they are GitHub-specific claims. I can add them in a `verifyClaims()` function instead?

laurentsimon:
friendly ping for review

asraa (approved):

laurentsimon:
@ianlewis, I need your review. Thanks

.github/workflows/scripts/e2e-verify.sh
```diff
@@ -19,13 +19,16 @@ source "./.github/workflows/scripts/e2e-utils.sh"
 
 BRANCH="main"
 
+# Note: this requires GH_TOKEN to be set in the workflows.
+THIS_FILE=$(gh api -H "Accept: application/vnd.github.v3+json" /repos/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID | jq -r '.path' | cut -d '/' -f3)
```
ianlewis: Is in necessary to add the Accept header? Is it not set by default? What does the gh command do without it?

ianlewis (approved):
LGTM

.github/workflows/scripts/e2e-verify.sh
laurentsimon: I don't think it's set by default because the GH documentation adds it in their example. I just copied the commands from the docs. But it's possible that it it is set by default and the doc is out of date

slsa/buildtype.go
ianlewis: Yeah, either way is fine. this OIDCClient is Github specific so verifying the claims as well would have been ok as well imho but `verifyClaims` works too.

PR #148, opened 2022-06-01
📖 Update the Go provenance example

laurentsimon:
See title


laurentsimon:
This is taken care of in https://github.com/slsa-framework/slsa-github-generator/pull/142, so closing. Sorry for the spam.

Issue #153, opened 2022-06-02
Use go-githubactions


ianlewis:
https://github.com/sethvargo/go-githubactions does a lot of the same stuff as the `github` package but does many things better.

For example, it gets github context values via the [built-in environment variables](
https://github.com/sethvargo/go-githubactions/blob/main/actions.go#L455-L501) rather than requiring it be serialized to json and added to a `GITHUB_CONTEXT` environment variable. This is desirable because the `github` context [doesn't seem to be available](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions) inside of an `actions.yml` so it can't be used by reusable actions.

Open issues:
- [x] Does the `GITHUB_EVENT_PATH` method work inside reusable workflows? - Yes
- [x] `GithubContext` needs a way to get the repository. (https://github.com/sethvargo/go-githubactions/pull/46)
- [x] Several other fields are missing from `GithubContext` that are present in [our version](https://github.com/ianlewis/slsa-github-generator/blob/main/github/workflow.go#L28-L48). Some of these fields we may not be using but some others may need to be added. (https://github.com/sethvargo/go-githubactions/pull/46)
  - `ref_type`
  - `base_ref`
  - `head_ref`
  - `run_attempt`

laurentsimon:
Do we need this dependency? If we can easily implement the same, I'd say we avoid adding yet another dependency to manage. In general, in the builder, we should minimize dependencies as much as possible.

Wdut?

ianlewis:
go-githubactions [depends](https://github.com/sethvargo/go-githubactions/blob/main/go.mod) only on [go-envconfig](https://github.com/sethvargo/go-envconfig/blob/main/go.mod) which has essentially no dependencies so transients should be minimal. I'd kind of prefer using it because it would require a lot of manual code to get context from env vars and it pretty much does exactly what we want.

It also does a couple other things nicely like dealing with newlines when writing output commands.

ianlewis:
It's also @sethvargo so I'm reasonably confident in the management of both packages.

ianlewis:
I'll look a bit more into how the `github` context can be used in custom actions in #164 which may make this mute. Removed from the v1 milestone since it's not really ugent.

sethvargo:
@sethvargo? Never heard of 'em 😄

ianlewis:
> @sethvargo? Never heard of 'em 😄

I may have been wrong about Seth's reliability. 😆

[...]

PR #158, opened 2022-06-02
Fix reading github context in detect-workflow

ianlewis:
Fix the detect-workflow action to get the github context from environment variables. The `github` context is not available in `actions.yml` so we can't use it as we have with reusable workflows.

We should consider using go-githubactions (#153) to get and manage the context as it will work within both actions and reusable workflows.

### DIFF
--- a/.github/actions/detect-workflow/main.go
+++ b/.github/actions/detect-workflow/main.go
@@ -33,6 +33,8 @@ type action struct {
 	client *github.OIDCClient
 }
 
+// TODO(github.com/slsa-framework/slsa-github-generator/issues/153): use go-githubactions
+
 func newAction(getenv func(string) string, c *github.OIDCClient) (*action, error) {
 	eventPath := getenv("GITHUB_EVENT_PATH")
 	if eventPath == "" {
--- a/.github/actions/detect-workflow/main.go
+++ b/.github/actions/detect-workflow/main.go
@@ -33,7 +33,7 @@ type action struct {
 	client *github.OIDCClient
 }
 
-// TODO(github.com/slsa-framework/slsa-github-generator/issues/153): use go-githubactions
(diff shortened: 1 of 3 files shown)


.github/actions/detect-workflow/main.go
```diff
@@ -16,23 +16,57 @@ package main
...
 )
 
-// getStr returns a string value from the given Event map. Values are specified
+type action struct {
+	getenv func(string) string
+	event  map[string]any
+	client *github.OIDCClient
+}
+
+// TODO(github.com/slsa-framework/slsa-github-generator/issues/153): use go-githubactions
```
laurentsimon, line 36: why do we need an external library for this?

laurentsimon (approved):

laurentsimon:
Note: instead of using the env variables, you could also define a new Action argument that defaults to the github context (and keep the code as-is). We do that for the Scorecard Action to pass the token https://github.com/ossf/scorecard-action/blob/main/action.yaml#L30-L33

laurentsimon (commented):
See my comment

.github/actions/detect-workflow/main.go
ianlewis: I'll leave out the TODO but we can discuss on #153

ianlewis:
> Note: instead of using the env variables, you could also define a new Action argument that defaults to the github context (and keep the code as-is). We do that for the Scorecard Action to pass the token https://github.com/ossf/scorecard-action/blob/main/action.yaml#L30-L33

Hmm, interesting. When I tried to use the `github` context in `action.yaml` it gave me errors saying it was unrecognized and I couldn't find any info that it's available in `action.yaml`. Maybe it's only available in `inputs` and not in `env`? I'll give in another try today.

Issue #159, opened 2022-06-02
E2E: go tag main SLSA3 config-ldflags noassets


laurentsimon:
Repo: https://github.com/slsa-framework/example-package/tree/v14.0.22
Run: https://github.com/slsa-framework/example-package/actions/runs/2426452970
Workflow name: go tag main SLSA3 config-ldflags noassets
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.go.tag.main.config-ldflags-noassets.slsa3.yml
Trigger: push
Branch: v14.0.22
Date: Thu Jun  2 06:35:25 UTC 2022

Issue #156, opened 2022-06-02
E2E: go tag main SLSA3 config-ldflags assets


laurentsimon:
Repo: https://github.com/slsa-framework/example-package/tree/v13.0.23
Run: https://github.com/slsa-framework/example-package/actions/runs/2425881289
Workflow name: go tag main SLSA3 config-ldflags assets
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.go.tag.main.config-ldflags-assets.slsa3.yml
Trigger: push
Branch: v13.0.23
Date: Thu Jun  2 03:46:29 UTC 2022

laurentsimon:
Repo: https://github.com/slsa-framework/example-package/tree/v13.0.24
Run: https://github.com/slsa-framework/example-package/actions/runs/2425932493
Workflow name: go tag main SLSA3 config-ldflags assets
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.go.tag.main.config-ldflags-assets.slsa3.yml
Trigger: push
Branch: v13.0.24
Date: Thu Jun  2 04:04:06 UTC 2022

PR #163, opened 2022-06-02
🐛 Always include context fields

laurentsimon:
The verifier expects all fields to be populated, even if they are empty. (see https://github.com/slsa-framework/slsa-verifier/blob/main/pkg/provenance.go#L596)

We did not catch this via the pre-submits because it's only triggered on push/tag events; and we can only tests these thoroughly in the e2e tests.

This PR should fix https://github.com/slsa-framework/slsa-github-generator/issues/159 and https://github.com/slsa-framework/slsa-github-generator/issues/156

### DIFF
--- a/slsa/buildtype.go
+++ b/slsa/buildtype.go
@@ -84,9 +84,9 @@ func (b *GithubActionsBuild) BuildConfig(context.Context) (interface{}, error) {
 }
 
 func addEnvKeyString(m map[string]interface{}, k string, v string) {
-	if v != "" {
-		m[k] = v
-	}
+	// Always record the value, even if it's empty. Let
+	// the consumer/verifier decide how to interpret the meaning.
+	m[k] = v
 }
 
 // getEntryPoint retrieves the path to the user workflow that initiated the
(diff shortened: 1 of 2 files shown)


slsa/buildtype.go
```diff
@@ -84,9 +84,9 @@ func (b *GithubActionsBuild) BuildConfig(context.Context) (interface{}, error) {
 }
 
 func addEnvKeyString(m map[string]interface{}, k string, v string) {
-	if v != "" {
-		m[k] = v
-	}
+	// Always record the value, even if it's empty. Let
```
asraa, line 87: If empty values should be allowed, should `i.Environment` always be set?

https://github.com/slsa-framework/slsa-github-generator/blob/b9e63b71c0f8f5a1de31d024a7ca5e9ef7ea3f0f/slsa/buildtype.go#L207-L209

laurentsimon: Will do. In practice this code would probably never be hit, since there will always have some non-empty values.

laurentsimon: Done.

asraa (approved):

PR #172, opened 2022-06-06
Rename generic workflow file

### LINKED ISSUE #168: Rename slsa2_provenance.yml
I think we can choose a scheme to name current and future workflows along the lines of `builder_<ecosystem>_<slsa lvl>.yml`.

For provenance only maybe we use something like `generator_generic_slsa2.yml`? Any other ideas?

ianlewis:
Fixes #168 

Should wait until their is consensus on naming and slsa lvl on #168 to merge.


laurentsimon (approved):
Thanks

PR #199, opened 2022-06-08
📖doc tweaks

joshuagl:
* Add the image of the cert as referenced by SPECIFICATION.md
* Move verification steps to top-level README (from internal/builders/go/README)
* Provide more detail about what SLSA requirements the workflows help to achieve (build and provenance, not source)


laurentsimon (approved):
Thanks!

PR #202, opened 2022-06-08
Fix formatting in language agnostic instructions

mihaimaruseac:
The markdown table should have just one row. However, the example is formatted in such a way that it gets 3 rows. This is just minor cleanup, with no functional change.


ianlewis (approved):
Thanks! I had tried previously to make this more readable but I guess it's strictly one line per row. :(

laurentsimon:
Thanks!

PR #214, opened 2022-06-09
Reviewdog pre-submit fixes

ianlewis:
Fixes to reviewdog pre-submit.

- Need to use the [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) event so that we can get permissions to post review comments (unfortunately requires PA token). pull_request_target runs workflow code at the PR base ref on a PR checkout so that workflow code that has not been reviewed and accepted is not run.
- yamllint config doesn't merge default config with custom config so you need extend the defaults explicitly.


.github/workflows/pre-submit.reviewdog.yml
```diff
@@ -1,5 +1,7 @@
 name: reviewdog
-on: [pull_request]
+
+"on": [pull_request_target]
```
laurentsimon, line 3: if you don't define the permissions, they are `write-all` by default. Define the exact permission you need access to, something like:
```
permissions:
  // others will default to `none`
  statuses: write
```

.github/workflows/pre-submit.reviewdog.yml
```diff
@@ -1,5 +1,7 @@
...
+
 jobs:
   golangci-lint:
     runs-on: ubuntu-latest
```
laurentsimon, line 7: if you want to checkout the code at the PR, you can use:
```
with:
        ref: ${{ github.event.pull_request.head.sha }}
```


I suggest you use workflow_run + pull_request as explained in https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

I think it'd be fine to just set the permissions in the existing workflow, since we don't run the PR's code anyway. But does not hurt to follow best practices :-)

ianlewis: Yeah, I'd like to be able to do that but I'm not sure the linters will work that way as the actions do the linting and PR commenting in one go. IIUC they would need to split up linting (untrusted pull_request) and PR commenting (trusted workflow_run) and the linting step would need to save intermediate linter results to an artifact.

As an aside, we should probably also consider setting the default permissions to the "restrictive" set here. I think this is the "Read repository contents permission" setting under "Actions > general"
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

.github/workflows/pre-submit.reviewdog.yml
ianlewis: Yeah, I guess I can just set this to empty since I'm using a PA? The linter actions seem to use the [checks API](https://docs.github.com/en/rest/checks) but for whatever reason there aren't any scopes you can add to Github Actions to allow it so you need to use a PA.

As an aside, we should probably also consider setting the default permissions to the "restrictive" set here. I think this is the "Read repository contents permission" setting under "Actions > general"
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

.github/workflows/pre-submit.reviewdog.yml
ianlewis: I decided to use the normal pull_request event and just not allow PR comments. We'll need to look at the logs for the check that fails instead of getting a friendly PR comment, but maybe that's ok for now.

.github/workflows/pre-submit.reviewdog.yml
ianlewis: Done.

laurentsimon (approved):

PR #225, opened 2022-06-09
✨ Add Version variable

laurentsimon:
Add `{{ .Version }}` which goreleaser supports

### DIFF
--- a/internal/builders/go/pkg/build.go
+++ b/internal/builders/go/pkg/build.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"syscall"
 )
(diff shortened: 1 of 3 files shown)


internal/builders/go/pkg/build.go
```diff
@@ -364,17 +378,54 @@ func (b *GoBuild) generateLdflags() (string, error) {
...
 		}
-		if !found {
-			res = v
-		}
+
 		a = append(a, res)
 	}
+
+	// Special variables.
+	for i, v := range a {
```
asraa, line 388: is there any way to factor out a `ReplaceVariable` function with this loop and use it for the above `Env.` vars too?

laurentsimon: a little tricky. I'me even thinking or removing this part of the code and just do a `ReplaceAll({{ .Value }})` instead of trying to accommodate spaces. Wdut?

laurentsimon: I'm going to remove this code and move to `ReplaceAll()`

laurentsimon: done, PTAL. We can add support for spaces later via the regex, if we want. For now it's simple.

internal/builders/go/pkg/build.go
```diff
@@ -338,43 +355,56 @@ func isAllowedEnvVariable(name string) bool {
...
-			end := strings.Index(string(v[start+len(ss):]), es)
-			if end == -1 {
-				return "", fmt.Errorf("%w: %s", errorInvalidEnvArgument, v)
-			}
 
-			name := strings.Trim(string(v[start+len(ss):start+len(ss)+end]), " ")
-			if name == "" {
-				return "", fmt.Errorf("%w: %s", errorEnvVariableNameEmpty, v)
+		// Special variables.
+		names := reVar.FindAllString(v, -1)
```
ianlewis, line 365: You could use FindAllStringSubmatch and not have to do the ReplaceAll calls below.

internal/builders/go/pkg/build.go
```diff
@@ -338,43 +355,56 @@ func isAllowedEnvVariable(name string) bool {
...
+			if name != "Version" {
+				return "", fmt.Errorf("%w: %s", errorInvalidEnvArgument, n)
 			}
 
+			version := getVersion()
+			v = strings.ReplaceAll(v, n, version)
+		}
+
+		// Dynamic env variables provided by caller.
+		names = reDyn.FindAllString(v, -1)
```
ianlewis, line 381: Same here.

ianlewis (approved):

internal/builders/go/pkg/build.go
laurentsimon: I don't think I can because the match is on `KEYWORD` in `{{ .KEYWORD }}` but we want to replace the entire string. Let me know if I'm missing something though

ianlewis: FindAllStringMatch gives you an array of full match and submatches. In the case of reVar you'll get something like `[["{{ .Env }}", "Env"] ["{{ .Version }}", "Version"], ...]`
https://go.dev/play/p/LnezjiVtBp3

laurentsimon: Thanks. SO if the match occurs multiple times, eg `[[{{ .Version }} Version] [{{ .Version }} Version]]`, I still need to iterate over them and replace them one by one, I think.
And I also need to parse each match to know if it's a full match or not (well, their position is always odd so that helps)

I think I'll keep as is for now, if you don't mind.

I would have liked to use the `re. ReplaceAllString`

ianlewis: Ok, The code works.

But just be clear I would have written it like this. You get the full value of with brackets in `n[0]` and just the name in `n[1]`.
```
names := reVar.FindAllStringSubmatch(v, -1)
for _, n := range names {
	// {{ .Version }} variable.
	if n[1] != "Version" {
		return "", fmt.Errorf("%w: %s", errorInvalidEnvArgument, n[1])
	}

	version := getVersion()
	v = strings.ReplaceAll(v, n[0], version)
}
```

It still works if there are multiple matches with the same name though a bit inefficient because ReplaceAll will have done the work on the first time through.
https://go.dev/play/p/U0jiQ2rMab1

PR #230, opened 2022-06-09
Use detect-workflow action in Go workflow

ianlewis:


.github/workflows/builder_go_slsa3.yml
```diff
@@ -69,56 +69,24 @@ on:
...
+      repository: ${{ steps.detect.outputs.repository }}
+      ref: ${{ steps.detect.outputs.ref }}
     runs-on: ubuntu-latest
     permissions:
       id-token: write # Needed to detect the current reusable repository and ref.
-      contents: read
-    outputs:
-      builder_repo: ${{ steps.detect.outputs.builder_repo }}
-      builder_ref: ${{ steps.detect.outputs.builder_ref }}
+      actions: read # Needed to detect info about the Github Actions environment.
```
ianlewis, line 79: This is not necessary either.

.github/workflows/slsa2_provenance.yml
```diff
@@ -36,14 +36,13 @@ on:
...
-      id-token: write # Needed to detect the current reusable repository and ref.
-      contents: read
-      actions: read
     outputs:
       repository: ${{ steps.detect.outputs.repository }}
       ref: ${{ steps.detect.outputs.ref }}
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write # Needed to detect the current reusable repository and ref.
+      actions: read # Needed to detect info about the Github Actions environment.
```
ianlewis, line 45: Remove actions: read

ianlewis, line 45: Remove actions: read

.github/workflows/slsa2_provenance.yml
```diff
@@ -22,7 +22,7 @@ permissions:
 #            Input and output argument definitions                #
 #                                                                 #
 ###################################################################
-on:
+"on":
```
joshuagl, line 25: is this intentional?

joshuagl (approved):
Looks great, thanks.

.github/workflows/slsa2_provenance.yml
ianlewis: It's a thing that yamllint picks up on because 'on' is a keyword in yaml that resolves to a boolean "true" value. In this case I don't thinks it really a big deal since it's a key name rather than a value, but yamllint complains nontheless.

I can maybe tweak the config we use in a later PR to allow it in this case
https://yamllint.readthedocs.io/en/stable/rules.html#module-yamllint.rules.truthy

ianlewis: https://github.com/slsa-framework/slsa-github-generator/issues/241

ianlewis:
Weird. The `Test_generateLdflags` Go unit test is failing and I didn't change any Go code...

PR #281, opened 2022-06-13
test presubmit

ianlewis:


laurentsimon:
looks like the tests is flaky. I'm encountering the same issue: sometimes it passes, sometimes not. I suspect it's due to env variables set and not unset properly.

ianlewis:
Yeah, I was testing linters and it looks like the yamllint one is broken since it passes though it finds issues. #283

PR #313, opened 2022-06-15
fix: require binary upload before binary download

asraa:

This is causing me failures: https://github.com/asraa/slsa-example/actions/runs/2502778013

[build / upload-assets](https://github.com/asraa/slsa-example/runs/6901412245?check_suite_focus=true#step:2:16)
Unable to find an artifact with the name: binary-linux-amd64

This is because binary-upload is not required for the upload-assets job.

### DIFF
--- a/.github/workflows/builder_go_slsa3.yml
+++ b/.github/workflows/builder_go_slsa3.yml
@@ -433,7 +433,7 @@ jobs:
     permissions:
       contents: write
     runs-on: ubuntu-latest
-    needs: [build, build-dry, provenance]
+    needs: [build, build-dry, binary-upload, provenance]
     if: startsWith(github.ref, 'refs/tags/') && inputs.upload-assets == true
     steps:
       # Verify binary hash.


laurentsimon (approved):
Thanks.
Did you hit this bug inn practice?

asraa:
Yep: This run https://github.com/asraa/slsa-example/actions/runs/2502778013

asraa:
> unless the testing framework has some configuration to avoid the problem.

There is a testing SetEnv that does cleanup after each test run, let me see if I can update the tests here to use that
https://cs.opensource.google/go/go/+/go1.18.3:src/testing/testing.go;l=1280

asraa:
Oh, it is between tests too :/

laurentsimon:
> > unless the testing framework has some configuration to avoid the problem.
> 
> There is a testing SetEnv that does cleanup after each test run, let me see if I can update the tests in another PR to use that https://cs.opensource.google/go/go/+/go1.18.3:src/testing/testing.go;l=1280

Awesome!

PR #314, opened 2022-06-15
test: use t.Setenv for auto cleanup 

asraa:

Fixes flakiness in tests, re-ran this 50 times locally in sequence with no failure.

### DIFF
--- a/.github/actions/detect-workflow/main_test.go
+++ b/.github/actions/detect-workflow/main_test.go
@@ -16,7 +16,6 @@ package main
 
 import (
 	"context"
-	"os"
 	"testing"
 	"time"
 
--- a/.github/actions/detect-workflow/main_test.go
+++ b/.github/actions/detect-workflow/main_test.go
@@ -105,7 +105,7 @@ func Test_action_getEventValue(t *testing.T) {
 func Test_action_getRepoRef(t *testing.T) {
 	t.Run("success", func(t *testing.T) {
 		now := time.Date(2022, 5, 3, 14, 49, 0, 0, time.UTC)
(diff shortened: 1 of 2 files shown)


asraa:
Problem fixed: I removed parallelism from the tests that manipulate arg vars.

laurentsimon:
> Problem fixed: I removed parallelism from the tests that manipulate env vars.

wait what? I thought I commented out all these `t.Parallel()`.

Thanks!

laurentsimon:
> > Problem fixed: I removed parallelism from the tests that manipulate env vars.
> 
> wait what? I thought I commented out all these `t.Parallel()`.
> 
> Thanks!

ah, so there is also a call a the top of each function, I missed this one :/

laurentsimon (approved):

laurentsimon:
> :/ weird.
> 
> ```
> 2022-06-15T15:26:23.6179049Z === RUN   Test_generateOutputFilename/filename_twice_unset_versions
> 2022-06-15T15:26:23.6180439Z     build_test.go:458:   any(
> 2022-06-15T15:26:23.6181010Z         - 	e"invalid filename: found character '/'",
> 2022-06-15T15:26:23.6181334Z           )
> ```
> 
> filename twice unset is taking the filename from the previous test ("invalid name with version",)

mhhh the `{{ .Tag }}` is evaluated by reading `GITHUB_REF_NAME`. Note that this variables is already set by default by GitHub. Maybe this is causing problems?

laurentsimon:
we may need to copy the variable like we do in scorecard, e.g. https://github.com/ossf/scorecard/blob/main/checks/branch_protection_test.go#L389

PR #316, opened 2022-06-15
chore: fix permissions for example use of generic builder

joshuagl:


laurentsimon (approved):
Thanks!

laurentsimon:
Note: @asraa is working in fixing the flaky pre-submits. We can try to re-run until then.

PR #333, opened 2022-06-17
Fix linter pre-submit

### LINKED ISSUE #283: linter pre-submits don't actually fail
The [reviewdog presubmits](https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/pre-submit.reviewdog.yml) don't actually fail even though they're set to fail on error.

Seems it's an issues with the Github actions that's been around a while? https://github.com/reviewdog/reviewdog/issues/907

ianlewis:
Fixes #283

Needs:
* #341 
* #340 
* #338 
* #466 

This fixes linter pre-submit issues and ensures that linters cause pre-submits to fail when there are errors.

reviewdog was abandoned for now because it seems to have issues where it returns a 0 error code even if the linter produces errors. It also uses of the Github API in a way that requires more permissions than we would like to give to PR code.

Instead, this PR runs the linters manually and has them output [GitHub Actions commands](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions) to output check errors.

Pros:
- Can run in a PR from a forked repo without any special permissions
- Produces comments on PR code so it's easy to see where the violation is

Cons:
- Runs on the whole repo and will match errors in files not on the PR

> For now you can check out the [files](https://github.com/slsa-framework/slsa-github-generator/pull/333/files) tab and scroll to the bottom to see examples of what the linter violations look like.


.github/workflows/pre-submit.lint.yml
```diff
@@ -0,0 +1,79 @@
...
+      - uses: actions/setup-go@b22fbbc2921299758641fab08929b4ac52b32923 # v3.2.0
+        with:
+          go-version: "1.18"
+      - env:
+          GOLANGCI_LINT_VERSION: "1.46.2"
+        run: |
+          set -euo pipefail
+
+          #Install golangci-lint
+          wget -qO- "https://github.com/golangci/golangci-lint/releases/download/v${GOLANGCI_LINT_VERSION}/golangci-lint-${GOLANGCI_LINT_VERSION}-linux-amd64.tar.gz" | tar -xzf -
```
naveensrinivasan, line 23: The issue with the `wget` approach is that there isn't a check for the SHA? Also, this approach can't keep track of updates for golangci-lint package.

The approach scorecard has taken is to have all the tools via a separate tools package and https://github.com/ossf/scorecard/blob/main/tools/tools.go and installing it in CI

naveensrinivasan: Or another approach is to use the https://github.com/golangci/golangci-lint-action

.github/workflows/pre-submit.lint.yml
```diff
@@ -0,0 +1,79 @@
...
+          golangci-lint run ./... --out-format=github-actions --timeout=5m
+
+  shellcheck:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
+      - env:
+          SHELLCHECK_VERSION: "0.8.0"
+        run: |
+          set -euo pipefail
```
naveensrinivasan, line 36: Why not use an action like this https://github.com/ludeeus/action-shellcheck?

It gets all the benefits of upgrades and maintenance.

.github/workflows/pre-submit.lint.yml
ianlewis: I added a checksum check for 1.46.2. Checksum is included here since attackers that can modify the release can modify the checksum file as well.

To be fair, most GitHub actions don't actually do this. For example the golangci-lint-action [just installs](https://github.com/golangci/golangci-lint-action/blob/master/src/install.ts#L34-L60) the tar gz and uses the binary.

.github/workflows/pre-submit.lint.yml
ianlewis: Partly for simplicity (I'm removing reviewdog because it had bugs) and partly to reduce dependencies on external actions.

.github/workflows/pre-submit.lint.yml
naveensrinivasan: Why not use the `tools` approach with a separate `go.mod`? Now the dep's can be out of sync with the version and it is hard to maintain with this approach.

.github/workflows/pre-submit.lint.yml
naveensrinivasan: I understand if there are dependencies on the `library` code. But these are tools.  What are the concerns of having dependencies on tools?

ianlewis: In this case it's not so bad because pull_request events have fewer permissions but we do need to be concerned about running 3rd party code in a context where it has permission to do things on the repo.

.github/workflows/pre-submit.lint.yml
naveensrinivasan: What I meant is that? How are we going to get updates for this dependency?

The goal is not be be out of sync.

.github/workflows/pre-submit.lint.yml
naveensrinivasan: What are the specific concerns?

Can you please be specific?

It would be helpful to understand. 

As long as the repository has the beam protection turned on and of the permissions have been set to least privileged it shouldn’t be an issue. 

Also AFAIK these actions don’t need any write permissions to the repository.

.github/workflows/pre-submit.lint.yml
ianlewis: I'm not sure what you mean by being out of sync. Out of sync with the latest version of golangci-lint?

My intention is that when we need to use a new version of golangci-lint we will update the version and the sha in this file. That's intentional as I don't think we need (or want) to update it automatically.

ianlewis: After thinking a sec, I think perhaps that you mean that if we use an action then dependabot will update the dependency for us?

I agree that that would be helpful, but I'm not sure I want to use a 3rd party action simply for that reason.

.github/workflows/pre-submit.lint.yml
ianlewis: My reasoning is:
1. The backstory for why this PR was created in the first place was because 3rd party actions (specifically reviewdog) had bugs that allowed pre-submits to pass even though there were issues on PRs. Debugging was very painful and time consuming, and I want us to have simple pre-submits that are understandable by anyone, and we able to fix bugs in pre-submit ourselves.
2. I would rather have a blanket policy of not using 3rd party actions rather than a policy which excludes the pull_request event because it's more likely to be broken later if we add new triggers to the workflow (push, or pull_request_target for example).
3. I think we need to limit the use of 3rd party dependencies when possible as a general rule. My overall view is that we need to be fairly conservative about how we use 3rd party dependencies and actively work to minimize them given that this repository hosts the reusable workflows that will be called by other projects. Esp. after we have more adoption, a compromise or incident on this repo would be a fairly high-impact event.

My benefit/analysis of using our own code:

Pros:
- presubmits are simple and are easier to debug
- 3rd party actions aren't used
  - They never get permissions on the repo (regardless of event)
  - We don't need to audit the code on the actions themselves

Cons:
- dependabot doesn't update the versions of the linters

In my view the pros outweigh the cons.

ianlewis:
@laurentsimon Can you give this a look when you get a chance?

.github/workflows/pre-submit.lint.yml
```diff
@@ -0,0 +1,64 @@
...
+
+permissions:
+  # Needed to check out the repo.
+  contents: read
+
+jobs:
+  golangci-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v3.0.2
```
laurentsimon, line 13: add the # tag=v3 for renovatebot. Same for other `uses`

ianlewis:
> LGTM, thanks. Mabye we should follow-up on what contributors need to install on their machine do run the checks locally before sending a PR?

I added some docs to the README

PR #351, opened 2022-06-18
Test : Included tests for build.go

naveensrinivasan:
- Included tests for build.go

### DIFF
--- a/internal/builders/go/pkg/build_test.go
+++ b/internal/builders/go/pkg/build_test.go
@@ -17,6 +17,7 @@ package pkg
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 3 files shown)


naveensrinivasan:
@laurentsimon  👀

internal/builders/go/pkg/build_test.go
```diff
@@ -1047,3 +1048,114 @@ func Test_generateCommand(t *testing.T) {
...
+		dry bool
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		wantErr bool
+	}{
+		{
+			name: "valid flags",
```
ianlewis, line 1069: Add whether the test is dry or not to the name

internal/builders/go/pkg/build_test.go
```diff
@@ -1047,3 +1048,114 @@ func Test_generateCommand(t *testing.T) {
...
+					Goos:    "linux",
+					Goarch:  "amd64",
+					Binary:  "binary",
+					Main:    asPointer("../builders/go/main.go"),
+					Dir:     asPointer("../builders/go"),
+					Ldflags: []string{},
+				},
+			},
+			args: args{
+				dry: true,
```
ianlewis, line 1133: Do you want to maybe add non-dry versions of these tests? or do they need to be dry for other reasons?

internal/builders/go/pkg/build_test.go
```diff
@@ -1047,3 +1048,114 @@ func Test_generateCommand(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			b := &GoBuild{
+				cfg:     tt.fields.cfg,
+				goc:     tt.fields.goc,
+				argEnv:  tt.fields.argEnv,
+				ldflags: tt.fields.ldflags,
+			}
+			defer os.Unsetenv("OUT_BINARY")
+			os.Setenv("OUTPUT_BINARY", tt.fields.cfg.Binary)
```
ianlewis, line 1147: Use t.Setenv so that the env var is cleaned up automatically.

ianlewis (approved):
Thanks!

internal/builders/go/pkg/build_test.go
```diff
@@ -1047,3 +1048,114 @@ func Test_generateCommand(t *testing.T) {
...
+		ldflags string
+	}
+	type args struct {
+		dry bool
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		args    args
+		wantErr bool
```
laurentsimon, line 1066: Could you use the expected error `error err` instead of `bool`?

asraa (approved):
Thanks! LGTM

Wondering as well (non-blocking), for dry=false, can we look for the output binary to make sure we build something?

internal/builders/go/pkg/build_test.go
naveensrinivasan: I also included the error

internal/builders/go/pkg/build_test.go
naveensrinivasan: Agreed. Thanks

internal/builders/go/pkg/build_test.go
naveensrinivasan: I took care of that. Thanks

naveensrinivasan:
> Thanks! LGTM
> 
> Wondering as well (non-blocking), for dry=false, can we look for the output binary to make sure we build something?

As I mentioned in the previous comment I will do that in the follow up PR.

internal/builders/go/pkg/build_test.go
laurentsimon: you should be able to remove the `wantErr` altogether as well?

internal/builders/go/pkg/build_test.go
```diff
@@ -1047,3 +1048,123 @@ func Test_generateCommand(t *testing.T) {
...
+			}
+			err := b.Run(tt.args.dry)
+			if err != nil != tt.wantErr {
+				t.Errorf("Run() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if tt.err != nil {
+				if err == nil {
+					t.Errorf("Run() error = nil, wantErr %v", tt.err)
+				} else if err.Error() != tt.err.Error() {
+					t.Errorf("Run() error = %v, wantErr %v", err, tt.err)
```
ianlewis, line 1165: Also to be consistent with other tests you can use `cmp.Diff(err, tt.err)` in the error message.

internal/builders/go/pkg/build_test.go
```diff
@@ -1047,3 +1048,123 @@ func Test_generateCommand(t *testing.T) {
...
+				b.goc = path
+			}
+			err := b.Run(tt.args.dry)
+			if err != nil != tt.wantErr {
+				t.Errorf("Run() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			if tt.err != nil {
+				if err == nil {
+					t.Errorf("Run() error = nil, wantErr %v", tt.err)
+				} else if err.Error() != tt.err.Error() {
```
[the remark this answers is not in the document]
naveensrinivasan: I took care of it.

internal/builders/go/pkg/build_test.go
naveensrinivasan: I addressed it.

PR #352, opened 2022-06-18
Refactor - Use interface for rekor and fulcio

naveensrinivasan:
- Refactored the implementation to accept the interface instead of the
  concrete implementation within the function.

- With this change `GenerateProvenance` can be tested with mocks. At the
  moment there aren't any unit tests for `GenerateProvenance`.

### DIFF
--- a/internal/builders/go/pkg/provenance.go
+++ b/internal/builders/go/pkg/provenance.go
@@ -20,11 +20,12 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/slsa-framework/slsa-github-generator/signing"
+
 	intoto "github.com/in-toto/in-toto-golang/in_toto"
 	slsa02 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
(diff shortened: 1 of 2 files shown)


naveensrinivasan:
@laurentsimon 👀

internal/builders/go/pkg/provenance.go
```diff
@@ -18,13 +18,13 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
+	"github.com/slsa-framework/slsa-github-generator/signing"
```
ianlewis, line 21: Move this import to the 3p block below. This block is stdlib packages.

ianlewis (approved):
Thanks!

laurentsimon (approved):
Thanks!

internal/builders/go/pkg/provenance.go
naveensrinivasan: Thanks! Took care of it.

PR #354, opened 2022-06-20
minor fixes, updates

chipzoller:
* Fixes broken links and a typo
* Clarifies the expected format of the decoded value of the input `base64-subjects`.

Signed-off-by: Chip Zoller <chipzoller@gmail.com>

### DIFF
--- a/internal/builders/generic/README.md
+++ b/internal/builders/generic/README.md
@@ -161,7 +161,7 @@ jobs:
 
 ### Workflow Inputs
 
-The [builder workflow](https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/generator_generic_slsa3.yml) accepts the following inputs:
+The [generic workflow](https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/generator_generic_slsa3.yml) accepts the following inputs:
 
 | Name              | Required | Description                                                                                                                        |
 | ----------------- | -------- | ---------------------------------------------------------------------------------------------------------------------------------- |
--- a/internal/builders/generic/README.md
+++ b/internal/builders/generic/README.md
@@ -64,6 +64,8 @@ output:
 $ sha256sum artifact1 artifact2 ... | base64 -w0
 ```
 
+This workflow expects the base64 encoded value for the subjects to decode to a string conforming to the expected output of the above command. Specifically, the decoded output is expected to be comprised of a hash value followed by a space followed by the artifact name.
+
 After you have encoded your digest, add a new job to call the reusable workflow.
 
 ```yaml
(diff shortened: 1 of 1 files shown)


internal/builders/generic/README.md
```diff
@@ -64,6 +64,8 @@ output:
 $ sha256sum artifact1 artifact2 ... | base64 -w0
 ```
 
+This workflow expects the base64 encoded value for the subjects to decode to a string conforming to the expected output of the above command. Specifically, the decoded output is expected to be comprised of a hash value followed by a space followed by the artifact name.
```
ianlewis, line 67: Technically the above command outputs the base64 value so the value of `base64-subjects` wouldn't wouldn't decode to that. It would decode to the value of the `sha256sum artifact1 artifact2 ...` part. Rather than "above command". I'd just say sha256sum command.

Also consider referring to the input name directly so users know exactly what we're talking about.
e.g. "This workflow expects the value of `base64-subjects to ..."
```suggestion
This workflow expects the `base64-subjects` input to decode to a string conforming to the expected output of the `sha256sum` command. Specifically, the decoded output is expected to be comprised of a hash value followed by a space followed by the artifact name.
```

ianlewis: Additionally, consider adding this as a comment in the example yaml below.
```yaml
with:
  # This workflow expects the value of base64-subjects to decode to a string conforming to the
  # expected output of the sha256sum command...
  base64-subjects: ...

File: internal/builders/generic/README.md [hunk 2]
```diff
@@ -159,19 +161,15 @@ jobs:
 
 ### Workflow Inputs
 
-The builder workflow
-[.github/workflows/generator_generic_slsa3.yml](.github/workflows/generator_generic_slsa3.yml) accepts
-the following inputs:
+The [builder workflow](https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/generator_generic_slsa3.yml) accepts the following inputs:
```
ianlewis, line 164: Let's change "builder workflow" to "generic workflow" here since that's the wording we're using elsewhere.
```suggestion
The [generic workflow](https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/generator_generic_slsa3.yml) accepts the following inputs:
```

internal/builders/generic/README.md
```diff
@@ -159,19 +161,15 @@ jobs:
...
 | ----------------- | -------- | ---------------------------------------------------------------------------------------------------------------------------------- |
-| `base64-subjects` | yes      | Artifacts for which to generate provenance, formatted the same as the output of sha256sum (SHA256 NAME\n[...]) and base64 encoded. |
+| `base64-subjects` | yes      | Artifact(s) for which to generate provenance, formatted the same as the output of sha256sum (SHA256 NAME\n[...]) and base64 encoded. The encoded value should decode to, for example: `90f3f7d6c862883ab9d856563a81ea6466eb1123b55bff11198b4ed0030cac86  foo.zip` |
 
 ### Workflow Outputs
 
-The builder workflow
-[.github/workflows/generator_generic_slsa3.yml](.github/workflows/generator_generic_slsa3.yml)
-produces the following outputs:
+The [builder workflow](https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/generator_generic_slsa3.yml) produces the following outputs:
```
ianlewis, line 172:
```suggestion
The [generic workflow](https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/generator_generic_slsa3.yml) produces the following outputs:
```

ianlewis (approved):
Thanks for this!

ianlewis:
> I've committed those suggestions. One other to consider is to clearly state that this assumes a Linux runner, perhaps one of a specific distribution. When I was testing this on my Mac locally, it does not have the program `sha256sum` and so calling out that this is a program for Linux which may deliver different results from other sha programs would be helpful.

Yep that's a reasonable suggestion.

Issue #177, opened 2022-06-07
Generic provenance: faster builds


laurentsimon:
We use pre-built binaries for the Go builder to speed up compilation.

Let's try doing the same for the generic provenance generators.

PR #356, opened 2022-06-20
Add release configs for the generic generator

ianlewis:
@laurentsimon 

Following the pattern of the Go builder and [RELEASE.md](https://github.com/slsa-framework/slsa-github-generator/blob/main/RELEASE.md) docs.

My understanding is that after I add these configs, I can cut a release which will include the generic generator's binary, and then I can use that to speed up the workflow.

Updates #177


laurentsimon (approved):
Thanks!

PR #361, opened 2022-06-20
Update provenance_format.md

ianlewis:
- Add section headers to group fields.
- Remove version specific heading. We should add "since v1.1" or similar tags to new fields (or "deprecated/removed in v1.1" for removed fields)?

### DIFF
--- a/PROVENANCE_FORMAT.md
+++ b/PROVENANCE_FORMAT.md
@@ -1,19 +1,39 @@
-Each of the projects utilize the same base GitHub workflow SLSA provenance. The common fields of the SLSA provenance predicate attested to are below.
+# Provenance format
 
-## Since v1.0.0
+The provenance format for each workflow conforms to the [SLSA v0.2 provenance
+format](https://slsa.dev/provenance/v0.2). Each of the projects populate the
+provenance with the same base information based on GitHub Actions. The common
+fields of the SLSA provenance predicate attested to are below.
 
-`BuildType`: This is the URI for the particular provenance builder, for example, the go or generic builder. 
+## Common Fields.
+
+### Builder
+
+`Builder.Id`: The builder ID refers to the fully qualified URI of the trusted builder's workflow and its reference.
+
 ```json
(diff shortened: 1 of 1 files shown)


PROVENANCE_FORMAT.md
```diff
@@ -1,19 +1,39 @@
...
-`Invocation.configSource`: This describes the calling workflow's source and the entrypoint of the build.
+
+In SLSA the [buildType](https://slsa.dev/provenance/v0.2#buildType) defines how
+many other fields are formatted including all of those below.
+
+### Invocation
+
+`invocation`: Identifies the event that kicked off the build. This describes the workflow run and includes GitHub workflow event information, entrypoint, and parameters from trigger events.
+
+`invocation.configSource`: This describes the calling workflow's source and the entrypoint of the build.
```
joshuagl, line 35: Would it make sense to tie this to the example by including the explicit field names in the description?
```suggestion
`invocation.configSource`: This describes the calling workflow's source (`uri` and `digest`) and the `entryPoint` of the build.
```

PROVENANCE_FORMAT.md
```diff
@@ -87,7 +113,11 @@ Each of the projects utilize the same base GitHub workflow SLSA provenance. The
     }
 ]
 ```
-`Metadata`: Other properties of the build, including `BuildInvocationID` as the unique `RunID` and `RunAttempt`. 
+
+### Metadata
+
+`metadata`: Other properties of the build, including `BuildInvocationID` as the unique `RunID` and `RunAttempt`.
```
joshuagl, line 119: Could we clarify this? Perhaps something like:
```suggestion
`metadata`: Other properties of the build, including `BuildInvocationID`, which is the unique `RunID` and `RunAttempt` separated by a '-'.
```

joshuagl (commented):
I made a couple of suggestions, curious to read what you think.

ianlewis:
> I made a couple of suggestions, curious to read what you think.

SG

joshuagl (approved):

Issue #376, opened 2022-06-21
[bug] Generic workflow fails on repositories that don't use Go


behnazh-w:
**Describe the bug**
If I understand correctly the generic workflow is meant to generate provenance for any language as long as the encoded hash digest of the built artifact is provided as input. However, when I try it on a repository that doesn't use Go (e.g., uses Python) it fails because it tries to download the Go module dependencies.

https://github.com/slsa-framework/slsa-github-generator/blob/668a33920550fa6983e6c850f69da89d28b057eb/.github/workflows/generator_generic_slsa3.yml#L71-L88

Why does the workflow need to download Go modules of the target repository? Please let me know if I'm missing something.

**To Reproduce**
Run the generic workflow on a repo that doen't use Go.

**Screenshots**
![image](https://user-images.githubusercontent.com/27933054/174910006-1da781ac-7738-455d-a2fe-d6e364045d09.png)

laurentsimon:
The code you linked is to build the builder, not the project.
Here's an e2e test that succeeds https://github.com/slsa-framework/example-package/blob/main/.github/workflows/e2e.generic-bazel.push.main.default.slsa3.yml#L27-L66

Maybe you're using a trigger we have not tested yet. Can you link to a workflow you use to call the generic workflow?

behnazh-w:
[Here](https://github.com/behnazh-w/slsa-examples/blob/main/.github/workflows/pull-request.yaml) is the example workflow that fails.

> The code you linked is to build the builder, not the project.

Note that my example repo is resolved [here](https://github.com/behnazh-w/slsa-examples/runs/7012632489?check_suite_focus=true#step:2:4) at the checkout step.

ianlewis:
Hi again @behnazh-w! Thanks for the bug report and for the links. Those were super helpful for figuring out what happened.

This is caused by the fact that we don't support the `pull_request` trigger for Github Actions. See the [docs](https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/generic#supported-triggers) for the generic workflow. Supporting it is kind of difficult for permissions related issues. Relevant GitHub issues are #358 and #131.

What happened in this case is that our [detect-workflow](https://github.com/slsa-framework/slsa-github-generator/tree/main/.github/actions/detect-workflow) action returned empty strings for the repository and ref which then got fed into the `checkout` action and it checked out your repository (since the repo on the github context is the default).
[...]

PR #382, opened 2022-06-23
Return an error if repo/ref not detected

ianlewis:
Related #376

I'm not sure why but it seems that in `pull_requests` (even though we don't really support them, and possibly other triggers) sometimes the `detect-workflow` action returns empty strings in the repository and/or ref outputs without producing an error. If fed directly into a later `checkout` action this checks out the repository in the `github` context (not the slsa-github-generator) repo.

This PR causes the `detect-workflow` action to return an error if no repository or ref are found.


laurentsimon (approved):
Thans!

PR #393, opened 2022-06-24
golanglintci settings

naveensrinivasan:
These settings are based on https://raw.githubusercontent.com/ossf/scorecard/main/.golangci.yml


.golangci.yml
```diff
@@ -0,0 +1,168 @@
...
+  # Maximum count of issues with the same text.
+  # Set to 0 to disable.
+  # Default: 3
+  max-same-issues: 0
+  new-from-rev: ""
+  # Fix found issues (if it's supported by the linter).
+  fix: true
+skip-files:
+  - cron/data/request.pb.go # autogenerated
+linters:
```
ianlewis, line 25: Since this is a Go 1.18 project I'm not sure that many of the linters will work. Have you tested this? currently we don't have a pre-submit that will run using this file.

.golangci.yml
```diff
@@ -0,0 +1,168 @@
...
+  # Set to 0 to disable.
+  # Default: 50
+  max-issues-per-linter: 0
+  # Maximum count of issues with the same text.
+  # Set to 0 to disable.
+  # Default: 3
+  max-same-issues: 0
+  new-from-rev: ""
+  # Fix found issues (if it's supported by the linter).
+  fix: true
```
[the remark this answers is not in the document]
naveensrinivasan: Agreed. Thanks

.golangci.yml
naveensrinivasan: I ran the check and removed deprecated linters. The next step would be to include the pre-submit that will use the file.

internal/builders/generic/attest.go
```diff
@@ -128,7 +128,7 @@ func getFile(path string) (io.Writer, error) {
 	if path == "-" {
 		return os.Stdout, nil
 	}
-	return os.OpenFile(path, os.O_WRONLY|os.O_CREATE, 0600)
+	return os.OpenFile(path, os.O_WRONLY|os.O_CREATE, 0o600)
```
naveensrinivasan, line 131: `gofmt` https://github.com/golang/go/issues/37453#issuecomment-791998830

.golangci.yml
ianlewis: Do you want to keep them here and comment them out? Maybe link to the issue? https://github.com/golangci/golangci-lint/issues/2649

I think we'll want to add them back eventually.

internal/builders/generic/attest.go
naveensrinivasan: I don't see it as an issue.

.golangci.yml
naveensrinivasan: Included the comments

.golangci.yml
```diff
@@ -0,0 +1,161 @@
...
+skip-files:
+  - cron/data/request.pb.go # autogenerated
```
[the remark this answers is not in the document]
naveensrinivasan: Agree. Thanks

ianlewis (approved):

PR #396, opened 2022-06-24
Refactor Println to use a stream

naveensrinivasan:
The default Println writes to stdout which makes it harder to test.

Refactored to write to `stdout` but configured with a variable which can
be updated for tests.

### DIFF
--- a/internal/builders/go/main.go
+++ b/internal/builders/go/main.go
@@ -55,7 +55,7 @@ func runBuild(dry bool, configFile, evalEnvs string) error {
 	if err != nil {
 		return err
 	}
-	fmt.Println(cfg)
+	fmt.Fprintln(pkg.Output, cfg)
 
 	gobuild := pkg.GoBuildNew(goc, cfg)
 
@@ -142,7 +142,7 @@ func main() {
 		check(err)
 
 	default:
-		fmt.Println("expected 'build' or 'provenance' subcommands")
+		fmt.Fprintln(pkg.Output, "expected 'build' or 'provenance' subcommands")
 		os.Exit(1)
 	}
 }
--- a/internal/builders/go/pkg/build.go
+++ b/internal/builders/go/pkg/build.go
@@ -52,6 +52,8 @@ var allowedEnvVariablePrefix = map[string]bool{
 	"GO": true, "CGO_": true,
 }
 
+var Output = os.Stdout
+
 type GoBuild struct {
 	cfg *GoReleaserConfig
 	goc string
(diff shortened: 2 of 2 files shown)


laurentsimon (commented):
I'm curious if an io writer instead would be better?

naveensrinivasan:
> I'm curious if an io writer instead would be better?

Isn't stdout an io.writer? The goal was to replace the `var Output = os.Stdout` variable with a byte buffer to test things. 

If you have better suggestions let me know. 👍

laurentsimon:
> > I'm curious if an io writer instead would be better?
> 
> Isn't stdout an io.writer? The goal was to replace the `var Output = os.Stdout` variable with a byte buffer to test things.
> 
> If you have better suggestions let me know. 👍

Sorry, I meant use an io.Writer as a field to the structure, and let the caller decide where to store the "logs". 
Let me know what you think.

@ianlewis feel free to chime in

internal/builders/go/pkg/build.go
```diff
@@ -52,6 +52,8 @@ var allowedEnvVariablePrefix = map[string]bool{
 	"GO": true, "CGO_": true,
 }
 
+var Output = os.Stdout
```
ianlewis, line 55: I'm thinking this should be a field on `GoBuild` rather than a package level global. That way all GoBuild instances aren't bound to using a single writer. It should also allow tests to run in parallel.

internal/builders/go/main.go
```diff
@@ -55,7 +55,7 @@ func runBuild(dry bool, configFile, evalEnvs string) error {
 	if err != nil {
 		return err
 	}
-	fmt.Println(cfg)
+	fmt.Fprintln(pkg.Output, cfg)
 
 	gobuild := pkg.GoBuildNew(goc, cfg)
```
ianlewis, line 60: Suggestion: Create a `GoBuildConfig` struct or something similar that has a `GoCompilerPath` string and `Output` writer fields and pass it here in lieu of `goc`.

```go
buildConfig := &GoBuildConfig{
   GoCompilerPath: goc,
   Output: os.Stdout,
}
gobuild := pkg.GoBuildNew(buildConfig, cfg)
```

The goal is to encapsulate using the `os.Stdout` writer to `main()` and avoid the use of global package level variables. I think that would make the code cleaner, easier to test, and easier to modify in the future.

internal/builders/go/pkg/build.go
```diff
@@ -155,10 +157,10 @@ func (b *GoBuild) Run(dry bool) error {
 		return err
 	}
 
-	fmt.Println("dir", dir)
-	fmt.Println("binary", binary)
-	fmt.Println("command", command)
-	fmt.Println("env", envs)
+	fmt.Fprintf(Output, "dir %s \n", dir)
```
ianlewis, line 160: Alternative suggestion: Use a `*log.Logger` for writing to stdout instead of `os.Writer` directly. Loggers are maybe a bit easier to work with. But maybe `os.Writer` is better if we expect users to parse the builder's output.

@laurentsimon Was the intention here that users could parse the builder's output? or is this just informational logging?

laurentsimon: just informational  logging

ianlewis:
I'm going to close this as it's pretty stale. Feel free to rebase and send a new PR if we still need this

PR #397, opened 2022-06-24
Tests - Included additional unit tests.

naveensrinivasan:


.github/actions/detect-workflow/main_test.go
```diff
@@ -172,3 +174,67 @@ func Test_action_getRepoRef(t *testing.T) {
...
+			},
+			wantErr: true,
+			want:    nil,
+		},
+		{
+			name: "success",
+			args: args{
+				getenv: func(k string) string {
+					// create a temp file with key:value
+					f, err := ioutil.TempFile("", "")
```
ianlewis, line 204: use `os.CreateTemp`

.github/actions/detect-workflow/main_test.go
```diff
@@ -172,3 +174,67 @@ func Test_action_getRepoRef(t *testing.T) {
...
+					return ""
+				},
+			},
+			wantErr: true,
+			want:    nil,
+		},
+		{
+			name: "success",
+			args: args{
+				getenv: func(k string) string {
```
ianlewis, line 202: Check the value of `k`. You can return "" or panic if you get something unexpected.

.github/actions/detect-workflow/main_test.go
```diff
@@ -172,3 +174,67 @@ func Test_action_getRepoRef(t *testing.T) {
...
+				event: map[string]any{
+					"test": "hoge",
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := newAction(tt.args.getenv, tt.args.c)
+			if (err != nil) != tt.wantErr {
```
ianlewis, line 228: You can return a specific error value and check for that instead.

.github/actions/detect-workflow/main_test.go
```diff
@@ -16,6 +16,8 @@ package main
 
 import (
 	"context"
+	"github.com/google/go-cmp/cmp"
```
ianlewis, line 19: This should go in the lower block of imports below the stdlib imports
Usually it's something like

```
import (
  <stdlib packages>

  <3rd party packages>

  <packages from this project/repo>
)
```

.github/actions/detect-workflow/main_test.go
naveensrinivasan: Is there a specific reason for the request for change?  `ioutil.TempFile` does create a `temp file`

A code review should provide the reason and justification. https://google.github.io/eng-practices/review/reviewer/comments.html#why

The "use os.CreateTemp" doesn't come out as a request, especially for someone who is trying to contribute. https://google.github.io/eng-practices/review/reviewer/comments.html#courtesy

ianlewis: Apologies. I would like this to be `os.CreateTemp`. The reason is that `ioutil.TempFile` just calls `os.CreateTemp` anyway as of Go 1.17 (see [godoc](https://pkg.go.dev/io/ioutil#TempFile)). Use of the `ioutil` package seems generally discouraged in the Go community.

naveensrinivasan: Thank you! Appreciate the explanation. I will change it.

ianlewis: np. Thanks for following up!

ianlewis: Also, I commented on this PR in a bit of haste. It was not my intention to be discourteous. Sorry about that! I'll try to explain my thoughts in more detail next time.

.github/actions/detect-workflow/main_test.go
naveensrinivasan: `getenv` : It is replacing/mocking the `os.Getenv'.  I don't see a value in validating the `k` input. Could you please explain your reasoning? This is just a test func.

.github/actions/detect-workflow/main_test.go
naveensrinivasan: This would mean refactoring all the errors are wrapping them up into custom errors. There are 3 errors within this small function.

```
func newAction(getenv func(string) string, c *github.OIDCClient) (*action, error) {
	eventPath := getenv("GITHUB_EVENT_PATH")
	if eventPath == "" {
		return nil, errors.New("GITHUB_EVENT_PATH not set")
	}

	payload, err := os.ReadFile(eventPath)
	if err != nil {
		return nil, err
	}

	var event map[string]any
	if err := json.Unmarshal(payload, &event); err != nil {
		return nil, err
	}

	return &action{
		getenv: getenv,
		event:  event,
		client: c,
	}, nil
}
```

This is out of scope for this PR and should be another PR for addressing [errors](https://github.com/polyfloyd/go-errorlint)

ianlewis: I see what you mean. I'm ok with addressing it in a later PR, but I thought that since you are improving the tests with this PR that it would be in scope to also modify the function in a way that makes it easier to test. As it is, your test will pass if any error is returned when I think the intention is to check for errors in a specific case. wdut?

Just to elaborate and be more specific about my thoughts, I think that since you are only testing the one case so far, you can address this by modifying the just the first returned error without changing the others. And, since it's not a dynamic error you can return a static error value.

```golang
var errEventPathNotSet = errors.New("GITHUB_EVENT_PATH not set")

func newAction(getenv func(string) string, c *github.OIDCClient) (*action, error) {
    eventPath := getenv("GITHUB_EVENT_PATH")
    if eventPath == "" {
        return nil, errEventPathNotSet
    }
    ...
}
```

and in the test you can set `wantErr` to an error type and check it with `errors.Is`.

```
tests := []struct {
	name    string
	args    args
	want    *action
	wantErr error
}{
    {
        ...
        wantErr: errEventPathNotSet,
     },
     ...
 }

if !errors.Is(err, tt.wantErr) {
    t.Fatalf(...)
}
```

I believe this is generally how we've tried to write tests that check errors throughout the project, using `errors.Is` or `errors.As`. You can see some examples (albiet a bit more complicated) in [oidc.go](https://github.com/slsa-framework/slsa-github-generator/blob/d98cbc192096122d80204d5d602958530fac6162/github/oidc.go#L64-L87) and [oidc_test.go](https://github.com/slsa-framework/slsa-github-generator/blob/d98cbc192096122d80204d5d602958530fac6162/github/oidc_test.go#L62-L64). Another example is [build.go](https://github.com/slsa-framework/slsa-github-generator/blob/d98cbc192096122d80204d5d602958530fac6162/internal/builders/go/pkg/build.go#L27-L33) and [build_test.go](https://github.com/slsa-framework/slsa-github-generator/blob/d98cbc192096122d80204d5d602958530fac6162/internal/builders/go/pkg/build_test.go#L98-L100).

We also have an internal [errors](https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/errors) package to help support this.

I hope that makes my thoughts clearer!

naveensrinivasan:
I’ll  close this PR as it is going out of scope.

PR #403, opened 2022-06-27
📖 Update doc

laurentsimon:
Explain the difference between generic generator and builders
Add example to generate provenance in combination with Goreleaser

Later, we can add an example for docker images and other ecosystems there.

closes https://github.com/slsa-framework/slsa-github-generator/issues/405

### DIFF
--- a/internal/builders/generic/README.md
+++ b/internal/builders/generic/README.md
@@ -288,7 +288,7 @@ This section explains how to generate non-forgeable SLSA provenance with existin
 If you use [Goreleaser](https://github.com/goreleaser/goreleaser-action) to generate your build, you can easily
 generate SLSA3 provenance as follows:
 
-```
+```yaml
 jobs:
   goreleaser:
     # Step 1: declare an output for the Goreleaser job.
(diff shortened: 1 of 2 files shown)


internal/builders/generic/README.md
```diff
@@ -270,3 +277,70 @@ generated as an [in-toto](https://in-toto.io/) statement with a SLSA predicate.
...
 }
 ```
+
+## Integration With Other Build Systems
+
+This section explains how to generate non-forgeable SLSA provenance with existing build systems.
+
+### Provenance for Goreleaser
+
+If you use [Goreleaser](https://github.com/goreleaser/goreleaser-action) to generate your build, you can easily
```
ianlewis, line 287: nit: GoReleaser seems to refer to itself with a capital 'R' in their docs.

ianlewis (approved):

internal/builders/generic/README.md
laurentsimon: good catch, thanks

PR #406, opened 2022-06-27
renovate bot setting to pin actions to a full length commit SHA

naveensrinivasan:
- https://docs.renovatebot.com/modules/manager/github-actions/#additional-information

- Pinned actions by SHA https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies

>Pin actions to a full length commit SHA

>Pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release.
> Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository,
> as they would need to generate a SHA-1 collision for a valid Git object payload.

- https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions
- https://github.com/renovatebot/.github/blob/b0c3aa85ef2bb242580f20b02b380ca532b4ce17/default.json#L13


renovate.json
```diff
@@ -1,7 +1,8 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
-    "config:base"
+    "config:base",
+    "helpers:pinGitHubActionDigests"
```
laurentsimon, line 5: is this needed? I thought renovatebot would detect our config and update? (assuming we already have all action pinned by hash)

naveensrinivasan: No, it doesn't default, AFAIK. We would have to opt in.

laurentsimon: The [PR](https://github.com/slsa-framework/slsa-github-generator/pull/401) that enabled renovatebot understood the pin that were already used, e.g., `Upgrade actions/checkout to 2541b1294d2704b0964813337f33b291d3f8596b`

IIIC, the additional option is used to migrate version pins to hash pins, or have both tag/hash (dockerfiles), e.g., https://docs.renovatebot.com/configuration-options/#pindigest

naveensrinivasan: OK. Updated to include the pindigests.

laurentsimon (approved):
Thanks!

PR #413, opened 2022-06-28
Create a release as part of generic workflow

### LINKED ISSUE #402: [feature] Upload provenance option for generic generator
It may be useful to provide an option to upload the provenance as release assets automatically. For example, if users use an existing solution to build-and-release, they may expect the same from the provenance generator.

ianlewis:
Fixes #402

### DIFF
--- a/.github/workflows/generator_generic_slsa3.yml
+++ b/.github/workflows/generator_generic_slsa3.yml
@@ -28,11 +28,6 @@ env:
   VERIFIER_RELEASE_BINARY_SHA256: 60c91c9d5b9a059e37ac46da316f20c81da335b5d00e1f74d03dd50f819694bd
   VERIFIER_RELEASE: v0.0.1
 
-###################################################################
-#                                                                 #
-#            Input and output argument definitions                #
-#                                                                 #
-###################################################################
 on:
   workflow_call:
     inputs:
@@ -40,8 +35,13 @@ on:
         description: "Artifacts for which to generate provenance, formatted the same as the output of sha256sum (SHA256 NAME\\n[...]) and base64 encoded."
         required: true
         type: string
+      upload-to-release:
+        description: "If true provenance is uploaded to a GitHub release for new tags."
+        required: false
+        type: boolean
+        default: true
       compile-generator:
-        description: "Build the generator from source. This increases build time by ~2mn."
+        description: "Build the generator from source. This increases build time by ~2m."
         required: false
         type: boolean
         default: false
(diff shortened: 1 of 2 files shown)


.github/workflows/generator_generic_slsa3.yml
```diff
@@ -28,29 +28,34 @@ env:
...
     inputs:
       base64-subjects:
         description: "Artifacts for which to generate provenance, formatted the same as the output of sha256sum (SHA256 NAME\\n[...]) and base64 encoded."
         required: true
         type: string
+      upload-to-release:
+        description: "If true provenance is uploaded to a GitHub release for new tags."
+        required: false
+        type: boolean
+        default: true
```
ianlewis, line 42: I'm not entirely convinced that we should make this the default behavior since users of the generic workflow will need to release their artifact binaries as well. It works ok if you [use the same softprops/action-gh-release](https://github.com/ianlewis/actions-test/blob/d3be73012eec38ca11d180b7e6b4330a61e0728d/.github/workflows/provenance-only.yml#L60-L81) but even then if users provide any [inputs](https://github.com/softprops/action-gh-release#inputs) to it like `name` or `tag_name` it's not going to work as expected.

.github/workflows/generator_generic_slsa3.yml
```diff
@@ -141,3 +154,40 @@ jobs:
...
+        run: |
+          set -euo pipefail
+
+          echo "hash of provenance $UNTRUSTED_PROVENANCE_NAME should be $UNTRUSTED_PROVENANCE_HASH"
+
+          COMPUTED_HASH=$(sha256sum "$UNTRUSTED_PROVENANCE_NAME" | awk '{print $1}')
+          echo "provenance hash computed is $COMPUTED_HASH"
+
+          # Compare hashes. Explicit exit to be safe.
+          echo "$UNTRUSTED_PROVENANCE_HASH $UNTRUSTED_PROVENANCE_NAME" | sha256sum --strict --check --status || exit -2
```
ianlewis, line 186: @laurentsimon I copied this from the Go builder's code but I wonder, is this strictly necessary? The provenance is signed so if it's modified at any point it would be detected by the verifier. A malicious user could modify it at any point after the workflow completes as well so I don't think this provides any additional security.

laurentsimon: You're correct. It's not strictly needed.

laurentsimon (approved):
Thanks

.github/workflows/generator_generic_slsa3.yml
ianlewis: Ok, so just to be clear I'm going to leave it as is, but I'm not necessarily going to write an e2e test to test it.

PR #432, opened 2022-06-29
fix: verifier sha refs

asraa:

Not sure why this didn't trigger an error before?

Maybe wait until https://github.com/slsa-framework/slsa-verifier/pull/114/ is approved

### DIFF
--- a/.github/workflows/builder_go_slsa3.yml
+++ b/.github/workflows/builder_go_slsa3.yml
@@ -29,7 +29,7 @@ env:
   # release binaries when the compile-builder input is false.
   VERIFIER_REPOSITORY: slsa-framework/slsa-verifier
   VERIFIER_RELEASE_BINARY: slsa-verifier-linux-amd64
-  VERIFIER_RELEASE_BINARY_SHA256: 14360688de2d294e9cda7b9074ab7dcf02d5c38f2874f6c95d4ad46e300c3e53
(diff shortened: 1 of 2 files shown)


laurentsimon:
> Signed-off-by: Asra Ali [asraa@google.com](mailto:asraa@google.com)
> 
> Not sure why this didn't trigger an error before?

The latest v1.1.1 release of the builder uses version v1.1.0 of the verifier, see https://github.com/slsa-framework/slsa-github-generator/blob/d995948e8d53cc639c0d3ef69db31dbc243519c4/.github/workflows/builder_go_slsa3.yml#L30-L31

We updated the version in the workflow to v1.1.1 after that in https://github.com/slsa-framework/slsa-github-generator/pull/411, but we don't test this path in e2e tests (always compile the builder). We would have caught this problem during the release process when we test the pre-compiled verifier.

> Maybe wait until [slsa-framework/slsa-verifier#114](https://github.com/slsa-framework/slsa-verifier/pull/114) is approved

I LGTMED'd it.

The has is `f92fc4e571949c796d7709bb3f0814a733124b0155e484fad095b5ca68b4cb21` same as the one in this PR.

laurentsimon (approved):
Thanks!

ianlewis:
Oh, duh. I totally forgot to do this. Thanks!

laurentsimon:
We could have pre-submit to catch these? (`git diff ... | grep`)?

Issue #26, opened 2022-04-12
Feature: GitHub action for common tasks


laurentsimon:
Across reusable workflow, we need to:
1. Checkout the repo at the right ref
2. Build the builder
3. Download-verify the builder

We need a common GitHub action to perform these task, to avoid copy/paste of the same code.

The actions can be hosted in this repo and called by reusable workflows:
```
uses: slsa-framework/slsa-github-generator/actions/checkout-builder
...
uses: slsa-framework/slsa-github-generator/actions/builder-builder
  with:
    ref: needs.detect-env.outputs.ref
    repo: needs.detect-env.outputs.repo

...

# this will verify the hash
uses: slsa-framework/slsa-github-generator/actions/download-builder
  with:
    name: ...
    sha256: ${{ needs.builder.outputs.sha256 }}
```

asraa:
I LOVE THIS IDEA. This would make it foolproof to setup trusted builders. Which is more important than distributing builders :)

ianlewis:
Is this achieving the same thing as #22 ?

laurentsimon:
No. https://github.com/slsa-framework/slsa-github-generator/issues/22 suggest we have an action to install a CLI.
This issue suggest we create actions for the building blocks used within the reusable workflow. 

I think https://github.com/slsa-framework/slsa-github-generator/issues/22 is a subset of this issue, since it's an action to help install the builder as a CLI (and it would need to perform the checkout as well)

ianlewis:
@laurentsimon @MarkLodato I'm looking to move the [detect-workflow action](https://github.com/slsa-framework/slsa-github-generator/tree/main/.github/actions/detect-workflow) to a separate repo so it's easier to invoke in our workflows and we can avoid the chicken-and-egg problem we get when actions are in the same repo.

Are you able to create new repos under slsa-framework and could you create it for me? Since our reusable workflows will be the primary users I think it's best created here instead of another org.

MarkLodato:
For my information, could you describe the chicken-and-egg problem? I would have thought that having it in the same repo is easier because then the whole workflow is hermetic.

laurentsimon:
The action needs to be called by the re-usable workflow to determine at which ref it must checkout itself, but to call itself it needs to know a ref to itself. That's the chicken-and-egg problem. @ianlewis confirmed?

I think this means any other Action we create will go into this separate repo, e.g. the Action that compares hashes, the Action that upload assets, etc

[...]

PR #435, opened 2022-06-29
feat: Action to checkout and setup Go env

laurentsimon:

This PR defines an action to checkout and setup the Go env. 

Note that the pre-submit will fail because the action points to the main branch, but it's not there yet. So we may need to ignore the pre-submit and merge. Suggestions welcome ,though

This Action will be used to build the `generate-builder` Action as well.


.github/actions/checkout-go/action.yml
```diff
@@ -0,0 +1,62 @@
...
+    # set by the `actions/checkout`. Even if it is highlely unlikely the `actions/checkout` team
+    # will set a default value in the future, we want to be sure it does not affect us if they do.
+    # This is why we use 2 blocks to call the `actions/checkout`:
+    #   1. if inputs.ref != ''
+    #   2. if inputs.ref == ''
+    - name: Checkout the repository with user ref
+      if: inputs.ref != ''
+      uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # v2.3.4
+      with:
+        fetch-depth: 1
```
ianlewis, line 44: I know we've always done this but do we really need to set fetch-depth? 1 is the default AFAICT

internal/builders/generic/README.md
```diff
@@ -343,4 +343,5 @@ jobs:
     uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.1.1
     with:
       base64-subjects: "${{ needs.goreleaser.outputs.hashes }}"
-```
\ No newline at end of file
+      upload-to-release: true
```
ianlewis, line 346: nit: The default is still true and I think the name will change anyway so you can remove this.
```suggestion
```

ianlewis (approved):

.github/actions/checkout-go/action.yml
laurentsimon: you're right, let me remove it.

PR #436, opened 2022-06-29
feat: Action to compute a sha256 of a file

laurentsimon:

This PR defines an action to compute the sha256 of a file.

I will send follow-up PRs to use these it as part of artifact verification and for other uses in the re-usable workflows. We may later convert this code to TS or another programming language.


.github/actions/compute-sha256/action.yml
```diff
@@ -0,0 +1,24 @@
...
+runs:
+  using: "composite"
+  steps:
+    - name: Compute the sha256
+      id: compute
+      env:
+        UNTRUSTED_PATH: "${{ inputs.path }}"
+      run: |
+
+        set -euo pipefail
```
ianlewis, line 21: I don't think we need this if we're just going to call out to another script. We can just do

```yaml
run: ./.github/actions/compute-sha256.sh
```

ianlewis:
How much value do we get out of just computing the sha256? I thought maybe our action would combine the generate sha and upload steps. And another action could combine the download and verify steps. I'm curious what you had in mind?

.github/actions/compute-sha256/action.yml
laurentsimon: yeah Ive been thinking about this too. I'm not yet 100% sure whether this action will be replaced/removed in the end. I was thinking that may be we will replace the script code with TS. 
But maybe it's better to use the script in this PR instead. Shall I do that?

laurentsimon: Let's go with the script option then.

laurentsimon: Well, let me know how you feel before I proceed :-)

ianlewis: I was just referring to the `set -euo pipefail` part.

ianlewis (approved):

.github/actions/compute-sha256/action.yml
laurentsimon: sorry, I mis-understood. Gotcha.

laurentsimon: done, thanks

PR #437, opened 2022-06-29
feat: Action to download and verify an artifact by hash

laurentsimon:
depends on https://github.com/slsa-framework/slsa-github-generator/pull/436 which needs to be merged first

This is part of https://github.com/slsa-framework/slsa-github-generator/issues/26

This PR defines an Action to download and verify an artifact by hash. We can later replace the script with a better programming language.

PR #438, opened 2022-06-29
feat: Action to build the builder

laurentsimon:
depends on https://github.com/slsa-framework/slsa-github-generator/pull/437 which needs to be merged first

This PR defines an Action to download and verify an artifact by hash. We can later replace the script with a better programming language.


.github/actions/generate-builder/generate-builder.sh
```diff
@@ -0,0 +1,27 @@
...
+
+else
+    echo "Fetching the builder with ref: $BUILDER_REF"
+
+    .github/actions/generate-builder/builder-fetch.sh
+
+    mv "$BUILDER_RELEASE_BINARY" "$BUILDER_BINARY"
+fi
+
+chmod u+x "$BUILDER_BINARY"
```
ianlewis, line 27: nit: we should try to keep newlines at the end of files

.github/actions/generate-builder/action.yml
```diff
@@ -0,0 +1,61 @@
...
+        COMPILE_BUILDER: "${{ inputs.compile-builder }}"
+        BUILDER_REF: "${{ inputs.ref }}"
+        BUILDER_DIR: "${{ inputs.directory }}"
+        BUILDER_BINARY: "${{ inputs.binary }}"
+        # Needed for the gh CLI used in builder-fetch.sh.
+        GH_TOKEN: "${{ inputs.token }}"
+      run: |
+        set -euo pipefail
+
+        ./.github/actions/generate-builder/generate-builder.sh
```
ianlewis, line 55:
```suggestion
      run: ./.github/actions/generate-builder/generate-builder.sh
```

.github/actions/compute-sha256/compute-sha256.sh
```diff
@@ -0,0 +1,11 @@
...
+set -euo pipefail
+
+if [[ ! -f "$UNTRUSTED_PATH" ]]; then
+    echo "File $UNTRUSTED_PATH not present"
+    exit 5
+fi
+
+DIGEST=$(sha256sum "$UNTRUSTED_PATH" | awk '{print $1}')
+
+echo "::set-output name=sha256::$DIGEST"
```
ianlewis, line 11: nit: while we're still using bash scripts, let's try to get in the habit of using lowercase for local unexported variables so they can be more easily differentiated.

```suggestion
digest=$(sha256sum "$UNTRUSTED_PATH" | awk '{print $1}')

echo "::set-output name=sha256::$digest"
```

ianlewis (approved):

.github/actions/generate-builder/generate-builder.sh
laurentsimon: I need to install a plugin to catch these, thanks

.github/actions/generate-builder/builder-fetch.sh
```diff
@@ -52,7 +52,7 @@ if [[ "$BUILDER_TAG" == "$(echo -n "$BUILDER_TAG" | grep -P '^[a-f\d]{40}$')" ]]
 fi
 
 if [[ "$BUILDER_TAG" != "$(echo -n "$BUILDER_TAG" | grep -P '^v\d*(\.([\d]{1,})){0,2}$')" ]]; then
-    echo "Invalid ref: $BUILDER_TAG"
+    echo "Invalid version: $BUILDER_TAG. Expected versin of the form vX.Y.Z"
```
ianlewis, line 55:
```suggestion
    echo "Invalid version: $BUILDER_TAG. Expected version of the form vX.Y.Z"
```

Issue #452, opened 2022-06-30
[bug] Builders broken


laurentsimon:
I broke the builders in my series of PRs where I introduced the component Actions.
Main problem is that we don't checkout the re-usable workflow before calling the internal Actions, so the Actions calls are failing.

I think we need to pin it to a good sha using `slsa-framework/.github/actions/name@sha1`, like we do for the detect-ref Action.

An alternative solution would be to checkout the re-usable workflow, but that's cumbersome and may complicate the code in order to avoid interference between the re-usable workflow/project we're building.. so seems prone to mistakes

Will tackle this tmr.

ianlewis:
I took a stab at this in #453. Unfortunately it looks like we still need to do the checkout even if we use `slsa-framework/.github/actions/name@sha1`

I think `detect-workflow` works without a checkout simply because we use a `Dockerfile` for that action.

ianlewis:
@laurentsimon 

#454 Updates the actions themselves to reference each other by sha. We need to disable pre-submit requirements for this to go in.

#453 This restores the pre-submits but workflows need to be updated with the right git sha after #454 is merged.

I added notes to each PR so that hopefully everything is clear.

laurentsimon:
Thank you @ianlewis , much appreciated.
I'll send a PR documenting how to handle Action

laurentsimon:
Actively working non fixing this. Will be a series of PRs

laurentsimon:
so:
- [ ] Ony inline scripts
- [ ] Fix typos
- [ ] Create a secure-upload-artifact Action
- [ ] Create README.md for Action development
- [ ] Create pre-submit to validate the README guidelines

laurentsimon:
This done. Closing

PR #454, opened 2022-06-30
Update actions ref

ianlewis:
Updates #452 

Updates references in re-usable actions to point to a repository sha and not require a checkout of the repository.

We will need to reference re-usable actions in our workflow pre-submits after this PR is merged so workflow pre-submits are disabled by the PR temporarily. They will be re-enabled in subsequent PR #453

NOTE: The "verify" required status check was removed from the repository branch protection rules to allow us to commit this PR. This will also be restored for #453


ianlewis:
> ooi, why all the single quote to double quote changes?

I'll try to limit those in the future. It's mostly because I run [prettier](https://prettier.io/) on save to keep yaml formatted consistently.

PR #459, opened 2022-06-30
presubmit: align verifier versions

asraa:

Might have overcomplicated this, but checks that the builders verifier version/tag is aligned with the latest in verifier's SHA256SUM.md


laurentsimon (approved):
LGTM

.github/workflows/pre-submit.units.yml
```diff
@@ -58,8 +59,68 @@ jobs:
...
           go mod vendor
 
-          # Build.
+          # Build Go builder.
           go build -mod=vendor
 
-          # Tests.
-          go test -mod=vendor -v ./...
+          # Generic tests.
+          cd ../generic
```
ianlewis, line 66: nit: rather than 'cd' into the directory you can build using the package name from the repository root.

```
go build -mod=vendor github.com/slsa-framework/slsa-github-generator/internal/builders/go
go build -mod=vendor github.com/slsa-framework/slsa-github-generator/internal/builders/generic
```

ianlewis (approved):

.github/workflows/pre-submit.units.yml
laurentsimon: IIUC, one limitation of this is that you need to `go mod vendor` all dependencies of all projects at once. Since builders are "independent" projects, `cd` into each independently allows us to only vendor the dependencies needed for the project. (It's a bit slower since we do it twice in the tests)

So Ideally we should probably do for each builder:
```
cd internal/builders/go
go mod vendor
go build -mod=vendor
cd -
```

and mirror this in our pre-submits.

Wdut?

Most dependencies come from the main API/cosign for our project, so it's not a huge gain. But it mirrors what we do in the builders. (during the series of patches)

ianlewis: It's not super important so I'm ok with the code as it is.

I have a couple of questions though (mostly for my own benefit):
1. I'm curious what advantage running`go mod vendor` gives us over just doing a go build and having it install the dependencies in the workspace (`GOPATH`). It seems to me it's just a difference in directory location since we aren't vendoring dependencies in the repo but maybe I'm missing something?
2. I don't think there is any advantage anymore since both the Go builder and generic generator are using the same go.mod ? Am I missing something there?

ianlewis (approved):

.github/workflows/pre-submit.units.yml
laurentsimon: ho you''re right, they use the same go/mod now. Ignore my comment then :)

ianlewis: > > It's not super important so I'm ok with the code as it is.
> > I have a couple of questions though (mostly for my own benefit):
> > 
> > 1. I'm curious what advantage running`go mod vendor` gives us over just doing a go build and having it install the dependencies in the workspace (`GOPATH`). It seems to me it's just a difference in directory location since we aren't vendoring dependencies in the repo but maybe I'm missing something?
> 
> `go vendor` is to satisfy `hermetic build` SLSA requirement. We could perform this step in another VM, copy the dependencies over to the build VM and remove network access, if we wanted to go the extra mile.

Hmm, Since `go vendor` is pulling the dependencies when it's run it's just unclear what the advantage is here. But it's not important as either way works and we can discuss later if it's really an issue.

PR #464, opened 2022-06-30
fix: update ref for all internal Actions

laurentsimon:
Fix part of https://github.com/slsa-framework/slsa-github-generator/issues/452#issuecomment-1171585190

This PR updates the refs for the Actions

### DIFF
--- a/.github/workflows/pre-submit.e2e.generic.default.yml
+++ b/.github/workflows/pre-submit.e2e.generic.default.yml
@@ -1,9 +1,8 @@
 name: pre-submit e2e generic default
 
 on:
-  # TODO(github.com/slsa-framework/slsa-github-generator/issues/452): restore pre-submit
-  # pull_request:
-  #   branches: [main]
+  pull_request:
+    branches: [main]
   workflow_dispatch:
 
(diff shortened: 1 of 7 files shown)


.github/workflows/pre-submit.e2e.generic.default.yml
```diff
@@ -1,7 +1,6 @@
 name: pre-submit e2e generic default
 
 on:
-  # TODO(github.com/slsa-framework/slsa-github-generator/issues/452): restore pre-submit
   # pull_request:
   #   branches: [main]
```
ianlewis, line 5: Did you not actually want to re-enable the pre-submit?

ianlewis (approved):

.github/workflows/pre-submit.e2e.generic.default.yml
laurentsimon: it's not working. I need to see whether that's because the `uses: ./.github/workflows/name` is always using main branch or not. I thought it was not, but I'm not sure anymore.
Will re-enable once I've validated everything works.

Issue #409, opened 2022-06-27
[feature] Generic container workflow


ianlewis:
A container workflow similar to the generic workflow that allows users to build a container themselves in the user workflow and then generate provenance for it.

- [x] Add container generator workflow (#471)
- [x] Pre-submit test (#604)
- [x] e2e tests (#603)
- [x] Documentation
  - [x] Basic usage (#471)
  - [x] #389 
  - [x] #491

PR #471, opened 2022-07-01
Add a generic container workflow

ianlewis:
Updates #409 

Adds a "generic" container workflow. Users build the container in the user workflow and call the reusable container workflow which generates provenance using `cosign attest` and uploads it to the container registry.

`cosign attest` [uploads additional metadata](https://github.com/sigstore/cosign/issues/2027) to the registry which is used during the verification process by `cosign verify-attestation`. The generic workflow cannot support uploading this additional metadata because our generic workflow is not container aware and users cannot upload the attestation produced by the workflow with `cosign attach attestation` because it does not upload it.

This PR adds a new workflow that is container aware. Users build and push their image to a registry. This workflow will use the generic generator to produce a SLSA predicate and then run the `cosign attest` command to sign and upload the provenance.

Support for `slsa-verifier` is being implemented in https://github.com/slsa-framework/slsa-verifier/issues/92

Depends on:
- #470 
- #472


internal/builders/container/README.md
```diff
@@ -0,0 +1,261 @@
...
+    permissions:
+      actions: read
+      id-token: write
+      contents: read
+      packages: write
+    if: startsWith(github.ref, 'refs/tags/')
+    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@main
+    with:
+      # XXX: should this use the sha instead of the tag?
+      image: ${{ needs.build.outputs.tag }}
```
ianlewis, line 143: I suppose that the image tag could be overwritten between the time it's pushed to the registry and the time `cosign attest` is run. I wonder if the sha should be used instead? I would think given how the `cosign` CLI works that this would be a common issue, but I don't really see any guidance in the [cosign docs](https://docs.sigstore.dev/cosign/attestation). The examples just seem to run on `latest`.

ianlewis: There was a discussion related to this on the slack channel. The TL;DR is to sign digests and not tags.
https://sigstore.slack.com/archives/C01PZKDL4DP/p1656926576991169

There is also an issue to discourage signing tags.
https://github.com/sigstore/cosign/issues/2047

.github/workflows/generator_container_slsa3.yml
```diff
@@ -0,0 +1,146 @@
...
+        with:
+          repository: "${{ needs.detect-env.outputs.repository }}"
+          ref: "${{ needs.detect-env.outputs.ref }}"
+          go-version: 1.18
+          binary: "${{ env.BUILDER_BINARY }}"
+          compile-builder: "${{ inputs.compile-generator }}"
+          # NOTE: We are using the generic generator.
+          directory: "${{ env.BUILDER_DIR }}/generic"
+
+      # Note: here we need packages permissions
```
joshuagl, line 106: Nit: should we move this comment up to the permissions section, like the one for `actions`?

internal/builders/container/README.md
```diff
@@ -0,0 +1,257 @@
...
+The [container workflow](https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/generator_container_slsa3.yml) accepts the following inputs:
+
+Inputs:
+
+| Name                | Required | Description                                                                                         |
+| ------------------- | -------- | --------------------------------------------------------------------------------------------------- |
+| `image`             | yes      | The OCI image name. This must not include a tag or digest.                                          |
+| `digest`            | yes      | The OCI image digest. The image digest of the form '<algorithm>:<digest>' (e.g. 'sha256:abcdef...') |
+| `registry-username` | yes      | Username to log into the container registry.                                                        |
+| `compile-generator` | false    | Build the generator from source. This increases build time by ~2m.                                  |
```
joshuagl, line 172: Nit:
```suggestion
| `compile-generator` | false    | Whether to build the generator from source. This increases build time by ~2m.                                  |
```

joshuagl (approved):
This looks great, thanks

.github/workflows/generator_container_slsa3.yml
ianlewis: Done.

internal/builders/container/README.md
```diff
@@ -0,0 +1,255 @@
...
+  needs: [build]
+  permissions:
+    actions: read # for detecting the Github Actions environment.
+    id-token: write # for creating OCID tokens for signing.
+    packages: write # for uploading attestations.
+  if: startsWith(github.ref, 'refs/tags/')
+  # TODO(https://github.com/slsa-framework/slsa-github-generator/issues/492): Use a tagged release once we have one.
+  uses: slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@main
+  with:
+    image: ${{ needs.build.outputs.tag }}
```
laurentsimon, line 65: this should be the hash/digest, as shown in https://github.com/actions/starter-workflows/blob/main/ci/docker-publish.yml#L93
You're correct about TOCTOU attack, in particular if the user pushed the image to a remote registry

internal/builders/container/README.md
```diff
@@ -0,0 +1,255 @@
...
+    registry-username: ${{ github.actor }}
+    # TODO(https://github.com/slsa-framework/slsa-github-generator/issues/492): Remove after GA release.
+    compile-generator: true
```
laurentsimon, line 68: remove `compile-generator` from example?

internal/builders/container/README.md
```diff
@@ -0,0 +1,255 @@
...
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@e551b19e49efd4e98792db7592c17c09b89db8d8 # v3.0.0
+        id: build
+        with:
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
+
+      - name: Output image and digest
```
laurentsimon, line 113: You may be able to directly use the results of build-and-push, like https://github.com/actions/starter-workflows/blob/main/ci/docker-publish.yml#L93

internal/builders/container/README.md
```diff
@@ -0,0 +1,255 @@
...
+### Workflow Inputs
+
+The [container workflow](https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/generator_container_slsa3.yml) accepts the following inputs:
+
+Inputs:
+
+| Name                | Required | Description                                                                                         |
+| ------------------- | -------- | --------------------------------------------------------------------------------------------------- |
+| `image`             | yes      | The OCI image name. This must not include a tag or digest.                                          |
+| `digest`            | yes      | The OCI image digest. The image digest of the form '<algorithm>:<digest>' (e.g. 'sha256:abcdef...') |
```
laurentsimon, line 168: the README's example workflow uses `image` to pass both to image + digest, but the doc here says otherwise.

internal/builders/container/README.md
ianlewis: Ah, yeah, I forgot to update this. will do.

internal/builders/container/README.md
ianlewis: Yeah, I forgot to update the example after I added the digest input

internal/builders/container/README.md
laurentsimon: Until this is GA, the example won't pass verification, so it's fine to remove it, I'd say.
But I see what you mean. SO long as we don't forget to update the doc after GA, that's fine with me.

internal/builders/container/README.md
ianlewis: Doesn't this have the same script expansion/injection problem as inputs? I think it's probably better to put all inputs/outputs through env vars (esp. if from the user, or an 3p action).

or am I misunderstanding your comment?

laurentsimon: it depends whether the underlying Action uses scripts or not. Passing variables via env variables won't protect against a buggy Action, since the value becomes the Action input. If the Action is vulnerable, it's still overall vulnerable.

I was suggesting doing:

```
uses: slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@main
    with:
      image: "${IMAGE_REGISTRY}/${IMAGE_NAME}@${{ steps.build.outputs.digest }}"
```

to simplify users' workflow.

Your solution has the benefit of encouraging users to use env variables, which is good too :)

internal/builders/container/README.md
ianlewis: Done.

internal/builders/container/README.md
ianlewis: I updated this. In reality the workflow takes the image and digest as separate inputs to make it clear that the hash is required.

Issue #469, opened 2022-07-01
[e2e]: go release main config-ldflags-assets-tag slsa3 


laurentsimon:
Repo: https://github.com/slsa-framework/example-package/tree/v23.0.11
Run: https://github.com/slsa-framework/example-package/actions/runs/2594418946
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.go.release.main.config-ldflags-assets-tag.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.go.release.main.config-ldflags-assets-tag.slsa3.yml
Trigger: release
Branch: v23.0.11
Date: Fri Jul  1 05:02:08 UTC 2022

laurentsimon:
Repo: https://github.com/slsa-framework/example-package/tree/v23.0.12
Run: https://github.com/slsa-framework/example-package/actions/runs/2600344027
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.go.release.main.config-ldflags-assets-tag.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.go.release.main.config-ldflags-assets-tag.slsa3.yml
Trigger: release
Branch: v23.0.12
Date: Sat Jul  2 04:42:02 UTC 2022

laurentsimon:
Repo: https://github.com/slsa-framework/example-package/tree/v23.0.13
Run: https://github.com/slsa-framework/example-package/actions/runs/2603849884
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.go.release.main.config-ldflags-assets-tag.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.go.release.main.config-ldflags-assets-tag.slsa3.yml
Trigger: release
Branch: v23.0.13
Date: Sun Jul  3 04:34:07 UTC 2022

laurentsimon:
Repo: https://github.com/slsa-framework/example-package/tree/v23.0.14
Run: https://github.com/slsa-framework/example-package/actions/runs/2607755340
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.go.release.main.config-ldflags-assets-tag.slsa3.yml
[...]

PR #490, opened 2022-07-04
Add builder as dependency of upload-assets

ianlewis:
Updates #469 

Fixes "hashes do not match" error when running the Go workflow. This adds the `builder` job as a dependency for `upload-assets` so that the outputs from `builder` can be used.


laurentsimon (commented):
Should the fix be to replace:
```
sha256: "${{ needs.builder.outputs.go-builder-sha256 }}"
```
by:

```
sha256: "${{ needs.build.outputs.go-binary-sha256 }}"
```
instead?

ianlewis:
> Should the fix be to replace:
> 
> ```
> sha256: "${{ needs.builder.outputs.go-builder-sha256 }}"
> ```
> 
> by:
> 
> ```
> sha256: "${{ needs.build.outputs.go-binary-sha256 }}"
> ```
> 
> instead?

Yeah, you're right.

laurentsimon (approved):

PR #494, opened 2022-07-05
Add additional guidance for users who don't pin by tag

### LINKED ISSUE #431: [bug] Unhelpful error message when generator can not be fetched in generic workflow (due to using a non-released ref)
**Describe the bug**
#86 introduced fetching the generator binary to speed up build times, however the fetch will always fail when not using a released version of the reusable action. That is, if I use the action `@main` and don't explicity opt in to compiling the generator my workflow runs fail with:
> Fetching the generator with ref: refs/heads/main
Invalid ref: refs/heads/main

This makes sense once you dig into it, but was surprising even as a user with some familiarity with how the project works.

**To Reproduce**

1. Create an action which includes the reusable workflow with a non-pinned ref: `uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@main`
2. Inherit the default option of not compiling the generator `compile-generator: false`
3. Run the action
> Fetching the generator with ref: refs/heads/main
Invalid ref: refs/heads/main

**Expected behavior**
A meaningful error message which helps me to resolve the problem.

joshuagl:
Adds some documentation on why slsa-verifier requires we pin by tag to the README and points users to that section of the README when fetching the builder fails.

Fixes #431

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,11 @@ This is a pre-release only and we will have the official release in July 2022.
 
 To verify the provenance, use the [github.com/slsa-framework/slsa-verifier](https://github.com/slsa-framework/slsa-verifier) project.
 
+**Note**: At present the GitHub Actions provided in this repository as builders and generators MUST be referenced by tag in order for the slsa-verifier to be able to verify the ref of the trusted builder/generator's reusable workflow.
+
+This is contrary to the [best practice](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) which recommends referencing by digest.
+The desire to be able to verify reusable workflows pinned by hash, and the reasons for the current status, are tracked as [Issue #12](https://github.com/slsa-framework/slsa-verifier/issues/12) in the slsa-verifier project.
+
 ### Installation
 
 To install the verifier, see [slsa-framework/slsa-verifier#installation](https://github.com/slsa-framework/slsa-verifier#installation).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -55,6 +55,11 @@ This is a pre-release only and we will have the official release in July 2022.
 
 To verify the provenance, use the [github.com/slsa-framework/slsa-verifier](https://github.com/slsa-framework/slsa-verifier) project.
 
+**Note**: At present the GitHub Actions provided in this repository as builders and generators MUST be referenced by tag in order for the slsa-verifier to be able to verify the ref of the trusted builder/generator's reusable workflow.
```
ianlewis, line 58:
```suggestion
**Note**: At present the GitHub Actions provided in this repository as builders and generators **MUST** be referenced by tag in order for the `slsa-verifier` to be able to verify the ref of the trusted builder/generator's reusable workflow.
```

README.md
```diff
@@ -55,6 +55,11 @@ This is a pre-release only and we will have the official release in July 2022.
...
+
+This is contrary to the [best practice](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) which recommends referencing by digest.
```
ianlewis, line 60:
```suggestion
This is contrary to the [best practice](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) which recommends referencing by digest, but intentional due to limits in GitHub Actions.
```

ianlewis (approved):
Thanks! This is a lot more clear for users.

PR #496, opened 2022-07-07
Clean up

naveensrinivasan:
- Fixed some typos
- Some code cleanup


internal/builders/go/README.md
```diff
@@ -101,16 +101,16 @@ We think gradual adoption is good for projects to get used to SLSA.
...
-| `{{ .Tag }}`         | `$GITHUB_REF_NAME` (on `release` and `push` new tag triggers) or `$(git describe --tags --always --dirty \| cut -c2-)`           | `v1.2.3-alpha+b2a98088`                    |
-| `{{ .Major }}`       | `$(git describe --tags --always --dirty \| cut -d '.' -f1 \| cut -c2-)`                                                          | `1`                                        |
-| `{{ .Minor }}`       | `$(git describe --tags --always --dirty \| cut -d '.' -f2`                                                                       | `2`                                        |
-| `{{ .Patch }}`       | `$(git describe --tags --always --dirty \| cut -d '.' -f3 \| cut -d '-' -f1 \| cut -d '+' -f1`                                   | `3`                                        |
+| Name                 | Value                                                                                                      | Example                                    |
+|----------------------|------------------------------------------------------------------------------------------------------------|--------------------------------------------|
```
laurentsimon, line 105: those changes seem to be breaking the table

laurentsimon (commented):
Thanks, so many typos we have :-) . The table seems broken.

internal/builders/go/README.md
naveensrinivasan: Took care of it. Thanks

laurentsimon (approved):
LGTM, thanks.

In the "Migration from goreleaser", do you think adding `yaml` to the snippet would help? Right now it shows all black and white.

naveensrinivasan:
> LGTM, thanks.
> 
> In the "Migration from goreleaser", do you think adding `yaml` to the snippet would help? Right now it shows all black and white.

I think it will.

.github/workflows/generator_generic_slsa3.yml
```diff
@@ -107,7 +107,7 @@ jobs:
       - name: Create and sign provenance
         id: sign-prov
         shell: bash
-        # NOTE: Inputs and github context are set to environment variables in
+        # NOTE: Inputs and GitHub context are set to environment variables in
```
ianlewis, line 110:
```suggestion
        # NOTE: Inputs and `github` context are set to environment variables in
```
The casing is intentional here as the `github` context is usually referred to in lowercase in GitHub docs.
https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

I suggest adding ` to it to make it clearer what's being referred to.

github/workflow.go
```diff
@@ -25,8 +25,8 @@ const (
 	githubContextEnvKey = "GITHUB_CONTEXT"
 )
 
-// WorkflowContext is the 'github' context given to workflows that contains
-// information about the Github Actions workflow run.
+// WorkflowContext is the 'GitHub' context given to workflows that contains
```
ianlewis, line 28:
```suggestion
// WorkflowContext is the `github` context given to workflows that contains
```

go.sum
```diff
@@ -41,7 +41,6 @@ cloud.google.com/go v0.97.0/go.mod h1:GF7l59pYBVlXQIBLx3a761cZ41F9bBH3JUlihCt2Ud
 cloud.google.com/go v0.98.0/go.mod h1:ua6Ush4NALrHk5QXDWnjvZHN93OuF0HfuEPq9I1X0cM=
```
ianlewis, line 41: Is it possible to update `go.sum` in a separate PR? I'm just trying to be careful with dependencies so I think we should update `go.mod` and `go.sum` on PRs where it's clearer why they're being updated.

ianlewis (approved):
Thanks for this. We should be better about typos :(

go.sum
naveensrinivasan: Thanks, I can do that.

naveensrinivasan: I rolled back the change

slsa/buildtype.go
```diff
@@ -90,9 +90,9 @@ func addEnvKeyString(m map[string]interface{}, k string, v string) {
 }
 
 // getEntryPoint retrieves the path to the user workflow that initiated the
-// workflow run. The `github` context contains the path in `workflow` but it
+// workflow run. The `GitHub` context contains the path in `workflow` but it
```
ianlewis, line 93:
```suggestion
// workflow run. The `github` context contains the path in `workflow` but it
```

ianlewis (approved):
Looks good!

Issue #501, opened 2022-07-07
[feature] Remove env variable duplication


laurentsimon:
Variables in https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/builder_go_slsa3.yml#L25-L32
are duplicated in each re-usable workflow. 

We need to have them shared in a single location, e.g. in the generate-builder Action.

PR #502, opened 2022-07-07
feat: Share env variables for pre-built builder binaries

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/501

### DIFF
--- a/.github/actions/generate-builder/action.yml
+++ b/.github/actions/generate-builder/action.yml
@@ -45,13 +45,13 @@ runs:
       env:
         # Builder.
         BUILDER_REPOSITORY: slsa-framework/slsa-github-generator  # The repository to download the pre-built builder binary from.
-        BUILDER_RELEASE_BINARY: "${{ inputs.binary }}"            # The name of the pre-built binary in the release assets. This is also use as the final binary name when compiling the builder from source.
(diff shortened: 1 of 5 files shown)


naveensrinivasan (commented):
This is great! 

Including some docs would be helpful! Especially when all of this isn't compiled code( shell script and YAML)

It would be hard for someone to come and contribute.

For example, how is the VERIFIER_* everything used? What is the order of things that have to be changed to update the VERIFIER_*?  What happens when there is a bug in VERIFIFER_ which code should be fixed.

If there are existing docs, then reference to those docs within the PR would be helpful. 
 
 This is to avoid arcane knowledge.

laurentsimon:
> This is great!
> 
> Including some docs would be helpful! Especially when all of this isn't compiled code( shell script and YAML)
> 

I've added comment next to each variable. Does it help?

naveensrinivasan:
> > This is great!
> > Including some docs would be helpful! Especially when all of this isn't compiled code( shell script and YAML)
> 
> I've added comment next to each variable. Does it help?

Thank you! It helps!

laurentsimon:
friendly ping for review

ianlewis (approved):
LGTM

Issue #500, opened 2022-07-07
[feature] randomize binary names


laurentsimon:
We currently use hard-coded names for builder binaries and project binaries we compile. This has some downsides:
1. If a user calls our builder multiple times in the same workflow, they will need to make the build sequential or it will likely fail. (I've seen one user do this already)
2. If user uses multiple builder calls in the same workflow, the workflow will race to download the binary and it could be corrupt and the run will fail
3. If the generated binary is the same as a file in the repo, it could overwrite it. (not sure if this could lead to problems or not)

So I propose randomizing the names of the generated binaries. Anyone OK with that?

I will add an e2e tests as well.

ianlewis:
I think randomizing the names will be the simplest approach.

I think that the user will use the outputs to determine the right file name for each individual run, right?

laurentsimon:
Internally, we'll use random names. The final output will still be determined by the developer. For example, in the case of Go builders, the .slsa-goreleaser defines the output name.

You're right, the developer can still use the `outputs` field.

ianlewis:
SG. I think the generic workflow will need a way to either set the name of the provenance file or randomize the name since it currently uses a static name.

laurentsimon:
Done. Closing

laurentsimon:
Re-opening because @ianlewis is write about generic workflow attestation name.

laurentsimon:
Something to add is presubmits to test multiple runs of the builders.

laurentsimon:
I added support for randomized builder binaries for the generic workflow. Last AI is to add a pre-submit if we think it's necessary

ianlewis:
> I added support for randomized builder binaries for the generic workflow. Last AI is to add a pre-submit if we think it's necessary

What would you test in the pre-submit? Maybe failure if two calls to the workflow have the same attestation-name?

ianlewis:
Hmm. `actions/upload-artifact` doesn't have an option to fail if the artifact already exists so two calls to the generic workflow with the same attestation-name will clobber each other.
https://github.com/slsa-framework/slsa-github-generator/actions/runs/2716814219

laurentsimon:
Good catch. We rely on the user to give to provide non-colliding names for the result. The attestations overwrite each other, but the verification will fail for one of the artifacts, so we fail close ,not open. 

[...]

PR #503, opened 2022-07-07
feat: Add a RNG Action

laurentsimon:
This is part of https://github.com/slsa-framework/slsa-github-generator/issues/500 in order to generate random names.
In a follow-up PR, I will use names as `some-name-<random>`

### DIFF
--- a/.github/actions/rng/action.yml
+++ b/.github/actions/rng/action.yml
@@ -1,4 +1,4 @@
-name: "RRNG"
+name: "RNG"
 description: "Generate random bytes using /dev/urandom. WARNING: only use for non-cryptographic purposes (the results will show in logs)."
 inputs:
   length:
--- a/.github/actions/rng/action.yml
+++ b/.github/actions/rng/action.yml
@@ -9,7 +9,7 @@ outputs:
   random:
     description: >
       The output of the RNG encoded in hexadecimal.
-      Note: Due to the encoding, the length of the string will be twice as long as the input length request by the user.
+      Note: Due to the encoding, the length of the string will be twice as long as the input length requested by the user.
     value: "${{ steps.prng.outputs.result }}"
 
 runs:
--- a/.github/actions/rng/action.yml
+++ b/.github/actions/rng/action.yml
@@ -2,11 +2,14 @@ name: "RNG"
 description: "Generate random bytes using /dev/urandom. WARNING: only use for non-cryptographic purposes (the results will show in logs)."
 inputs:
   length:
-    description: "Length to generate (in bytes)."
-    required: true
+    description: "Number of raw random bytes to generate."
+    default: 16
+    required: false
 outputs:
   random:
-    description: "The output of the RNG (hex-encoded)."
+    description: >
+      The output of the RNG encoded in hexadecimal.
+      Note: Due to the encoding, the length of the string will be twice as long as the input length request by the user.
     value: "${{ steps.prng.outputs.result }}"
 
 runs:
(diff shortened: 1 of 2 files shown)


naveensrinivasan (approved):
This is cool! 

The project needs a README with all the actions.

laurentsimon:
Any feedback on the name of the Action. PRNG makes it sound like a cryptographic PNRG, but it's not intended to be used as such. I think I'll rename to `RNG` instead :)

naveensrinivasan:
After thinking about this should probably be written in `go`. This is adding tech debt.

ianlewis:
> After thinking about this should probably be written in `go`. This is adding tech debt.

Yeah, I'd like to reduce our use of inline shell scripts if we can too. Though it's pretty small so I could go either way.

I think installing Go compiler and compiling the program will add too much overhead to these kind of small actions. And we also  really want to avoid having to do a checkout of the repository as part of the action as there are version skew and security implications.

If we want to write them in another language I think the best choice would probably be [writing it in Javascript](https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action). It works similarly to the Dockerfile actions, but doesn't require building the image every time before running it.

.github/actions/rng/action.yml
```diff
@@ -0,0 +1,25 @@
...
+      id: prng
+      env:
+        LENGTH: "${{ inputs.length }}"
+      run: |
+        set -euo pipefail
+
+        value=$(head -c"$LENGTH" /dev/urandom | xxd -p)
+        echo "::set-output name=result::$value"
+
+
```
ianlewis, line 25: nit: remove extra lines at the end

.github/actions/rng/action.yml
```diff
@@ -0,0 +1,25 @@
+name: "RNG"
+description: "Generate random bytes using /dev/urandom. WARNING: only use for non-cryptographic purposes (the results will show in logs)."
+inputs:
+  length:
+    description: "Length to generate (in bytes)."
+    required: true
```
ianlewis, line 6: nit: Is there a reason it shouldn't have a default?

.github/actions/rng/action.yml
```diff
@@ -0,0 +1,25 @@
...
+  using: "composite"
+  steps:
+    - name: Generate random
+      id: prng
+      env:
+        LENGTH: "${{ inputs.length }}"
+      run: |
+        set -euo pipefail
+
+        value=$(head -c"$LENGTH" /dev/urandom | xxd -p)
```
ianlewis, line 22: nit: I think you can just do
```
xxd -l "$LENGTH" -p /dev/urandom
```

Also, both the above command and one you have here produces something longer than `$LENGTH` characters. Maybe something like this to ensure it's LENGTH characters long?
```
xxd -p -l $LENGTH /dev/urandom | head -c$LENGTH
```

ianlewis (approved):

.github/actions/rng/action.yml
laurentsimon: I'm not sure we can decide a default length. I could have it be 64 bytes default... which value would you set as default?

.github/actions/rng/action.yml
laurentsimon: ouch, `length` was intended to be the length of the raw random value (since this is directly related to the entropy), not the encoded one. I think I should rename the input argument then, or have a better description for it.
What would you suggest?

laurentsimon: Thanks. I'll use your first command. `length` is supposed to be the number of bytes of raw data not the length of the encoded data. I'll update the description to clarify

laurentsimon: I also found that we need to pass `-c $LENGTH` to avoid CR during display.

.github/actions/rng/action.yml
laurentsimon: thanks

.github/actions/rng/action.yml
laurentsimon: I've made the default 16, which is 128bit is good for most purposes

Issue #398, opened 2022-06-26
[feature] Use go code within the builder_go_slsa3.yml


naveensrinivasan:
**Is your feature request related to a problem? Please describe.**
This https://github.com/slsa-framework/slsa-github-generator/blob/024238bc10cc620be603a7f5329be7a672171ba7/.github/workflows/builder_go_slsa3.yml#L15 has lots of shell script which is harder to maintain and harder to debug

**Describe the solution you'd like**

Use go code instead of using shell script. Another option is to use something like https://github.com/magefile/mage for doing this. Also `mage` has `0` dependencies.

ianlewis:
I'd prefer not to use bash where it makes sense as well. I think the main issue with this is that we have to compile and run the code each run, it pretty hard for small snippets and increases the build times since Go compilation time isn't trivial. We've needed to reduce run time for #74 before.

I'm not totally sure about the builder workflows themselves but it might make sense for us to rewrite some of our [e2e tests](https://github.com/slsa-framework/example-package/tree/main/.github/workflows) in Go since much more extensive bash.

laurentsimon:
+1 to what @ianlewis said.

We have https://github.com/slsa-framework/slsa-github-generator/issues/87, https://github.com/slsa-framework/slsa-github-generator/issues/26 that are related.

ianlewis:
Yeah, it's a balance between
1. complexity - number of programming languages, libraries, coding styles etc.
2. reusability - ability to reuse code across actions/workflows
3. performance - need to compile, speed of action/workflow execution
4. security - attack surface, # and severity of potential bugs

As @laurentsimon said, I think the solution will not necessarily be just porting the existing scripts to Go but mostly creating reusable actions that are written in another language. That way it strikes a better balance with the concerns above. If we have that we can maybe just small snippets of bash and minimize the amount of code that we have to maintain.

ianlewis:
Adding comments from PR #503
https://github.com/slsa-framework/slsa-github-generator/pull/503#issuecomment-1180617335
> > > @naveensrinivasan 
> > > After thinking about this should probably be written in `go`. This is adding tech debt.
> > 
> > @ianlewis 
> > Yeah, I'd like to reduce our use of inline shell scripts if we can too. Though it's pretty small so I could go either way.
> > I think installing Go compiler and compiling the program will add too much overhead to these kind of small actions. And we also really want to avoid having to do a checkout of the repository as part of the action as there are version skew and security implications.
[...]

PR #514, opened 2022-07-09
Reduce the binary size  - slsa-builder

naveensrinivasan:
This will reduce the binary size of the builder.

It is helpful to have reduced binary size.


ianlewis (approved):
I think think this is ok, but just to be clear, I believe this only reduces the binary size for binaries built by the workflow during pre-submits and not the actually builder binary itself.

naveensrinivasan:
> I think think this is ok, but just to be clear, I believe this only reduces the binary size for binaries built by the workflow during pre-submits and not the actually builder binary itself.

Is that so? Which one reduces the binary size of the actual binary? 

Thanks

ianlewis:
> > I think think this is ok, but just to be clear, I believe this only reduces the binary size for binaries built by the workflow during pre-submits and not the actually builder binary itself.
> 
> Is that so? Which one reduces the binary size of the actual binary?
> 
> Thanks

Ah, you're right. You did get `config-release.yml`. I think that's the one we really care about. The other one `.github/workflows/configs-go/config-ldflags-main-dir.yml` is just used by the workflow for pre-submits so you can probably leave that off.

naveensrinivasan:
> > > I think think this is ok, but just to be clear, I believe this only reduces the binary size for binaries built by the workflow during pre-submits and not the actually builder binary itself.
> > 
> > 
> > Is that so? Which one reduces the binary size of the actual binary?
> > Thanks
> 
> Ah, you're right. You did get `config-release.yml`. I think that's the one we really care about. The other one `.github/workflows/configs-go/config-ldflags-main-dir.yml` is just used by the workflow for pre-submits so you can probably leave that off.

👍

naveensrinivasan:
I need some help with debugging this failure https://github.com/slsa-framework/slsa-github-generator/runs/7264731922?check_suite_focus=true

I started going down the rabbit hole and `pre-submit.e2e.go.default` and I was lost. There was lots of `shell` script.

I would appreciate as to why things are failing.

On a different note this is really hard to debug which has been already discussed here https://github.com/slsa-framework/slsa-github-generator/issues/398#issuecomment-1166687054

ianlewis:
That's a test error saying the ldflags in the pre-submit are not what we expect they are. It's failing because they were updated in the config file but net the test itself.

I think it should go back to passing if you revert the changes to `.github/workflows/configs-go/config-ldflags-main-dir.yml`. We only really care about `config-release.yml` anyway.

.github/workflows/configs-go/config-release.yml
```diff
@@ -12,3 +12,7 @@ goos: linux
 goarch: amd64
 dir: internal/builders/go/
 binary: slsa-builder-go-{{ .Os }}-{{ .Arch }}
+# ldflags: -s -w is used for reducing the size of the binary.
+ldflags:
+  - "-s"
+  - "-w"
```
laurentsimon, line 18: We're tryying to add a OEL for each file. Can you add it?

.github/workflows/configs-go/config-release.yml
```diff
@@ -12,3 +12,7 @@ goos: linux
 goarch: amd64
 dir: internal/builders/go/
 binary: slsa-builder-go-{{ .Os }}-{{ .Arch }}
+# ldflags: -s -w is used for reducing the size of the binary.
```
laurentsimon, line 15: Can you update the comment to be more specific, e.g. `-s -w ldflags remove the debugging information from the binary, which reduce the size of the binary.`?

.github/workflows/configs-go/config-release.yml
naveensrinivasan: I can. Isn't this a nit?

laurentsimon: yep. I also get caught with these myself... I think there's a plugin we can install in VSCode to do that automatically... Prettier?

.github/workflows/configs-go/config-release.yml
naveensrinivasan: I did that.  Thanks!

.github/workflows/configs-go/config-release.yml
naveensrinivasan: I am fine with nit. I took care of it. We want to be cognizant with new contributions :)

laurentsimon: Agreed. We don't have a `CONTIBUTION.md`, I think you're in the best position to help us create one and list all the quirks / recommendations /etc you've encountered so far.

PR #519, opened 2022-07-11
Port the sha256 shell to go code

naveensrinivasan:
The goal for this port from shell script to `go` was to reduce the
amount of shell script.

It is easier to refactor managed code compared to shell script.

This has been discussed https://github.com/slsa-framework/slsa-github-generator/issues/398

### DIFF
--- a/.github/actions/compute-sha256/main.go
+++ b/.github/actions/compute-sha256/main.go
@@ -1,0 +1,40 @@
+package main
+
+import (
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"os/exec"
+)
+
+func main() {
+	if len(os.Args) < 2 {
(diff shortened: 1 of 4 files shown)


.github/actions/compute-sha256/main.go
```diff
@@ -0,0 +1,40 @@
...
+
+	data, err := ioutil.ReadFile(file)
+	if err != nil {
+		panic(fmt.Sprintf("failed to read file: %s", file))
+	}
+
+	hash := sha256.Sum256(data)
+	log.Printf("computed sha: %s\n", hex.EncodeToString(hash[:]))
+
+	cmd := exec.Command("/usr/bin/env", "bash", "-c",
```
laurentsimon, line 32: Note: this can be done via `printf`. Given the discussion in https://github.com/slsa-framework/slsa-github-generator/pull/503#issuecomment-1180617335, we may want to use TS instead of Go.

PR #533, opened 2022-07-12
feat: bump Action hashes

laurentsimon:
Update Action hashes to use the rng Action to randomize binary names


asraa (approved):
nice, same hash for the same repo actions :P

Issue #544, opened 2022-07-14
[release] Generic generator


laurentsimon:
We plan to release the generic generator for non-container artifact next week. This will be a "silent" release. Once it's is released, we will reach out to individual repositories via issues / PRs for adoption, in order to gather early feedback.

Task to complete for the release:
- [X] E2e 2 hash verification verify; done except the asset which we plan to disable
- [x] Test data on verifier https://github.com/slsa-framework/slsa-verifier/issues/133 @asraa will handle this
- [x] E2e tests for multiple calls to the generic builder https://github.com/slsa-framework/example-package/actions/runs/2679093902
- [x] E2e tests for multiple calls to the Go builder https://github.com/slsa-framework/example-package/actions/runs/2679105904
- [x] Option to set the name of the attestation https://github.com/slsa-framework/slsa-github-generator/pull/545
- [x] E2e test for wrong subject for generic workflow, see [e2e.generic.schedule.main.adversarial-invalidsubjects.slsa3.yml](https://github.com/slsa-framework/example-package/blob/main/.github/workflows/e2e.generic.schedule.main.adversarial-invalidsubjects.slsa3.yml)
- [x] E2e tests for adversarial attestation path see [e2e.generic.schedule.main.adversarial-invalidpath.slsa3.yml](https://github.com/slsa-framework/example-package/blob/main/.github/workflows/e2e.generic.schedule.main.adversarial-invalidpath.slsa3.yml). 
- [x] Fix pre-submits https://github.com/slsa-framework/slsa-github-generator/pull/549
- [x] Example for Java / Maven in https://github.com/slsa-framework/slsa-github-generator/pull/556 @loosebazooka will take care of it, thank you!

I'll take a stab at these. If you want to help, let me know.

Later, we will have a proper announcement with blog post, maybe combining the work on generic containers.

/cc @ianlewis @asraa @joshuagl

laurentsimon:
I finished the items on the list.

ianlewis:
All of the tasks listed above are complete so I think we can close the issue.

PR #545, opened 2022-07-14
feat: Input argument for the name of attestation of generic workflow

laurentsimon:
See https://github.com/slsa-framework/slsa-github-generator/issues/544

This PR adds an input to the generic re-usable workflow for the attestation filename.

### DIFF
--- a/.github/workflows/generator_generic_slsa3.yml
+++ b/.github/workflows/generator_generic_slsa3.yml
@@ -37,7 +37,7 @@ on:
       attestation-name:
         description: >
           The artifact name of the signed provenance. 
-          The file must have the intoto.json extension.
+          The file must have the intoto.jsonl extension.
 
           Default: attestation.intoto.jsonl
         required: false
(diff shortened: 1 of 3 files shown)


.github/workflows/generator_generic_slsa3.yml
```diff
@@ -34,6 +34,11 @@ on:
         required: false
         type: boolean
         default: false
+      attestation-name:
+        description: "The artifact name of the signed provenance. (Default: attestation.intoto.jsonl)"
```
asraa, line 38: should we fix the extension to be intoto.jsonl?

asraa (approved):
one small nit -- but otherwise lgtm

.github/workflows/generator_generic_slsa3.yml
laurentsimon: Good point. I'll add code to do that. Thanks

PR #570, opened 2022-07-18
doc: Update the list / status of builders / generators

laurentsimon:


laurentsimon:
> Seems fine. Might be worth getting Kara to over the language a bit before a release is done.

Good point.

PR #591, opened 2022-07-20
Add digest input to container docs

### LINKED ISSUE #543: [bug] Missing input digest in readme example for container workflow
**Describe the bug**
While using the recently released generator_container_slsa3.yml and following the exemplary workflow provided in the README, the error message "Input digest is required, but not provided while calling." is thrown.
Looking at the input defined in the example only the following are defined as input parameters:

```
with:
    image: ${{ needs.build.outputs.tag }}
    registry-username: ${{ github.actor }}
    # TODO(https://github.com/slsa-framework/slsa-github-generator/issues/492): Remove after GA release.
    compile-generator: true
```

**To Reproduce**
Using the example workflow described [here](https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/container).

**Expected behavior**
No error is thrown that the input digest is required.

Thank you (:
(probably @ianlewis)

ianlewis:
Fixes #543


internal/builders/container/README.md
```diff
@@ -134,6 +136,7 @@ jobs:
     uses: slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@main
     with:
       image: ${{ needs.build.outputs.image }}
+      digest: ${{ needs.build.outputs.digest }}
```
laurentsimon, line 139: have we decided yet if image and digest should be separate inputs or not?
Fo example, in https://github.com/actions/starter-workflows/blob/main/ci/docker-publish.yml#L93, it's passed as one parameter.

laurentsimon (approved):

internal/builders/container/README.md
ianlewis: I kind of wanted to have them be separate to make it clear that only the digest is supported. I wanted to force folks to use the digest to avoid racing with something that could modify the image tag.

I could merge them into one and do some validation on the image string to make sure it includes the digest and return an error if not, but I thought that this would better UX. wdut?

laurentsimon: I'm not a UX expert tbh. I think it'd be fine to have a single option. So long as it's a required option and we validate it, I suppose it's fine.

Having them separate works too. We can re-visit later without blocking this PR for now.

@asraa @joshuagl do you have any thoughts on the above?

ianlewis: SG. Agreed that it might not be a *nicer* UX, but a more explicit one.

PR #596, opened 2022-07-20
fix: Remove `build:id` in generic examples

laurentsimon:


internal/builders/generic/README.md
```diff
@@ -370,19 +370,12 @@ jobs:
...
-        #
-        # =================================================
-        id: build
         run: |
           # Your normal build workflow targets here
           bazel build //path/to/target_binary //path/to_another/binary
           # ======================================================
           #
-          # Step 3: Copy the binaries from `bazel-bin` path (i.e.,
+          # Step 2: Copy the binaries from `bazel-bin` path (i.e.,
```
ianlewis, line 378: nit: do we really need to number these?

ianlewis (approved):

internal/builders/generic/README.md
laurentsimon: Wanted to order the steps. Shall we remove them?

ianlewis: I'm ok with keeping them if you think it's helpful.

olivekl (commented):
In this case I'd label them, yes, because it's otherwise hard to parse visually. BUT: I'd suggest taking the steps out of the comments entirely and having them in regular text, followed by the code example for that step. That way it's very clear (visually) to the user what the steps are when glancing over the docs, and they're only copying code comments into their system that are descriptive info reminding them of what the code does.

laurentsimon:
> In this case I'd label them, yes, because it's otherwise hard to parse visually. BUT: I'd suggest taking the steps out of the comments entirely and having them in regular text, followed by the code example for that step. That way it's very clear (visually) to the user what the steps are when glancing over the docs, and they're only copying code comments into their system that are descriptive info reminding them of what the code does.

So you mean something like the following:
- Step 1: add bla to your code
```yaml
uses: builder@v1.2
```

- Step 2: add bla 2 to your code
```yaml
jobs:
  id: bla
```

internal/builders/generic/README.md
```diff
@@ -394,7 +387,7 @@ jobs:
 
       # ========================================================
       #
-      # Step 4: Add a step to generate the provenance subjects
+      # Step 2: Add a step to generate the provenance subjects
```
olivekl, line 390: Should this be Step 3?

olivekl:
> So you mean something like the following, instead of inlining the comments in the code:
> 
> * Step 1: add bla to your code
> 
> ```yaml
> uses: builder@v1.2

Exactly.

internal/builders/generic/README.md
laurentsimon: it's step 4, but the previous comment is part of the script above so shows up with a different color. I'll separate steps an code s you suggested in another comment. Thanks!

laurentsimon:
> > So you mean something like the following, instead of inlining the comments in the code:
> > 
> > * Step 1: add bla to your code
> > 
> > ```yaml
> > uses: builder@v1.2
> > ```
> 
> Exactly.

I've made the changes. Thank you both for the feedback.

ianlewis:
Thanks @olivekl !

PR #598, opened 2022-07-21
feat: Share util functions

laurentsimon:
This PR does 3 things:
1. Share util functions between builders
2. Ensure the files we create do not already exist (`O_EXCL`)
3. Add sanity checks when downloading an artifact for the non-existence of the file

I started looking into 

https://github.com/slsa-framework/slsa-github-generator/blob/0f97cca414ee57448ec44667e92aa65c46a23eb6/internal/builders/go/main.go#L87 

and https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/attest.go#L165 

Then I realized that a malicious developer can use this function to sign any file present in the generator directory. The reason is that if the file already exists, this function does not return an error and continue silently. Fortunately we don't checkout the developer repository (by design), because it might have allowed the developer to set the path to a file they control.

I think an attacker could only have selected files with `.intoto.jsonl` extension, so just a few testcase files. I have not tested filenames like `some-file\0`(to remove the extension in https://github.com/slsa-framework/slsa-github-generator/blob/0f97cca414ee57448ec44667e92aa65c46a23eb6/internal/builders/go/main.go#L86), but I doubt it would work...

So I created the `CreateNewFileUnderCurrentDirectory()` function which uses `O_EXCL` and use it across the 2 existing builders.

Besides these changes, the rest is mostly re-factoring.


ianlewis:
Thanks, I'll give this a review tomorrow

internal/builders/go/main_test.go
```diff
@@ -270,6 +267,7 @@ func Test_runVerify(t *testing.T) {
 			s := r.end()
 
 			if !errCmp(err, tt.err) {
+				fmt.Println(err, tt.err)
```
asraa, line 270: nit: leftover debug?

internal/utils/marshall.go
```diff
@@ -62,3 +38,24 @@ func UnmarshallList(arg string) ([]string, error) {
 	}
 	return res, nil
 }
+
+func MarshallToString(args interface{}) (string, error) {
```
asraa, line 42: super small nit: Marshall to Marshal everywhere?

internal/utils/marshall.go
```diff
@@ -62,3 +38,24 @@ func UnmarshallList(arg string) ([]string, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("base64.StdEncoding.EncodeString: %w", err)
+	}
+	return encoded, nil
+}
+
+func MarshallToBytes(args interface{}) ([]byte, error) {
+	encoded, err := MarshallToString(args)
+	if err != nil {
+		return []byte{}, nil
```
asraa, line 58: Did you want to return `nil, err` here?

asraa (commented):
Thanks! Really minor nits

internal/builders/go/main_test.go
laurentsimon: good catch. Thanks!

internal/utils/marshall.go
laurentsimon: probably. I'll make the change. Thank you!

internal/utils/marshall.go
laurentsimon: yes!

asraa (approved):

laurentsimon:
I'll wait for an ack from @ianlewis before merging.

internal/builders/go/pkg/config.go
```diff
@@ -125,7 +135,24 @@ func validateMain(cf *goReleaserConfigFile) error {
...
+	if err := utils.PathIsUnderCurrentDirectory(*cf.Main); err != nil {
+		return convertPathError(err, "PathIsUnderCurrentDirectory")
+	}
+	return nil
+}
+
+func convertPathError(e error, msg string) error {
+	// TODO(https://github.com/slsa-framework/slsa-github-generator/issues/599): use same error contructions.
+	var errInternal *utils.ErrInternal
+	var errPath *utils.ErrInvalidPath
```
ianlewis, line 147: nit: you can move these inside the if block below

internal/utils/path.go
```diff
@@ -0,0 +1,75 @@
...
+type ErrInternal struct {
+	errors.WrappableError
+}
+
+// ErrInvalidPath indicates an invalid path.
+type ErrInvalidPath struct {
+	errors.WrappableError
+}
+
+func PathIsUnderCurrentDirectory(path string) error {
```
ianlewis, line 36: nit: add a short godoc to these functions.

ianlewis (approved):

internal/builders/go/pkg/config.go
laurentsimon: good catch. Done.

internal/utils/path.go
laurentsimon: done. Let's have pre-submit to catch these in the future if we can

ianlewis: Yeah, this should be captured under #450

PR #606, opened 2022-07-21
Fix missing digest input and update attestation

Root-DE:
This Pull Request:
- adds the digest input
- Fix: attestation generation (adds the missing subjects)
- remove already set env variables
co-authored-by @jobroe10

We made these changes while using the container workflow and realising it was not fully working

### DIFF
--- a/.github/actions/compute-sha256/action.yml
+++ b/.github/actions/compute-sha256/action.yml
@@ -15,17 +15,15 @@ runs:
     - name: Compute the sha256
       id: compute
       shell: bash
-      env:
-        UNTRUSTED_PATH: "${{ inputs.path }}"
       run: |
         set -euo pipefail
 
-        echo "Computing SHA256 for $UNTRUSTED_PATH"
-        if ! [[ -f "$UNTRUSTED_PATH" ]]; then
-            echo "File $UNTRUSTED_PATH not present"
+        echo "Computing SHA256 for $BUILDER_RELEASE_BINARY"
+        if ! [[ -f "$BUILDER_RELEASE_BINARY" ]]; then
+            echo "File $BUILDER_RELEASE_BINARY not present"
             exit 5
         fi
-        digest=$(sha256sum "$UNTRUSTED_PATH" | awk '{print $1}')
+        digest=$(sha256sum "$BUILDER_RELEASE_BINARY" | awk '{print $1}')
         echo "computed sha: $digest"
         
         echo "::set-output name=sha256::$digest"
--- a/.github/actions/generate-builder/action.yml
+++ b/.github/actions/generate-builder/action.yml
@@ -43,17 +43,6 @@ runs:
     - name: Generate builder
       shell: bash
       env:
-        # Builder.
-        BUILDER_REPOSITORY: slsa-framework/slsa-github-generator  # The repository to download the pre-built builder binary from.
-        BUILDER_RELEASE_BINARY: "${{ inputs.binary }}"            # The name of the pre-built binary in the release assets. This is also used as the final binary name when compiling the builder from source.
-        # Verifier
-        # NOTE: These VERIFIER_* variables are used in the builder-fetch.sh script for verification of builder
-        # release binaries when the compile-builder input is false.
-        VERIFIER_REPOSITORY: slsa-framework/slsa-verifier         # The repository to download the pre-built verifier binary from.
-        VERIFIER_RELEASE_BINARY: slsa-verifier-linux-amd64        # The name of the verifier binary in the release assets.
-        VERIFIER_RELEASE_BINARY_SHA256: f92fc4e571949c796d7709bb3f0814a733124b0155e484fad095b5ca68b4cb21  # The expected hash of the verifier binary.
-        VERIFIER_RELEASE: v1.1.1                                   # The version of the verifier to download.
-
         COMPILE_BUILDER: "${{ inputs.compile-builder }}"
         BUILDER_REF: "${{ inputs.ref }}"
         BUILDER_DIR: "${{ inputs.directory }}"
@@ -62,7 +51,5 @@ runs:
       run: ./.github/actions/generate-builder/generate-builder.sh
 
     - name: Compute sha256 of builder
-      uses: slsa-framework/slsa-github-generator/.github/actions/compute-sha256@1d646d70aeba1516af69fb0ef48206580122449b
+      uses: slsa-framework/slsa-github-generator/.github/actions/compute-sha256@9c56e223d2486617188b6a11ed385ac1a94d77c0
       id: compute
-      with:
-        path: "${{ inputs.binary }}"
--- a/.github/workflows/generator_container_slsa3.yml
+++ b/.github/workflows/generator_container_slsa3.yml
@@ -136,9 +136,9 @@ jobs:
 
           # Generate a predicate only.
           predicate_name="predicate.json"
-          ./"$BUILDER_BINARY" attest --signature="" --predicate="$predicate_name"
-
+          echo $(echo $(echo $UNTRUSTED_DIGEST | cut -d: -f2) $UNTRUSTED_IMAGE) | base64
+          ./"$BUILDER_RELEASE_BINARY" attest --subjects $(echo $(echo $(echo $UNTRUSTED_DIGEST | cut -d: -f2) $UNTRUSTED_IMAGE) | base64 -w0) --predicate="$predicate_name"
           COSIGN_EXPERIMENTAL=1 cosign attest --predicate="$predicate_name" \
             --type slsaprovenance \
             --force \
-            "${UNTRUSTED_IMAGE}@${UNTRUSTED_DIGEST}"
+            "${UNTRUSTED_IMAGE}"
--- a/internal/builders/container/README.md
+++ b/internal/builders/container/README.md
@@ -66,7 +66,7 @@ provenance:
     registry-username: ${{ github.actor }}
     digest: ${{ needs.build.outputs.digest }}
     # TODO(https://github.com/slsa-framework/slsa-github-generator/issues/492): Remove after GA release.
-    compile-generator: true
+    compile-generator: true # currently has to be set true to work
   secrets:
     registry-password: ${{ secrets.GITHUB_TOKEN }}
 ```
@@ -139,7 +139,7 @@ jobs:
       registry-username: ${{ github.actor }}
       digest: ${{ needs.build.outputs.digest }}
       # TODO(https://github.com/slsa-framework/slsa-github-generator/issues/492): Remove after GA release.
-      compile-generator: true
+      compile-generator: true # currently has to be set true to work
     secrets:
       registry-password: ${{ secrets.GITHUB_TOKEN }}
 ```
--- a/internal/builders/container/README.md
+++ b/internal/builders/container/README.md
@@ -64,6 +64,7 @@ provenance:
   with:
     image: ${{ needs.build.outputs.tag }}
     registry-username: ${{ github.actor }}
+    digest: ${{ needs.build.outputs.digest }}
     # TODO(https://github.com/slsa-framework/slsa-github-generator/issues/492): Remove after GA release.
     compile-generator: true
   secrets:
@@ -85,6 +86,7 @@ jobs:
       packages: write
     outputs:
       image: ${{ steps.image.outputs.image }}
+      digest: ${{ steps.build.outputs.digest }}
     runs-on: ubuntu-latest
     steps:
       - name: Checkout the repository
@@ -135,6 +137,7 @@ jobs:
     with:
       image: ${{ needs.build.outputs.image }}
       registry-username: ${{ github.actor }}
+      digest: ${{ needs.build.outputs.digest }}
       # TODO(https://github.com/slsa-framework/slsa-github-generator/issues/492): Remove after GA release.
       compile-generator: true
     secrets:
--- a/.github/actions/secure-download-artifact/action.yml
+++ b/.github/actions/secure-download-artifact/action.yml
@@ -24,7 +24,7 @@ runs:
 
     - name: Compute the hash
       id: compute
-      uses: slsa-framework/slsa-github-generator/.github/actions/compute-sha256@1d646d70aeba1516af69fb0ef48206580122449b
+      uses: slsa-framework/slsa-github-generator/.github/actions/compute-sha256@9c56e223d2486617188b6a11ed385ac1a94d77c0
       with:
         path: "${{ inputs.path }}"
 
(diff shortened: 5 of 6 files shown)


Root-DE:
@ianlewis what do you think?

.github/actions/generate-builder/action.yml
```diff
@@ -43,17 +43,6 @@ runs:
...
-        BUILDER_REPOSITORY: slsa-framework/slsa-github-generator  # The repository to download the pre-built builder binary from.
-        BUILDER_RELEASE_BINARY: "${{ inputs.binary }}"            # The name of the pre-built binary in the release assets. This is also used as the final binary name when compiling the builder from source.
-        # Verifier
-        # NOTE: These VERIFIER_* variables are used in the builder-fetch.sh script for verification of builder
-        # release binaries when the compile-builder input is false.
-        VERIFIER_REPOSITORY: slsa-framework/slsa-verifier         # The repository to download the pre-built verifier binary from.
-        VERIFIER_RELEASE_BINARY: slsa-verifier-linux-amd64        # The name of the verifier binary in the release assets.
-        VERIFIER_RELEASE_BINARY_SHA256: f92fc4e571949c796d7709bb3f0814a733124b0155e484fad095b5ca68b4cb21  # The expected hash of the verifier binary.
-        VERIFIER_RELEASE: v1.1.1                                   # The version of the verifier to download.
-
```
ianlewis, line 56: These environment variables are necessary for supporting when `compile-builder` is false so they should not be deleted.

.github/actions/generate-builder/action.yml
```diff
@@ -62,7 +51,5 @@ runs:
       run: ./.github/actions/generate-builder/generate-builder.sh
 
     - name: Compute sha256 of builder
-      uses: slsa-framework/slsa-github-generator/.github/actions/compute-sha256@1d646d70aeba1516af69fb0ef48206580122449b
+      uses: slsa-framework/slsa-github-generator/.github/actions/compute-sha256@9c56e223d2486617188b6a11ed385ac1a94d77c0
```
ianlewis, line 54: I'm afraid this won't work because this commit digest doesn't exist in the repo.

```shell
$ git log 9c56e223d2486617188b6a11ed385ac1a94d77c0
fatal: bad object 9c56e223d2486617188b6a11ed385ac1a94d77c0
```

.github/actions/compute-sha256/action.yml
```diff
@@ -15,17 +15,15 @@ runs:
...
-        echo "Computing SHA256 for $UNTRUSTED_PATH"
-        if ! [[ -f "$UNTRUSTED_PATH" ]]; then
-            echo "File $UNTRUSTED_PATH not present"
+        echo "Computing SHA256 for $BUILDER_RELEASE_BINARY"
+        if ! [[ -f "$BUILDER_RELEASE_BINARY" ]]; then
+            echo "File $BUILDER_RELEASE_BINARY not present"
             exit 5
         fi
-        digest=$(sha256sum "$UNTRUSTED_PATH" | awk '{print $1}')
+        digest=$(sha256sum "$BUILDER_RELEASE_BINARY" | awk '{print $1}')
```
ianlewis, line 26: Where is `BUILDER_RELEASE_BINARY` set? I don't believe this will work. Also, we want to use the `path` input for the action as this is intended to be a generic reusable action.

.github/workflows/generator_container_slsa3.yml
```diff
@@ -136,9 +136,9 @@ jobs:
 
           # Generate a predicate only.
           predicate_name="predicate.json"
-          ./"$BUILDER_BINARY" attest --signature="" --predicate="$predicate_name"
-
+          echo $(echo $(echo $UNTRUSTED_DIGEST | cut -d: -f2) $UNTRUSTED_IMAGE) | base64
+          ./"$BUILDER_RELEASE_BINARY" attest --subjects $(echo $(echo $(echo $UNTRUSTED_DIGEST | cut -d: -f2) $UNTRUSTED_IMAGE) | base64 -w0) --predicate="$predicate_name"
```
ianlewis, line 140: A couple of things.

1. I intentionally set the `--signature=""` option so that this command won't sign the provenance and will just output the SLSA predicate.
2. We don't need to set the subjects here because the `cosign attest` command below will set the subjects for us.

.github/workflows/generator_container_slsa3.yml
```diff
@@ -136,9 +136,9 @@ jobs:
...
           COSIGN_EXPERIMENTAL=1 cosign attest --predicate="$predicate_name" \
             --type slsaprovenance \
             --force \
-            "${UNTRUSTED_IMAGE}@${UNTRUSTED_DIGEST}"
+            "${UNTRUSTED_IMAGE}"
```
ianlewis, line 144: We absolutely need to add the digest here so that we sign the right image. Without it `cosign` will use the `latest` tag which could have been modified in the container repository after the image was build by this GitHub actions run.

internal/builders/container/README.md
```diff
@@ -85,6 +86,7 @@ jobs:
       packages: write
     outputs:
       image: ${{ steps.image.outputs.image }}
+      digest: ${{ steps.build.outputs.digest }}
```
ianlewis, line 89: I added the digest to the docs in #591 so I think this will be unnecessary if you merge with main.

ianlewis (commented):
Thanks for checking out the container workflow!

I'm not sure I understand the issues we are solving with some of these changes. To help us understand better, could you create an issue and describe some of the problems you ran into?

.github/actions/generate-builder/action.yml
Root-DE: This looks like an GH Issue because this commit exists in our fork (https://github.com/Root-DE/slsa-github-generator/commit/9c56e223d2486617188b6a11ed385ac1a94d77c0) but you can't find it even when using the general GH search. To enable contributions it seems like we would need a branch (This is only necessary due to the necessaty of hard coded commits hashs). Can you create a branch for us? We would then merge our stuff into this one for testing.

.github/actions/generate-builder/action.yml
ianlewis: The environment variables being set isn't really a problem. The script that runs as part of that action just changes behavior based on the `compile-builder` flag. However, the container workflow will not work yet with `compile-builder` set to false because we haven't created a release that includes the container workflow.

Currently you will need to use `main` or a recent commit digest in your workflow call and set `compile-builder` to true for it to work.

Once we cut a release that includes the container workflow then setting it to false should work but not until then.

.github/actions/generate-builder/action.yml
ianlewis: You're right. It does exist in your fork, but here you are referencing the main repository (`slsa-framework/slsa-github-generator`) so this action call will fail because that commit digest does not exist in the `slsa-framework/slsa-github-generator` repo. It only exists in the `Root-DE/slsa-github-generator` repo.

If you need to update this sha to incorporate some changes then you need get a PR merged in the main repository and then create a new PR reference that previous merge commit digest.

BTW, what was the reason you needed to update these digests?

.github/actions/compute-sha256/action.yml
ianlewis: I'm not sure I follow.

This is a completely separate GitHub action (`compute-sha256`) so I'm pretty sure that the env vars won't carry over from the `generate-builder` action. Those env vars are only available to the "Generate builder" step in the `generate-builder` action.

I would need more info on what you are doing to be able to tell why you got that error. Could you submit an issue with more details including a link to the workflow you are trying to run?

.github/workflows/generator_container_slsa3.yml
Root-DE: We changed to $BUILDER_RELEASE_BINARY according to [this comment](https://github.com/slsa-framework/slsa-github-generator/pull/606#discussion_r927428530) see [the test run from us using $BUILDER_BINARY](https://github.com/Root-DE/Example-Image/runs/7377872538?check_suite_focus=true#step:6:225):
image

------------------

When we run this with $BUILDER_RELEASE_BINARY and --signature="" we got the following error:
invalid suffix: "". Must be .intoto.jsonl
This errors originates from attest.go file. According to the file it has to end with .intoto.jsonl, it defaults to attestation.intoto.jsonl, thats why we removed the --signature="".

------------------


Running it without --signature="" we received the error:
expected at least one subject
image
you can verify this [here](https://github.com/Root-DE/Example-Image/runs/7378040918?check_suite_focus=true#step:6:225)
Thats why we set the subject in the pr.

internal/builders/container/README.md
Root-DE: thanks for the hint you're right, we can drop this one

.github/actions/compute-sha256/action.yml
ianlewis: I think I understand a bit better now. The `BUILDER_BINARY` is set by the [container workflow](https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/workflows/generator_container_slsa3.yml#L19) having a different name than the other workflows.

`BUILDER_BINARY` is not used by the `generate-builder` action. It's simply [set as an input to the action](https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/workflows/generator_container_slsa3.yml#L99) and should be fine. `BUILDER_RELEASE_BINARY` is set to that input [here](https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/actions/generate-builder/action.yml#L48). So I think the error you are getting is a result of deleting the env vars from the `generate-builder` action below.

.github/workflows/generator_container_slsa3.yml
ianlewis: Ok, I see. This may be a separate bug that we should fix separately. I believe this was caused by #545

Please submit issues when you encounter these errors.

ianlewis: Added #614 to track this error.

.github/actions/generate-builder/action.yml
ianlewis: Added #613 to track this change.

.github/workflows/generator_container_slsa3.yml
jobroe10: You're right, we just re-evaluated and now it also works with the digest

.github/actions/generate-builder/action.yml
jobroe10: I see, we updated these digest due to the changes made in compute-sha256

.github/actions/compute-sha256/action.yml
jobroe10: BUILDER_BINARY is set as the input but doesn't [this line in the generate-builder action](https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/actions/generate-builder/action.yml#L48) include it which is then used as an [input for compute-sha256](https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/actions/generate-builder/action.yml#L68). But on the other side generate-builder.sh is building the executable with the name $BUILDER_RELEASE_BINARY. 

Flow:
generator_container_slsa3.yml (BUILDER_BINARY: generic-generator, which is used as an input for generate builder binary: "${{ env.BUILDER_BINARY }}")
--> generate-builder/action.yml (path: "${{ inputs.binary }}" uses it as an input for compute-sha256)
--> compute-sha256/action.yml (gets path as input UNTRUSTED_PATH: "${{ inputs.path }}" and uses it to look for the executable if ! [[ -f "$UNTRUSTED_PATH" ]]; then...)

We assume that the error 
```
Computing SHA256 for generic-generator
File generic-generator not present
```
results from the executable being generated with the name slsa-generator-generic-linux-amd64 and compute-sha256 is [looking for a file with the name generic-generator](https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/actions/compute-sha256/action.yml#L24).

ianlewis: Normally the way you describe it is true.

1. [`BUILDER_BINARY` is set to `generic-generator`](https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/workflows/generator_container_slsa3.yml#L19)
2. [`BUILDER_BINARY` is used as input into `generate-builder`](https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/workflows/generator_container_slsa3.yml#L99)
3. [`generate-builder` sets the input value to `BUILDER_RELEASE_BINARY`](https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/actions/generate-builder/action.yml#L48)
4. [`generate-builder.sh` uses `BUILDER_RELEASE_BINARY` as the name of the binary to build. ](https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/actions/generate-builder/generate-builder.sh#L13)

However in your PR you [removed](https://github.com/slsa-framework/slsa-github-generator/pull/606/files#diff-bff0b44f6cbd4c751293462017ea9ce8c048288fad10886047a63503ff7112adL46) the environment variables set in `generate-builder`. This causes the value of `BUILDER_RELEASE_BINARY` to be the value `slsa-generator-generic-linux-amd64` inherited from the workflow as set [here](https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/workflows/generator_container_slsa3.yml#L20). So the `generate-builder` action builds the builder with the name `slsa-generator-generic-linux-amd64`.

Then the [next step](https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/actions/generate-builder/action.yml#L64-L68) that generates the sha256 hash of the binary accepts `inputs.binary` as the binary name which is `generic-generator`. This is where your error is occurring. The `compute-sha256` action is expecting to find a binary named `generic-generator` but the name of the binary that was created is  `slsa-generator-generic-linux-amd64`.

If you [restore the environment variables](https://github.com/slsa-framework/slsa-github-generator/pull/606/files#diff-bff0b44f6cbd4c751293462017ea9ce8c048288fad10886047a63503ff7112adL46) in the `generate-builder` action. This error should go away.

jobroe10: You are right, we just verified it with the other adaptions we made. Thank you for taking the time to look into this! 
If you're interested, [this run](https://github.com/Root-DE/Example-Image/actions/runs/2718167924) finished successfully, pushed image and signed attestation. We also used the cosign verifier to verify the signed attestation which works.

PR #595, opened 2022-07-20
Haskell provenance

mihaimaruseac:
Allows using either Stack or Cabal build tools

Tested via example repo ([Cabal](https://github.com/mihaimaruseac/slsa-lvl3-generic-provenance-in-haskell-example/blob/v0.0.1.1/.github/workflows/ci.yaml), [Stack](https://github.com/mihaimaruseac/slsa-lvl3-generic-provenance-in-haskell-example/blob/v0.0.1.0/.github/workflows/ci.yaml)).

Also did some cleanup (removed trailing whitespace, fixed off-by-one errors)

PR #608, opened 2022-07-21
Add provenance for Haskell

mihaimaruseac:
Signed-off-by: Mihai Maruseac <mihaimaruseac@google.com>

Does #595 after the formatting from #596, with the same layout


laurentsimon (approved):
Thank you!

PR #611, opened 2022-07-22
Add attestation-name pre-submit

### LINKED ISSUE #500: [feature] randomize binary names

ianlewis:
Adds a test for when `attestation-name` is the same for multiple calls to the generic workflow.

The pre-submit job "pre-submit e2e generic adversarial-attestation-name" needs to be added to required checks on the main branch after merge.

Fixes #500

### DIFF
--- a/.github/workflows/pre-submit.e2e.generic.adversarial-attestation-name.yml
+++ b/.github/workflows/pre-submit.e2e.generic.adversarial-attestation-name.yml
@@ -42,10 +42,12 @@ jobs:
     runs-on: ubuntu-latest
     needs: [build, build2]
     if: always()
-    env:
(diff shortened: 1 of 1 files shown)


ianlewis:
Close in favor of e2e tests

Issue #583, opened 2022-07-19
[feature] Add a roadmap issue


laurentsimon:
It would be useful to have an issue with a roadmap, to let people know what we're working on in the next 3-6 months.

ianlewis:
Depending on what we think the roadmap should have in it, one easy way to communicate what we're working on is to use [milestones](https://github.com/slsa-framework/slsa-github-generator/milestones).

Github Projects are also a way do to it but requires more effort to keep up to date.

laurentsimon:
Thanks for the feedback. Do you think just an issue that explains / lists milestone would be enough? I wonder if people look at milestones directly, or whether having an issue (or a discussion or a project) would help give more visibility for someone who visits the repo. I know people tends to look at issues when visiting a repo, for example.

ianlewis:
> Thanks for the feedback. Do you think just an issue that lists the milestone would be enough? I wonder if people look at milestones directly, or whether having an issue (or a discussion or a project) would help give more visibility for someone who visits the repo. I know people tends to look at issues when visiting a repo, for example.

Yeah, I'm not sure what folks would really want in practice. Milestones can answer the question "What are we working on in the next 3-6 months", but you're right that it doesn't have the opportunity for discussion. I think milestones are easy to keep updated and should be good enough for now if we link to them on the README to indicate that we are using them. Folks can always create issues if they want something.

We can put out more high level roadmap docs or something if there is need for them. I just feel like these docs often are *super* high level and don't really say much. And when they do they aren't really updated.

Some example roadmaps I found:
- [Kubernetes sig-release](https://github.com/kubernetes/sig-release/blob/master/roadmap.md) 
  - other sigs don't seem to have roadmaps? or I couldn't find any recent ones. They seem to only put out annual reports after the fact
- [sigstore roadmap](https://github.com/sigstore/community/blob/36e27e4743af3229aa2004761017d736a28a76e7/ROADMAP.md) - Last updated in November 2021
- etcd used to have one but they [deleted it](https://github.com/etcd-io/etcd/commit/a722827c96162d7d06e72cadfaed223793dc4d78).

Do you have any other examples of good roadmaps created by other projects?

laurentsimon:
There's also SLSA's roadmap here https://github.com/slsa-framework/slsa/issues/384
[...]

PR #612, opened 2022-07-22
Add links to milestones as a roadmap

ianlewis:
Updates #583


laurentsimon (approved):
Thanks. We may want to update the list of builders we have. Right now it says "WIP", maybe we can add a link to the corresponding milestone too.

PR #616, opened 2022-07-22
Remove legacy env vars

### LINKED ISSUE #613: [refactor] Clean up env vars
Env vars need to be cleaned up in the container workflow. Many env vars set here are legacy vars that are no longer used.
https://github.com/slsa-framework/slsa-github-generator/blob/1a55da84338004ca26d38b941483c89ab93a695d/.github/workflows/generator_container_slsa3.yml#L19-L30

ianlewis:
Fixes #613

This PR removes legacy env vars from the container workflow. These environment variables were previously used by the scripts that are now part of the [`generate-builder`](https://github.com/slsa-framework/slsa-github-generator/tree/main/.github/actions/generate-builder) action when the script was invoked directly by the workflow itself.

### DIFF
--- a/.github/workflows/generator_container_slsa3.yml
+++ b/.github/workflows/generator_container_slsa3.yml
@@ -16,18 +16,8 @@ name: SLSA container image provenance
 
 env:
   # Generator
-  BUILDER_BINARY: generic-generator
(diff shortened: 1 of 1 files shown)


laurentsimon (approved):
Thanks!

PR #622, opened 2022-07-23
Update tags for renovatebot

laurentsimon:
Signed-off-by: laurentsimon <64505099+laurentsimon@users.noreply.github.com>


loosebazooka (approved):

.github/workflows/pre-submit.units.yml
```diff
@@ -65,12 +65,12 @@ jobs:
     if: ${{ always() }}
     steps:
       - name: Checkout
-        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # v2.3.4
+        uses: actions/checkout@ec3a7ce113134d7a93b817d10a8272cb61118579 # tag=v2
```
ianlewis, line 68: nit: use `tag=v2.3.4`

This lets us know what the actual version is and renovate bot should still send us upgrade PRs even if we reference a specific version. Here's an example: https://github.com/slsa-framework/slsa-github-generator/pull/513/files

ianlewis (approved):

.github/workflows/pre-submit.units.yml
laurentsimon: I thought renovatebot would do that for us automatically. But apparently it does not. Will update, thanks!

PR #632, opened 2022-07-26
Runner API

ianlewis:
Adds a simple `CommandRunner` that runs commands using `syscall.Exec`. Env vars and working dir are properly captured for each command and can be included directly in `buildConfig` in provenance for various build types.

This can be used to support the Go builder as well as future builders like containers, ko, etc.

### DIFF
--- a/internal/runner/runner.go
+++ b/internal/runner/runner.go
@@ -1,0 +1,143 @@
+package runner
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+	"syscall"
+)
+
+// CommandRunner runs commands and returns the build steps that were run.
+type CommandRunner struct {
+	// Env is global environment variables passed to all commands.
+	Env []string
+
+	// Steps are the steps to execute.
+	Steps []*CommandStep
+}
+
+// CommandStep is a command that was executed by the builder.
+type CommandStep struct {
+	// Command is the command that was run.
+	Command []string `json:"command"`
+
+	// Env are the environment variables passed to the command.
+	Env []string `json:"env"`
+
+	// WorkingDir is the working directory the command was executed in.
+	WorkingDir string `json:"workingDir"`
+}
+
+// Run excutes a series of commands and returns the steps that were executed
+// successfully. Commands are run in sequence and are expected to return a zero
+// exit status.
+//
+// Global environment variables are merged with steps environment variables in
+// the returned steps. In the case of duplicates the last occurrence has precidence.
+// Environment variables are *not* inherited from the current process.
+//
+// The returned CommandSteps should be included in the buildConfig provenance.
+// These are *not* the same as the runner commands. Env vars are sanitized, pwd
+// is changed to the absolute path, and only commands that executed
+// successfully are returned.
+func (r *CommandRunner) Run() (steps []*CommandStep, err error) {
+	var originalwd string
+	originalwd, err = os.Getwd()
+	if err != nil {
+		return // steps, err
+	}
+	defer func() {
+		// Change directories back to the original working directory but only
+		// return the error returned by Chdir if no other error occurred.
+		if chDirErr := os.Chdir(originalwd); err == nil {
+			// NOTE: err is returned by the function after the defer is called.
+			err = chDirErr
+		}
+	}()
+
+	for _, step := range r.Steps {
+		if len(step.Command) == 0 {
+			err = errors.New("command is empty")
+			return // steps, err
+		}
+
+		cmd := step.Command[0]
+		args := step.Command[1:]
+		// TODO: Add some kind of LD_PRELOAD protection?
+		env := dedupEnv(append(r.Env, step.Env...))
+
+		// Set the POSIX PWD env var.
+		var pwd string
+		pwd, err = filepath.Abs(step.WorkingDir)
+		if err != nil {
+			return // steps, err
+		}
+		env = append(env, "PWD="+pwd)
+
+		err = os.Chdir(pwd)
+		if err != nil {
+			return // steps, err
+		}
+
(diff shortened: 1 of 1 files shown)


internal/runner/runner.go
```diff
@@ -0,0 +1,148 @@
...
+	if err = os.Chdir(pwd); err != nil {
+		return nil, err
+	}
+
+	// NOTE: We use syscall.Exec directly because we are executing the
+	// commands in sequence and we want full control over the environment.
+	// The stdlib exec package deduplicates env vars and so it's hard to
+	// know exactly what the command was run with.
+	// NOTE: The executed command will inherit stdout/stderr.
+	if err := syscall.Exec(cmd, args, env); err != nil {
```
laurentsimon, line 96: I think this command will never return, because exec overwrite the address space of the process. I think this needs to be fork-then-exec instead. Let me know if I mis-understood.

internal/runner/runner.go
```diff
@@ -0,0 +1,148 @@
...
+	env = append(env, "PWD="+pwd)
+
+	if err = os.Chdir(pwd); err != nil {
+		return nil, err
+	}
+
+	// NOTE: We use syscall.Exec directly because we are executing the
+	// commands in sequence and we want full control over the environment.
+	// The stdlib exec package deduplicates env vars and so it's hard to
+	// know exactly what the command was run with.
```
laurentsimon, line 94: this command is controlled by the developer, so we cannot trust what it returns, correct? How do we plan to use the returned values?

internal/runner/runner.go
ianlewis: Yeah, I thought that too but I decided to try it since that was what was being used in the Go builder.
https://github.com/slsa-framework/slsa-github-generator/blob/9875820c239da24d02e0d3b6da405beed814246a/internal/builders/go/pkg/build.go#L164

After looking at the runtime code and testing it out it's pretty clear that the Go runtime is just running `execve` though.

In the case of the go builder it kind of works since it's the last thing the Go program really does. But it's unexpected the way it works for things like tests as these checks never actually get run if dry is `false`.
https://github.com/slsa-framework/slsa-github-generator/blob/9875820c239da24d02e0d3b6da405beed814246a/internal/builders/go/pkg/build_test.go#L1155-L1165

Probably, just using the `os/exec` package makes the most sense since it's a bit annoying to implement fork/exec yourself. `os/exec` does do a bit of changes to the pwd and env vars it's given so it's hard to know what the program was actually run with.

internal/runner/runner.go
ianlewis: I'm not sure what you're referring to. Do you mean stderr/stdout? Right now I'm thinking that the only thing that would be used would be the `error` value returned. The only real value that impacts that would be the command's exit code. stdout/stderr would just get printed for debugging purposes.

internal/runner/runner.go
ianlewis: Probably [`os.StartProcess`](https://pkg.go.dev/os#StartProcess) is what I actually want to use.

laurentsimon: I 100% agree that implementing fork-then-exec is dangerous and prone to mistakes. If os.StartProcess does that for us, awesome!

internal/runner/runner.go
laurentsimon: I may have added my comment in the wrong place. I meant that:
```
return &CommandStep{
		Command:    append([]string{cmd}, args...),
		Env:        env,
		WorkingDir: pwd,
	}, nil
```

seemed to return the steps to add in the provenance. Since it's never called, I think that was more a test than anything. So you can probably ignore my comment :)

ianlewis: So to be clear,

- step output will go to stdout but not parsed by the runner API or builder in any way.
- Only error code is checked to determine if the step was executed successfully or not
- We return separate command steps than those that are given in order to fill in values like working directory and some default environment variables to most accurately describe how the command was executed.

internal/runner/runner.go
ianlewis: I decided to just go with using os/exec

In Go 1.19 they added a `cmd.Environ` method that allows us to get the environment that is used by the command. We can use this to get a more accurate set of env vars that os/exec uses once we update to Go 1.19.

internal/builders/go/pkg/build.go
```diff
@@ -116,25 +120,41 @@ func (b *GoBuild) Run(dry bool) error {
...
+			},
+		}
+
+		steps, err := r.Dry()
 		if err != nil {
 			return err
 		}
 
-		menv, err := utils.MarshalToString(env)
+		menv, err := utils.MarshalToString(steps[0].Env)
```
laurentsimon, line 143: I know it' obvious, but maybe add a comment that there's a single command in the steps, which is why we're using `[0]`

internal/runner/runner.go
```diff
@@ -0,0 +1,145 @@
...
+	env := make([]string, len(r.Env), len(r.Env)+len(step.Env)+1)
+	copy(env, r.Env)
+	env = append(env, step.Env...)
+
+	// Set the POSIX PWD env var.
+	pwd, err := filepath.Abs(step.WorkingDir)
+	if err != nil {
+		return nil, err
+	}
+	env = append(env, "PWD="+pwd)
```
laurentsimon, line 115: without this, it would run in `$HOME`?

laurentsimon (approved):

internal/runner/runner.go
ianlewis: No, but I think that some programs or scripts could depend on `PWD` being set so this is just here for completeness.

internal/builders/go/pkg/build.go
ianlewis: Done.

PR #640, opened 2022-07-29
Fix builder dir in container workflow

ianlewis:


.github/workflows/generator_container_slsa3.yml
```diff
@@ -89,7 +89,7 @@ jobs:
           binary: "${{ env.BUILDER_BINARY }}"
           compile-builder: "${{ inputs.compile-generator }}"
           # NOTE: We are using the generic generator.
-          directory: "${{ env.BUILDER_DIR }}/generic"
+          directory: "${{ env.BUILDER_DIR }}"
```
asraa, line 92: how come this is needed when builder_dir is already
`  BUILDER_DIR: internal/builders/generic # Source directory if we compile the builder.`?

asraa (approved):

PR #651, opened 2022-07-31
New: Ensure code coverage is met

naveensrinivasan:
- This will enable the desired code coverage is met for the project
- The coverage is set to 70 to start off with. This setting is in the
  Makefile
- The coveragethreshold.json is an override for packages which have
  different coverage needs from the global coverage threshold.
- The code coverage tool uses standard go tool.

### DIFF
--- a/hack/codecoverage/README.md
+++ b/hack/codecoverage/README.md
@@ -4,9 +4,8 @@ The goal of the coverage tool is to measure the coverage of the code base for Go
 
 ## Usage
 Execute the following command to get the coverage and store it in a file:
-1. `go test  -coverprofile=coverage ./... | THRESHOLD_FILE=./coverage.json COVERAGE_PERCENTAGE=70 go run ./hack/codecoverage/main.go`
-2. The `THRESHOLD_FILE` is the path to the file containing the coverage threshold.
-3. The THRESHOLD_FILE contains the percentage of the code coverage that is required to pass for certain packages. This is usually because they don't match the desired coverage.
+1. `go test  -coverprofile=coverage ./... |  run ./hack/codecoverage/main.go coverage.json 70`
+2. The coverage.json contains the percentage of the code coverage that is required to pass for certain packages. This is usually because they don't match the desired coverage.
 ```json
 {
   "github.com/slsa-framework/slsa-github-generator/github":70.4,
--- a/hack/codecoverage/README.md
+++ b/hack/codecoverage/README.md
@@ -1,0 +1,35 @@
+# Go Coverage tool
+
+The goal of the coverage tool is to measure the coverage of the code base for Golang.
+
+## Usage
+Execute the following command to get the coverage and store it in a file:
+1. `go test  -coverprofile=coverage ./... | THRESHOLD_FILE=./coverage.json COVERAGE_PERCENTAGE=70 go run ./hack/codecoverage/main.go`
+2. The `THRESHOLD_FILE` is the path to the file containing the coverage threshold.
+3. The THRESHOLD_FILE contains the percentage of the code coverage that is required to pass for certain packages. This is usually because they don't match the desired coverage.
+```json
+{
+  "github.com/slsa-framework/slsa-github-generator/github":70.4,
+  "github.com/slsa-framework/slsa-github-generator/internal/builders/generic": 52.3,
+  "github.com/slsa-framework/slsa-github-generator/internal/builders/go":17.1,
+  "github.com/slsa-framework/slsa-github-generator/internal/errors":100.0,
+  "github.com/slsa-framework/slsa-github-generator/internal/utils":72.1,
+  "github.com/slsa-framework/slsa-github-generator/signing/envelope":82.4,
+  "github.com/slsa-framework/slsa-github-generator/slsa":54.6
+}
+```
+3. The `COVERAGE_PERCENTAGE` is the percentage of the code coverage that is required to pass for all the packages except the ones that are mentioned in the `THRESHOLD_FILE`.
+4. The coverage tool will fail if the coverage is below the threshold for any package.
+``` shell
+2022/07/29 16:14:41 github.com/slsa-framework/slsa-github-generator/pkg/foo is below the threshold of 71.000000
+exit status 1
+```
+
+### Design choices
+
+1. The coverage tool should not depend on any other tools. It should work of the results from the `go test` command.
+2. Coverage threshold should be configurable for each repository - for example `70%` within the repository.
+3. A setting file should override the coverage threshold for a given package within the repository. `github.com/foo/bar/xyz : 61`
+4. The coverage tool should use native `go` tools and shouldn't depend on external vendors.
+5. The coverage tool should be configurable as part of the PR to fail if the desired threshold is not met.
+6. Contributors should be able to run it locally if desired before doing a PR.
--- a/hack/codecoverage/main.go
+++ b/hack/codecoverage/main.go
@@ -23,19 +23,22 @@ import (
 )
 
 func main() {
-	thresholdFile := os.Getenv("THRESHOLD_FILE")
+	if len(os.Args) != 3 {
+		log.Fatal("Usage: <threshold-file> <coverage-percent>")
+	}
+	thresholdFile := os.Args[1]
 	if thresholdFile == "" {
-		log.Fatalf("THRESHOLD_FILE environment variable is not set")
+		log.Fatalf("The thresholdfile cannot be empty.")
 	}
 	thresholdMap, err := parseCoverageThreshold(thresholdFile)
 	if err != nil {
 		log.Fatalf("Error parsing threshold file: %v", err)
 	}
-	coveragePercentage := os.Getenv("COVERAGE_PERCENTAGE")
+	coveragePercentage := os.Args[2]
 	if coveragePercentage == "" {
-		log.Fatalf("COVERAGE_PERCENTAGE environment variable is not set")
+		log.Fatalf("The coverage percentage cannot be empty.")
 	}
-	coveragePercentageFloat, err := strconv.ParseFloat(coveragePercentage, 32)
+	coveragePercentageFloat, err := strconv.ParseFloat(coveragePercentage, 64)
 	if err != nil {
 		log.Fatalf("Error parsing coverage percentage: %v", err)
 	}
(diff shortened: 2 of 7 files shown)


hack/codecoverage/main.go
```diff
@@ -0,0 +1,87 @@
...
+	"bufio"
+	"encoding/json"
+	"log"
+	"os"
+	"strconv"
+	"strings"
+)
+
+func main() {
+	thresholdFile := os.Getenv("THRESHOLD_FILE")
```
ianlewis, line 26: nit: I think it would be more natural to have `THRESHOLD_FILE` and `COVERAGE_PERCENTAGE` be flags.

hack/codecoverage/main.go
```diff
@@ -0,0 +1,87 @@
...
+			parts := strings.Fields(line)
+			if len(parts) < 5 {
+				continue
+			}
+			percentage, err := strconv.ParseFloat(strings.Trim(parts[4], "%"), 32)
+			if err != nil {
+				log.Fatalf("invalid line: %s", line)
+			}
+			pack := parts[1]
+			if val, ok := thresholdMap[pack]; !ok {
```
ianlewis, line 56: nit: if you're going to use an if/else then have the positive code first. Otherwise else becomes a double-negative.

i.e.
```golang
if val, ok := thresholdMap[pack]; ok {
    // value is in thresholdMap
    ...
} else {
    // value is not in thresholdMap
    ...
}
```

hack/codecoverage/main.go
```diff
@@ -0,0 +1,87 @@
...
+				if float32(int(percentage*100)/100) < float32(int(coveragePercentageFloat*100)/100) {
```
ianlewis, line 57: If percentage is a float is there really a need to convert to int? They way you have it here only the first 2 decimal places are used even though the `coveragethreshold.json` could have values with more decimal places.

Would something like this not work? I think they would both be float64.
```golang
if percentage < coveragePercentageFloat {
   ...
}
```

hack/codecoverage/main.go
```diff
@@ -0,0 +1,87 @@
...
+	// read stream from stdin
+	scanner := bufio.NewScanner(os.Stdin)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.Contains(line, "coverage: ") {
+			parts := strings.Fields(line)
+			if len(parts) < 5 {
+				continue
+			}
+			percentage, err := strconv.ParseFloat(strings.Trim(parts[4], "%"), 32)
```
ianlewis, line 51: nit: why not just parse 64 bits rather than 32?

hack/codecoverage/main.go
naveensrinivasan: I agree. I have changed it. Thanks

hack/codecoverage/main.go
naveensrinivasan: I agree. I have changed it. Thanks

hack/codecoverage/main.go
naveensrinivasan: I agree. I have changed it. Thanks

hack/codecoverage/main.go
naveensrinivasan: I agree. I have changed it. Thanks

hack/codecoverage/README.md
```diff
@@ -0,0 +1,34 @@
+# Go Coverage tool
```
laurentsimon, line 1: why is the folder called "hack"?

laurentsimon (commented):
I like the idea of improving test coverage, but I have somme high-level questions:

1. A coverage percentage is not very actionable for developers. Would it not be better to provide more actionable information in a PR for both reviewers and PR author? For example, annotation / information about functions added in a PR that are not hit by unit tests, that would be really useful to start with.
1. The config file is going to go out-of-sync as soon as we add files, etc. 
2. Should this feature live in an external GHA rather than in this repo?
3. Making this blocking on PRs is probably just going to lead to the PR author reducing the threshold, since it's hard to tell which lines / functions are not tested.

hack/codecoverage/README.md
naveensrinivasan: > A coverage percentage is not very actionable for developers. Would it not be better to provide more actionable information in a PR for both reviewers and PR author? For example, annotation / information about functions added in a PR that are not hit by unit tests, that would be really useful to start with.

That is way harder, and there are existing tools like VSCode, and Go compiler which can do that. 

> The config file is going to go out-of-sync as soon as we add files, etc.

Yes, if we don't meet 70% coverage. 

> Should this feature live in an external GHA rather than in this repo?

As a dev, I would like to run the checks locally, and this gives that option

>Making this blocking on PRs is probably just going to lead to the PR author reducing the threshold, since it's hard to tell which lines / functions are not tested.

The goal is to figure out the code coverage without using external dependencies. The code coverage is provided by VSCode and Go compiler. This isn't the design of this tool. Only then the code coverage is met. High-quality tests are critical for having confidence.


Right now, we are flying blind without knowing what the coverage of the project is. Any time new code gets added/updated, the project or the PR reviewer isn't aware of the test coverage.

What would be your suggestion to address this concern? Something like codecov would also suffice. Do you think we should enable an online tool?

laurentsimon: > As a dev, I would like to run the checks locally, and this gives that option

any reason why running `go test` locally does not suffice? 

> What would be your suggestion to address this concern? Something like codecov would also suffice. Do you think we should enable an online tool?

If codecov does it and they don't require write permissions to the repo, that sounds like an easy win

naveensrinivasan: >If codecov does it and they don't require write permissions, that sounds like an easy win, no?

@bobcallaway Do you have thoughts on codecov or similar other online tools?

asraa: One thing we can do is open an HTML report of code coverage per file using the native go tools. Generate the coverage profile with go test, and then run `go tool cover`. We can make this a non-gating CI that allows reviewers (who ultimately should decide what can/cant be testable) to see if newly added code was tested. 

The CI job can also add a comment decomenting coverage % in each pkg. If all of this can be done without new code to maintain, that would be great.

e.g.
```
go test -v -coverprofile cover.out ./YOUR_CODE_FOLDER/...
go tool cover -html=cover.out -o cover.html
open cover.html
```

laurentsimon: +1 on having nothing to maintain ourselves.

naveensrinivasan: How do you make sure there is high test coverage without failing CI? Leaving it up to the reviewers to decide isn't going to get high enough coverage. 

Also, the PR requires 70% coverage. If the Reviewer agrees that 70% isn't attainable, then the reviewer can ask the PR author to update the file with different coverage.

bobcallaway: Codecov does have the ability to print an easy-to-read analysis of what the proposed change in the PR did to overall coverage, but it does not enforce a minimum level. Unless reviewers are actively evaluating the coverage (from codecov, the tool in this PR, or something else) then I would argue the measurement is being treated as noise.

laurentsimon: We have had good coverage without imposing a threshold so far. We'd like to keep it that way. Going back and forth between the reviewer and author to meet an arbitrary threshold is not worth it. 

What you describe about `the reviewer can ask the PR author to update the file with different coverage` is going to become the default behavior if we set arbitrary threshold.

laurentsimon: > Codecov does have the ability to print an easy-to-read analysis of what the proposed change in the PR did to overall coverage, but it does not enforce a minimum level. Unless reviewers are actively evaluating the coverage (from codecov, the tool in this PR, or something else) then I would argue the measurement is being treated as noise.

that's exactly the problem.

bobcallaway: I wouldn't say the threshold is arbitrary, I think it's more about setting a floor and then continuing to work upwards?

naveensrinivasan: >  Imposing arbitrary threshold is not.

> Although there is no “ideal code coverage number,” at Google we offer the general guidelines of 60% as “acceptable”, 75% as “commendable” and 90% as “exemplary.” However we like to stay away from broad top-down mandates and encourage every team to select the value that makes sense for their business needs.


source for ☝️  https://testing.googleblog.com/2020/08/code-coverage-best-practices.html

I think this blog post summarizes well.

laurentsimon: `A high code coverage percentage does not guarantee high quality in the test coverage` 
`More important than the percentage of lines covered is human judgment over the actual lines of code (and behaviors)  that aren’t being covered` 

Qualitative feedback is nice. Imposing threshold does not help IMO. I've seen this in scorecard: when the coverage goes down in a PR, we just let the PR go thru because that's the best we can do. Over time we don't even look at it because it's just noise and not actionable.

Having a general view of code coverage is great. Having it in PR as a percentage number provides no benefits IMO.

naveensrinivasan: OK cool! I understand if you have strong opinions on this 👍. I will close this PR.

naveensrinivasan:
https://github.com/slsa-framework/slsa-github-generator/pull/651#discussion_r935027013

hack/codecoverage/README.md
bobcallaway: I'm not sure I understand your argument @laurentsimon - `we just let the PR go thru because thats the best we can do` - why not ask the contributor to add unit test coverage for what they're proposing to not erode the baseline percentage?

ianlewis: 1. I think having a floor is ok. Maybe somewhere around 50% minimum (though the 60% Google number is also fine)? But not necessarily any targets.
2. We should strive to ask for tests on PRs when appropriate.

I can go either way on making this a blocking pre-submit, but I agree there are merits either way.

> I'm not sure I understand your argument @laurentsimon - `we just let the PR go thru because thats the best we can do` - why not ask the contributor to add unit test coverage for what they're proposing to not erode the baseline percentage?-

I think that the biggest problem would be for a change that doesn't necessarily merit unit-tests or where unit-tests are especially hard to write, it might make doing it on that particular PR overly expensive just to maintain an arbitrary coverage baseline. We also have e2e tests that might cover certain scenarios better than unit-tests.

A better rule might be that we need to hit a baseline coverage % before we do a release?

asraa: > A better rule might be that we need to hit a baseline coverage % before we do a release?

This is a good thing to enforce on "pre-release" -- if we do this it would be good to have a report generated per PR because accumulating tests at the end is a lot harder than per-change. 

> why not ask the contributor to add unit test coverage for what they're proposing to not erode the baseline percentage?

I think the problem was that per-file doesn't give reviewer or author enough actionable guidance. I still like the HTML report generated to help the reviewer evaluate test coverage of any added lines.

Issue #710, opened 2022-08-12
[feature] Add a release URL to builders and generators for uploading assets


asraa:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
For some workflow triggers like `workflow_dispatch`, the `github.ref` does not indicate the release URL. We should expose an option like 
```
          upload_url: ${{ steps.create-release.outputs.upload_url }}
```

that can be used to indicate the release URL to upload assets to.

That would simplify the workflow, and reduce the amount of friction for users. Otherwise, they need to specify their own upload, or manually upload the artifact themselves.

See https://github.com/GoogleContainerTools/jib/commit/e85d34e1661acf43ce078412719a2c2ce1a42f59

**Describe alternatives you've considered**

**Additional context**

laurentsimon:
Repositories to update:
-[ ] https://github.com/GoogleContainerTools/jib/pull/3726

laurentsimon:
We also need e2e tests for this.

laurentsimon:
See this PR which has the code to use: https://github.com/GoogleContainerTools/jib/pull/3726/files

ianlewis:
> See this PR which has the code to use: https://github.com/GoogleContainerTools/jib/pull/3726/files

This PR uses `actions/upload-release-asset` which isn't maintained anymore in favor of `softprops/action-gh-release` which is the action we use. Maybe the right input we need use is `tag_name`?
https://github.com/softprops/action-gh-release#inputs

laurentsimon:
you're correct. Note that I only re-used the GHA they already use in their code, it's not something I asked them to use. 
I've created https://github.com/slsa-framework/slsa-github-generator/pull/713 to own the code for us internally, and maybe other repos could use it?

hazcod:
Can you mention here what release it'll end up in @laurentsimon ? Thanks!

laurentsimon:
Should be part of the `[2022 Stability improvements](https://github.com/slsa-framework/slsa-github-generator/milestone/6)`. Thanks for the interest! Which project do you want to use it for?

hazcod:
> Should be part of the `[2022 Stability improvements](https://github.com/slsa-framework/slsa-github-generator/milestone/6)`. Thanks for the interest! Which project do you want to use it for?

Thanks! I'm writing an SDK for the API on Intigriti and want to ensure we do all of our due diligence.

PR #713, opened 2022-08-13
feat: upload provenance to release via a `tag-name`

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/710


.github/workflows/generator_generic_slsa3.yml
```diff
@@ -154,9 +158,36 @@ jobs:
           path: "${{ inputs.attestation-name }}"
           sha256: "${{ needs.generator.outputs.attestation-sha256 }}"
 
-      - name: Release
+      - name: Upload provenance
```
asraa, line 161: I'm a little worried people will assume they need `upload-assets` to be true when they specify an `upload-url` 

Maybe we can clarify that in the description? Or allow that case, and only run this job when `upload-url` is nil.

laurentsimon: I agree it's confusing. How about requiring `upload-assets` be true *always* to an upload to occur? 

If `upload-assets`, the upload will occur automatically on new tags. If not new tag, users need this optional `upload-url` to indicate where to upload. 

Wdut?

laurentsimon: I've updated the doc to show what it would look like for the suggestion I made.

.github/workflows/builder_go_slsa3.yml
```diff
@@ -37,10 +37,19 @@ on:
         required: true
         type: string
       upload-assets:
-        description: "If true provenance is uploaded to a GitHub release for new tags."
+        description: >
```
laurentsimon, line 40: @olivekl any feedback on the change in the file?
We're adding a new option called `upload-url` in this PR.
`upload-assets` is the option we already had, and it indicates that the workflow should automatically upload the provenance / binaries on new tags / releases.

`upload-url` is a new option for users to pass an upload url, because they sometimes create the release themselves in their workflow. In this case the workflow does not run on new release, it runs on a trgger event like `workflow_dispatch` (manual trigger) and the workflow creates the release before handing over the execution to our sSLSA generator.

/cc @asraa

.github/workflows/builder_go_slsa3.yml
```diff
@@ -309,3 +318,45 @@ jobs:
           files: |
             ${{ needs.build-dry.outputs.go-binary-name }}
             ${{ needs.provenance.outputs.go-provenance-name }}
+
+  upload-url:
+    runs-on: ubuntu-latest
+    needs: [generator]
+    permissions:
+      contents: write
+    if: inputs.upload-url != ''
```
asraa, line 327: check if `upload-assets` is true?

.github/workflows/builder_go_slsa3.yml
```diff
@@ -282,7 +291,7 @@ jobs:
   #           Upload binaries and provenances as assets             #
   #                                                                 #
   ###################################################################
-  upload-assets:
+  upload-tag:
```
asraa, line 294: what if this is run with `upload-url` specified too? Should we double check that it is not?

asraa (commented):
Thanks! I like this change. Let me know about the failsafing comment

.github/workflows/builder_go_slsa3.yml
laurentsimon: done.

.github/workflows/builder_go_slsa3.yml
laurentsimon: I've left it as-is to keep it simple.
If `upload-assets: true`, always upload for tags
if `upload-url` is also set, upload to the URL. This way is a user wants to upload to a different url from the tag, they can. And if they pass both, we will upload twice... which still works.

Wdut?

.github/workflows/generator_generic_slsa3.yml
```diff
@@ -154,9 +163,36 @@ jobs:
...
         uses: softprops/action-gh-release@1e07f4398721186383de40550babbdf2b84acfc5 # tag=v0.1.14
         id: release
         with:
           files: |
             ${{ inputs.attestation-name }}
```
ianlewis, line 171: So when `upload-assets=true` and `upload-url` is set we upload twice? once as we do now (if this is a `workflow_dispatch` I guess this is a noop?), and once at the upload url specified?

.github/actions/upload-assets/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "upload-release-asset",
```
ianlewis, line 2: upload-assets? or maybe rename the action's directory to `upload-release-asset`?

.github/workflows/generator_generic_slsa3.yml
```diff
@@ -154,9 +163,36 @@ jobs:
...
+      - name: Download the provenance
+        uses: slsa-framework/slsa-github-generator/.github/actions/secure-download-artifact@e178bc7b362c7cb93899a9721da7a0d7d1286869
+        with:
+          name: "${{ inputs.attestation-name }}"
+          path: "${{ inputs.attestation-name }}"
+          sha256: "${{ needs.generator.outputs.attestation-sha256 }}"
+
+      - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # tag=v3.0.2
+      - name: Upload provenance
+        uses: ./.github/actions/upload-assets
```
ianlewis, line 191: What was the reason we needed to create our own action for this?

.github/workflows/generator_generic_slsa3.yml
laurentsimon: you''re correct. I've tried to not add much logic here, since the case you describe will hardly ever happen... I argue :)

But I can update if you think it's important.

Just let me know.

.github/actions/upload-assets/package.json
laurentsimon: will do thanks

.github/workflows/generator_generic_slsa3.yml
laurentsimon: There is no available and maintained action to do that. It seems to not be a common case. The action that job project uses is https://github.com/actions/upload-release-asset, but it's not maintained. I've just copied their code and updated to latest dependency versions.

.github/workflows/generator_generic_slsa3.yml
ianlewis: Based on your other comments it seemed like `upload-url` would only have an effect if `upload-assets=true` so I'm a bit confused.

Just to be clear, we expect them to set one or the other, but not both?

.github/workflows/generator_generic_slsa3.yml
ianlewis: Since the normal supported wa to upload to releases is to use the `softprops/action-gh-release` and that action has a `tag_name` in order to support this use case, I wonder if we should just accept a `tag-name` input instead of `upload-url` and pass it along to `softprops/action-gh-release` (that we already call)?

ianlewis: I fear we are adding a new action that we need to support just because our user jib is using the `upload-url` option on the unsupported action. I'm sure jib could update to use a `tag-name` to specify the release on `workflow_dispatch` events.

.github/workflows/generator_generic_slsa3.yml
laurentsimon: so we expect `upload-assets=true` to always be set, optionally with a url

.github/workflows/generator_generic_slsa3.yml
laurentsimon: Great idea. I'll re-work this PR then. Thanks for the suggestion

laurentsimon: Changes made. PTAL

.github/workflows/generator_generic_slsa3.yml
laurentsimon: no longer relevant as per https://github.com/slsa-framework/slsa-github-generator/pull/713#discussion_r960293407

.github/workflows/builder_go_slsa3.yml
```diff
@@ -37,10 +37,21 @@ on:
         required: true
         type: string
       upload-assets:
-        description: "If true provenance is uploaded to a GitHub release for new tags."
+        description: >
+          If true provenance is uploaded to a GitHub release for new tags.
+          When the workflow does not run on a new tag, such as on a workflow_dispatch,
+          the `upload-url` argument must be provided as well.
```
ianlewis, line 43: Update this since we don't have an `upload-url` argument anymore.

.github/workflows/generator_generic_slsa3.yml
```diff
@@ -175,9 +190,38 @@ jobs:
...
+        if: inputs.upload-tag-name != ''
+        id: release-tag-name
         with:
+          tag_name: "${{ inputs.upload-tag-name }}"
           files: |
             ${{ needs.generator.outputs.provenance-name }}
+
+      - name: Output release ID
+        shell: bash
+        if: release
```
ianlewis, line 212:
```suggestion
        id: release
```

ianlewis (approved):

laurentsimon:
@ianlewis yamlling seems to be failing. Does
```
Successfully installed pathspec-0.10.1 yamllint-1.26.3
Error: [trailing-spaces] trailing spaces
```

means it found a problem with a yaml file? Or is it failing itself?

ianlewis:
> @ianlewis yamlling seems to be failing. Does
> 
> ```
> Successfully installed pathspec-0.10.1 yamllint-1.26.3
> Error: [trailing-spaces] trailing spaces
> ```
> 
> means it found a problem with a yaml file? Or is it failing itself?

The error shows up in the PR diff view. I don't think the output shows up in the log very well for some reason. It seems to omit the file name and line number.
https://github.com/slsa-framework/slsa-github-generator/pull/713/files#diff-a526f8e747aef206fec2ce8469bbc904ee1a8d047bbb486743cca4e2dd5e5037R175

.github/workflows/generator_generic_slsa3.yml
```diff
@@ -157,16 +172,16 @@ jobs:
           if-no-files-found: error
           retention-days: 5
 
-  # create-release creates a release and uploads provenance to the release if
-  # `upload-assets` is true.
-  create-release:
+  # upload-assets uploads provenance to the release
```
ianlewis, line 175: Fix trailing spaces
```suggestion
  # upload-assets uploads provenance to the release
```

ianlewis (approved):

.github/workflows/pre-submit.actions.yml
```diff
@@ -9,7 +9,7 @@ permissions: read-all
 
 jobs:
   checkout:
-    name: verify no checkout in Actions
```
ianlewis, line 12: If the name of a job changes then the protected branch's required pre-submits need to change as well. Ideally GitHub would allow you to set required pre-submits by an ID rather than name but it seems to use the name. We should think of job names as IDs.

I'm assuming that you're changing the name here just to add capitalization for consistency's sake? It's kind of hard keeping track of which of these are required for pre-submits so I suggest reverting the names and adding a comment that they shouldn't be changed unless there is a really good reason.

.github/workflows/pre-submit.units.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           make unit-test
 
   check-verifier:
-    name: verify slsa-verifier is latest
```
ianlewis, line 35: See other comment about required pre-submits

asraa:
ping -- does this need another review after a merge?

laurentsimon:
Let me get back to this PR when I'm back. It's been a while so I need to look into it. Thanks for pinging!

PR #758, opened 2022-08-23
contents: write is required for the generic builder

sethmlarson:
Signed-off-by: Seth Michael Larson <sethmichaellarson@gmail.com>

PR #761, opened 2022-08-24
Fix permissions in generic workflow doc

ianlewis:
Fix permissions used by the generic workflow in docs. `contents: write` is needed for writing artifacts to releases. This continues the doc fix that was made in #758 


laurentsimon (approved):
Thanks!

Issue #822, opened 2022-09-07
[discussion] Include Environment Variables


ianlewis:
# Background:

For reproducibility, build steps are included in provenance by some builders. These build steps include information to reproduce the command, such as the working directory and environment variables.

POSIX [defines some environment variables](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html) which mirror values set elsewhere. These include the `PWD`, `USER`, and `HOME` environment variables. Some programs may use the environment variables as a way to determine the current working director or current user for example.

Some values, such as the current working directory, are set in a separate field from the environment variables and thus the information in the environment variables would be redundant.

# Question

Should these environment variables be included in the build steps in the provenance? If so, which ones?

ianlewis:
Currently we are setting the `PWD` environment variable but are leaning towards removing it since it's redundant info and we can say that our `buildConfig` implicitly indicates that it should be set when reproducing the build step.

laurentsimon:
@TomHennen @MarkLodato @joshuagl @msuozzo

mlieberman85:
+1 to removing it.  Does https://github.com/slsa-framework/slsa-github-generator/generic@v1 have a defined spec yet? Just making that clear in the spec would be useful.

laurentsimon:
We don't have a "spec", just examples https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/README.md#provenance-format

For go builder we have *some* explanation in https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/go#buildconfig-format, but it's probably not enough to explain what is recorded, what is not, and maybe take the view of a re-builder?

Great idea to have a spec, thanks! All the builders will have the same output, so it should be easy. 

What would you like to see captured in the specs?

shaunmlowry:
I think it's dangerous to start arbitrarily removing environment variables from provenance just because they're redundant on a specific platform (or set of platforms).  What about build steps that run on Windows where PWD isn't a thing, and might validly be used for some user-defined purpose by such a build step?

laurentsimon:
Note that we have not released our builders with `PWD` set, this is just a change we made a few days ago, but it's not released, which is why we're removing it.

I think you also need more context: we don't record env variables, we only record the ones the user asked us to set, and we don't record those that exist "by default". The user never asked us to set `PWD` in this case.

[...]

PR #825, opened 2022-09-08
Remove PWD from provenance env

ianlewis:
Updates #822 

Remove PWD from the environment variables in provenance build steps as it is redundant with the given working directory. `PWD` is set when executing the command but is not included in the environment variables that are added to the `buildConfig` in the provenance.

Before:
```json
"steps": [
    {
        "command": ["bash", "-c", "echo $STEP2"],
        "env": [
            "STEP2=fuga",
            "PWD=/home/user"
        ],
        "workingDir": "/home/user"
    },
    ...
]
```

After:
```json
"steps": [
    {
        "command": ["bash", "-c", "echo $STEP2"],
        "env": [
            "STEP2=fuga"
        ],
        "workingDir": "/home/user"
    },
    ...
]
```


asraa (approved):
LGTM -- CommandRunner contains posixEnv for Env but the CommandStep will render env without it for now. I see

ianlewis:
> LGTM -- CommandRunner contains posixEnv for Env but the CommandStep will render env without it for now. I see

Yes. I think we should set the variable but the question is whether it needs to be included in the provenance.

Issue #823, opened 2022-09-08
[bug]  slsa-goreleaser action fails on "Create and sign provenance"


gorsat:
I'm manually running a go-ossf-slsa3-publish.yml action in a private repo by pushing the "Run workflow" button on the actions page.
It fails on the step "Create and sign provenance"

Output from the failure:
```
>Run set -euo pipefail
provenance generator is slsa-builder-go-linux-amd64
getting entrypoint: getting workflow run: GET https://api.github.com/repos/gorsat/project_root/actions/runs/3011826089: (https://api.github.com/repos/gorsat/idco/actions/runs/3011826089:): 403 Resource not accessible by integration []
```
The only thing that is even slightly noteworthy about my config is that the file I'm building (yes, it's just one file) is in a subdir of the project root, so I have this in my .slsa-goreleaser.yml:
```
main: my-file.go
dir: ./my-sub-dir
```
In addition, I've commented out some of the lines marked "optional" in this file. 

But the error message looks like an authorization issue. So I'm wondering if the repo being private has something to do with it.

ianlewis:
@gorsat Thanks for submitting the issue! I believe you're right and it is caused by the repo being private. In the Go workflow we aren't requesting the `actions: read` scope for the step where we are detecting the workflows environment.

https://github.com/slsa-framework/slsa-github-generator/blob/11f701ad156cb8616ed5f5a444f1da6e2b77ed24/.github/workflows/builder_go_slsa3.yml#L78-L79

Unfortunately, We will need to update the workflow and create another release for this to work with private repos.

In the meantime you can give the generic generator starter workflow a try. This one *should* work as we request the right permissions.
https://github.com/gorsat/reponame/actions/new?category=none&query=slsa

laurentsimon:
One thing to know: the certificate with your repository name is uploaded to a transparency log for audit purposes. So by using this builder, you are revealing the name of your private repository.

Is that OK with you?

Let's add a disclaimer in the README to make this more explicit.

gorsat:
Thanks for the heads-up. Yes, that's okay for me, but, yeah, a disclaimer would be great, especially for businesses.

laurentsimon:
OK. So I propose

1. We amend the documentation
2. We add a `private-repository: true` option. At runtime we will check if the repo is private. If it is, we we only proceed if the option is set. Otherwise we will fail.

Wdut?

gorsat:
I like it.

Issue #34, opened 2022-04-20
Support private PKI


ianlewis:
It would be useful to support private PKI.
- Private rekor server.
- Private keys obtained from vault or something else?

steiza:
Yeah, so when I heard about this project I thought of it as made up of two steps:

1) Taking workload identity information (like from GitHub Actions) and encoding it into an `in-toto` document with `SLSA` predicates

2) Generating a signed attestation of the `in-toto` document, so it can be verified downstream

For (2), there are many advantages of using public `sigstore` infrastructure: individual projects don't have to manage PKI, there's a public transparency log, ... etc. The expectation is that this is what most open source projects would do.

But there might be cases where people don't want to use public `sigstore` infrastructure, maybe because the don't want their transparency logs to be public, or because they have their own PKI they way to use. In that situation, it would be nice if this project supported just generating the `in-toto` document (i.e. just step 1), or maybe even different options for signing other than public `sigstore` infrastructure (i.e. step 2).

That said, there is a lot of value in having an opinionated paved path, and maybe having different options for (2) adds more confusion than it's worth. There's nothing stopping people from implementing support for their own PKI as they need it!

ianlewis:
I think it's definitely something that could be considered and supported. In order to keep it simpler, maybe it's something we should wait on implementing until folks are specifically asking for it?

If we did implement it, we could allow folks to specify their own address for rekor/fulcio servers in a config file at the base of the repo like the one slsa-github-generator-go has.

asraa:
> we could allow folks to specify their own address for rekor/fulcio servers in a config file at the base of the repo like the one slsa-github-generator-go has.

The biggest challenge around here is verifying against the correct TUF root: folks would need to specify addresses for rekor/fulcio but also the trusted root CA and rekor public key they expect. That information would need to be pushed to the SLSA verifier as well.

I don't know the best solution for this yet. Do we specify the whole TUF root? On the generator side I suppose we only need to specify the rekor/fulcio, and the end-users need to know the TUF root.

ianlewis:
> The biggest challenge around here is verifying against the correct TUF root

How do sigstore tools handle this currently?

masaki-wise:
Hi folks,

I was wondering if there are any updates on supporting private PKI as mentioned in this issue? It would be great to know if there are any ongoing discussions or planned developments regarding this feature.

mlieberman85:
I don't think it ever got off the ground. Do you have a specific PKI use case?

PR #836, opened 2022-09-09
Add privacy-check action

ianlewis:
Updates #823 

Adds a `privacy-check` action to the repo. The action checks if the repository is private and fails if it is. Users can override this by explicitly setting the `override` input.

Workflows will use this action to check if repos are private in a subsequent PR because the action needs to be referenced by sha.


laurentsimon (approved):
LGTM. I'd like to have https://github.com/slsa-framework/slsa-github-generator/issues/823#issuecomment-1242062292 resolved before merging.

I think we need explicit opt- in from users who have private repos, otherwise we're leaking their repo name silently.

ianlewis:
> LGTM. I'd like to have [#823 (comment)](https://github.com/slsa-framework/slsa-github-generator/issues/823#issuecomment-1242062292) resolved before merging.
> 
> I think we need explicit opt- in from users who have private repos, otherwise we're leaking their repo name silently.

This is because the repo name appears in on the rekor transparency log? (probably another good reason why repo metadata on the cert isn't a great solution).

ianlewis:
From https://github.com/slsa-framework/slsa-github-generator/issues/823#issuecomment-1242062292 by @laurentsimon 

> One thing to know: the certificate with your repository name is uploaded to a transparency log for audit purposes. So by using this builder, you are revealing the name of your private repository.
> 
> Is that OK with you?
> 
> Let's add a disclaimer in the README to make this more explicit.

That's a good catch. Private PKI and transparency log can be supported more properly via #34

https://github.com/slsa-framework/slsa-github-generator/issues/823#issuecomment-1242100438 by @laurentsimon 

> OK. So I propose
> 
> 1. We amend the documentation
> 2. We add a `private-repository: true` option (or a better name). At runtime we will check if the repo is private. If it is, we we only proceed if the option is set. Otherwise we will fail.
> 
> Wdut?

I'll update this PR to support `private-repository` on all the workflows.

ianlewis:
@laurentsimon I updated this PR to just add a new `privacy-check` action that I will use in a later PR. Please see the updated PR description.

laurentsimon (approved):
LGTM. Thanks!
Please add a pre-submit to verify the "compilation" step gives the right `dist` output.

PR #850, opened 2022-09-14
Add call to privacy check to workflows

### LINKED ISSUE #823: [bug]  slsa-goreleaser action fails on "Create and sign provenance"

ianlewis:
Fixes #823 

- Adds call to the privacy-check to workflows. Each workflow gets a new `private-repository` input to override this check.
- The Go builder now has `actions: read` permissions in order to support private repos.

### DIFF
--- a/.github/workflows/builder_go_slsa3.yml
+++ b/.github/workflows/builder_go_slsa3.yml
@@ -74,7 +74,7 @@ jobs:
       - name: Check private repos
         uses: slsa-framework/slsa-github-generator/.github/actions/privacy-check@a3c7a56c8749c2c423f01bbcfd063315efc07a22
         with:
-          message: "Repository is private. The workflow has halted in order to keep the repository name from being exposed in the public transparency log. Set 'private-repository' to override."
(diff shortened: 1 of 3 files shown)


laurentsimon (approved):
Looks great!

PR #881, opened 2022-09-20
feat: Add npm builder workflow

laurentsimon:
This PR adds a new workflow `builder_node_slsa3.yml`. It contains the structure to build upon for follow-up PRs that will create the node builder itself.

### DIFF
--- a/.github/workflows/builder_node_slsa3.yml
+++ b/.github/workflows/builder_node_slsa3.yml
@@ -1,0 +1,390 @@
+# Copyright The GOSST team.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: SLSA node builder
+
+permissions:
+  contents: read
+
+env:
+  # Builder.
+  BUILDER_BINARY: slsa-builder-go-linux-amd64 # Name of the binary in the release assets.
+  BUILDER_DIR: internal/builders/node # Source directory if we compile the builder.
+
+###################################################################
+#                                                                 #
+#            Input and output argument definitions                #
+#                                                                 #
+###################################################################
+on:
+  workflow_call:
+    inputs:
+      # ======= General arguments ============
+      node-version:
+        description: >
+          The node version to use.
+
+          This argument is passed, unchanged, to `actions/node-setup`'s
+          as `node-version` parameter.
+
+        required: true
+        type: string
+      # TODO: support uploading assets to GitHub releases once we have confirmation
+      # that we can run `npm pack` and `npm --publish` provide the same result.
+      directory:
+        description: "The directory to change to before running commands."
+        required: false
+        type: string
+        default: ""
+      private-repository:
+        description: "If true, private repositories can post to the public transparency log."
+        required: false
+        type: boolean
+        default: false
+
+      # ======= Build arguments ============
+      ci-arguments:
+        # TODO: check whether we should filter out some arguments.
+        description: >
+          Optional list of arguments for `npm ci`, separated by a comma `,`.
+
+          Example: `--arg1, --arg2`.
+        required: false
+        type: string
+        default: ""
+
+      run-scripts:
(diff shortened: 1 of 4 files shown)


.github/workflows/builder_node_slsa3.yml
```diff
@@ -0,0 +1,388 @@
...
+    inputs:
+      node-version:
+        description: >
+          The node version to use.
+
+          This argument is passed, unchanged, to `actions/node-setup`'s
+          as `node-version` parameter.
+        required: true
+        type: string
+      # TODO: support uploading assets to GitHub release.
```
ianlewis, line 55: nit: Consider adding a github issue and adding the URL to the comment. Same for other TODOs.

ianlewis: We should also add a new milestone and add these issues to it.

.github/workflows/builder_node_slsa3.yml
```diff
@@ -0,0 +1,388 @@
...
+          # npm ci <args>
+          ./"$BUILDER_BINARY" ci \
+            --ci-arguments "$UNTRUSTED_CI_ARGUMENTS" \
+            --directory "$UNTRUSTED_DIR"
+
+      # TODO(hermeticity) OS-level.
+      # - name: Disable hermeticity
+      # uses: slsa/hermeticity@xxx
+      #   with:
+      #     to-state: enabled
```
ianlewis, line 240: nit: Maybe we can omit this comment?

.github/workflows/builder_node_slsa3.yml
```diff
@@ -0,0 +1,388 @@
...
+# limitations under the License.
+
+name: SLSA node builder
+
+permissions:
+  contents: read
+
+env:
+  # Builder.
+  BUILDER_BINARY: slsa-builder-go-linux-amd64 # Name of the binary in the release assets.
```
ianlewis, line 22: do you really want to use the Go builder when pulling a release?

.github/workflows/builder_node_slsa3.yml
```diff
@@ -0,0 +1,388 @@
...
+          path: "${{ steps.sign-prov.outputs.signed-provenance-name }}"
+          if-no-files-found: error
+          retention-days: 5
+
+  ###################################################################
+  #                                                                 #
+  #           Publish the package and its provenance                #
+  #                                                                 #
+  ###################################################################
+  publish:
```
ianlewis, line 344: I'm guessing we should probably have an option that allows you to skip the publishing step.

.github/workflows/builder_node_slsa3.yml
ianlewis: I added a [milestone](https://github.com/slsa-framework/slsa-github-generator/milestone/8) that we can add future issues to and can track publicly as part of the roadmap.

.github/workflows/builder_node_slsa3.yml
```diff
@@ -0,0 +1,388 @@
...
+
+      - name: Download dependencies
+        env: 
+          UNTRUSTED_CI_ARGUMENTS: "${{ inputs.ci-arguments }}"
+          UNTRUSTED_DIR: ${{ inputs.directory }}"
+        run: |
+          set -euo pipefail
+
+          # npm ci <args>
+          ./"$BUILDER_BINARY" ci \
```
ianlewis, line 232: I wonder if we can't combine the `ci` and `build` steps. especially if we will always be running `npm ci`

.github/workflows/builder_node_slsa3.yml
laurentsimon: yeah I was planning on creating an issue for it. Thanks for the reminder.

.github/workflows/builder_node_slsa3.yml
laurentsimon: oversight. Fixed.

.github/workflows/builder_node_slsa3.yml
laurentsimon: How about we don't provide it right away and see if people ask for it? By default, this option would be set to `true`, so we could add it later. Wdut?

.github/workflows/builder_node_slsa3.yml
laurentsimon: One reason (maybe not a strong one) is if we want to turn on some OS-level hermeticity after downloading the dependencies in the future.

.github/workflows/builder_node_slsa3.yml
laurentsimon: I shortened it a to a single line. Unless we think it would be done in the builder. Let me know.

laurentsimon:
friendly ping

ianlewis (approved):

PR #905, opened 2022-09-23
feat: add log when verify-checkout fails

laurentsimon:
Will also help debug a renovatebot PR that fails

### DIFF
--- a/.github/actions/verify-checkout/action.yaml
+++ b/.github/actions/verify-checkout/action.yaml
@@ -12,5 +12,7 @@ runs:
         git_sha=$(git log -1 --format='%H')
         if [[ "$git_sha" != "$GITHUB_SHA" ]]; then
             echo "mismatch git sha $git_sha != $GITHUB_SHA"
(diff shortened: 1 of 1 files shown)


.github/actions/verify-checkout/action.yaml
```diff
@@ -12,5 +12,7 @@ runs:
         git_sha=$(git log -1 --format='%H')
         if [[ "$git_sha" != "$GITHUB_SHA" ]]; then
             echo "mismatch git sha $git_sha != $GITHUB_SHA"
+            echo "logs:"
+            git log
```
ianlewis, line 16: Is this going to output all the commits? maybe something like `git log -20` to limit the length of the output?

ianlewis (approved):

.github/actions/verify-checkout/action.yaml
laurentsimon: ouch, let me send a PR to do that.

Issue #880, opened 2022-09-20
[feature] Improve code readability


laurentsimon:
We have several internal actions, pinned at different hash commits. This makes it hard for someone to read the code or perform audits.

Can we un-pin internal actions? Except for `detect-env`, I think it may be do-able.

laurentsimon:
Here is one idea. Instead of pinning by hash all actions, what we may do instead is `checkout` at a specific location:
```yaml
- uses: actions/checkout@xxx
  with:
    path: __builder_source__
```

I think this should simplify not jut code review, but our job. It will also simplify pre-submit, since right now they don't run on the same code that is submitted, if there are changes in the repo.

laurentsimon:
assigning to myself.

laurentsimon:
someone's build can overwrite it, but we assume they have code execution, so they can generate and change the hash already. They could have tampered with the env variable with the previous code, for example.

asraa:
> but we assume they have code execution,

OH OK: I think I understand now. Yes, in their build VM where they have code execution, the only checkout action done (to calculate sha) can be tampered with.

laurentsimon:
follow-up: there's another thing to be carful about: when we download artifacts from other jobs. We need to make sure it does not overwrite the builder code. The secure-download-artifact action does check for this already, and I will add an additional check to string match the builder folder (not really necessary but just in case).

ianlewis:
@laurentsimon We still pin almost all of the actions called from the reusable workflows. For example:
https://github.com/slsa-framework/slsa-github-generator/blob/ce2408f54c685a307925330e1d9ceda98c853fd8/.github/workflows/generator_generic_slsa3.yml#L71

Do we want to do a builder checkout for these others besides `secure-download-artifact` too? or is the idea to pin all of these at a release per #1084?

laurentsimon:
the references using the hash was a mistake from me: I sent https://github.com/slsa-framework/slsa-github-generator/pull/1126 to fix it. I also added pre-submit to enforce the references at main.

PR #941, opened 2022-09-29
fix: verify-checkout uses wrong sha to validate for pull_requests

laurentsimon:
This is the reason the pre-submits are currently failing for renovatebot.


ianlewis (approved):
Strange, how did this ever work?

laurentsimon:
It never did, because we never bumped the versions. I'm hoping to get rid of all the pinning in https://github.com/slsa-framework/slsa-github-generator/issues/880

PR #943, opened 2022-09-30
bugfix: pass Github token through to all container sub-workflows

flxw:
I realized that the container provenance workflow does not pass the token through to all the composite workflows and actions.
This causes problems with private repositories.
This PR should fix it.

It's my first contribution to this project, so please let me know if I should create an issue first, or do some more bookkeeping. Thanks for reviewing!

### DIFF
--- a/.github/workflows/generator_container_slsa3.yml
+++ b/.github/workflows/generator_container_slsa3.yml
@@ -48,6 +48,11 @@ on:
         required: false
         type: boolean
         default: false
+      token:
+        description: "GitHub token"
+        required: false
+        type: string
+        default: ${{ github.token }}
 
 jobs:
   privacy-check:
@@ -100,6 +105,7 @@ jobs:
           compile-builder: "${{ inputs.compile-generator }}"
           # NOTE: We are using the generic generator.
           directory: "${{ env.BUILDER_DIR }}"
+          token: "${{ inputs.token }}"
 
       - uses: sigstore/cosign-installer@f3c664df7af409cb4873aa5068053ba9d61a57b6 # tag=v2.6.0
       - name: Login
--- a/.github/actions/generate-builder/action.yml
+++ b/.github/actions/generate-builder/action.yml
@@ -39,6 +39,7 @@ runs:
         repository: "${{ inputs.repository }}"
         ref: "${{ inputs.ref }}"
         go-version: "${{ inputs.go-version }}"
+        token: "${{ inputs.token }}"
 
     - name: Generate builder
       shell: bash
--- a/internal/builders/container/README.md
+++ b/internal/builders/container/README.md
@@ -175,6 +175,7 @@ Inputs:
 | `digest`            | yes      | The OCI image digest. The image digest of the form '<algorithm>:<digest>' (e.g. 'sha256:abcdef...') |
 | `registry-username` | yes      | Username to log into the container registry.                                                        |
 | `compile-generator` | false    | Whether to build the generator from source. This increases build time by ~2m.                       |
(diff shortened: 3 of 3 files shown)


.github/workflows/generator_container_slsa3.yml
```diff
@@ -48,6 +48,11 @@ on:
         required: false
         type: boolean
         default: false
+      token:
```
laurentsimon, line 51: for secrets inputs to re-usable workflow, there is a separate secret field, see https://github.com/slsa-framework/slsa-github-generator/blob/cd7ae8d43f0389bd7ba47374dd8c93c1219cb589/.github/workflows/generator_container_slsa3.yml#L24

I think this parameter is not needed, in the sense that the default token is already available by default. (we use it to retrieve an OIDC token.)
I think all we need is the change to generate-builder/action.yml: add the `token` input and set the default as `default: ${{ github.token }}`.

laurentsimon (commented):
Thanks for the PR! I think we can further simplify it. I'll let @ianlewis take a look since he implemented this workflow.

In the description of the PR, could you provide a link to the logged error the PR fixes?

.github/workflows/generator_container_slsa3.yml
ianlewis: The `token` input to the `generate-builder` action already defaults to `github.token`. So it should work ok already.

https://github.com/slsa-framework/slsa-github-generator/blob/fc6500582079e2a1c41f181e7f04189b2a94816f/.github/actions/generate-builder/action.yml#L23-L26

I also don't think this might not be necessary. The reusable workflows declare what permissions are needed so the default `github.token` should always have the correct permissions. We also need to use the default token for OIDC to link the provenance to the workflow's identity. Using one token for OIDC and one for checkout and other actions might be a bit confusing to users.

ianlewis (commented):
Thanks for the PR! was there a specific use case you're trying to enable? I want to make sure we aren't missing anything.

flxw:
Hi all, thanks for the fast feedback on this!
At Celonis, we run the container generator workflow inside a private repository.
Without passing a token that is allowed to operate on the private repository, I have seen the _Generate builder_ step of the provenance job fail:

grafik

Does that help? Unfortunately, the logs don't say more, the whole thing just ends abruptly. It was tricky to debug.

ianlewis:
> Hi all, thanks for the fast feedback on this! At Celonis, we run the container generator workflow inside a private repository. Without passing a token that is allowed to operate on the private repository, I have seen the _Generate builder_ step of the provenance job fail:
> Does that help? Unfortunately, the logs don't say more, the whole thing just ends abruptly. It was tricky to debug.

Thanks. The container workflow hasn't been fully released so you might run into issues here and there. There was a similar permissions issue with the container workflow on #823.

What version of the workflow were you using. What does your job look like that calls the container workflow?

flxw:
> > Hi all, thanks for the fast feedback on this! At Celonis, we run the container generator workflow inside a private repository. Without passing a token that is allowed to operate on the private repository, I have seen the _Generate builder_ step of the provenance job fail:
> > Does that help? Unfortunately, the logs don't say more, the whole thing just ends abruptly. It was tricky to debug.
> 
> Thanks. The container workflow hasn't been fully released so you might run into issues here and there. There was a similar permissions issue with the container workflow on #823.
> 
> What version of the workflow were you using. What does your job look like that calls the container workflow?

Sure thing, I have noticed that it's not yet released - I see myself as an early adopter and want to help where I can.
I hope that I didn't come across as criticizing.

I used the version that was on the `main` branch then, and the workflow is taken from the [README](https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/container#getting-started).

Issue #942, opened 2022-09-30
[bug] ”Generate Builder“ step fails


jenstroeger:
**Describe the bug**

It looks like the [Generate Builder](https://github.com/slsa-framework/slsa-github-generator/blob/fc6500582079e2a1c41f181e7f04189b2a94816f/.github/workflows/generator_generic_slsa3.yml#L109-L118) step fails:
```
Run ./.github/actions/generate-builder/generate-builder.sh
  ./.github/actions/generate-builder/generate-builder.sh
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    BUILDER_BINARY: slsa-generator-generic-linux-amd64
    BUILDER_DIR: internal/builders/generic
    BUILDER_REPOSITORY: slsa-framework/slsa-github-generator
    BUILDER_RELEASE_BINARY: slsa-generator-generic-linux-amd64
    VERIFIER_REPOSITORY: slsa-framework/slsa-verifier
    VERIFIER_RELEASE_BINARY: slsa-verifier-linux-amd64
    VERIFIER_RELEASE_BINARY_SHA256: f92fc4e571949c796d7709bb3f0814a733124b0155e484fad095b5ca68b4cb21
    VERIFIER_RELEASE: v1.1.1
    COMPILE_BUILDER: false
    BUILDER_REF: refs/tags/v1.2.0
    GH_TOKEN: ***
Fetching the builder with ref: refs/tags/v1.2.0
Builder version: v1.2.0
BUILDER_REPOSITORY: slsa-framework/slsa-github-generator
verifier hash computed is f92fc4e571949c796d7709bb3f0814a733124b0155e484fad095b5ca68b4cb21
verifier hash verification has passed
FAILED: SLSA verification failed: could not find a matching valid signature entry
Error: Process completed with exit code 6.
```
when invoked as a job similar to [this public example](https://github.com/jenstroeger/python-package-template/blob/f41b0e6a46061081370253b9ac17a1b62b5c85dc/.github/workflows/release.yaml#L67-L77) but in a private repo:
```yaml
  provenance:
    needs: build
    # The generator should be referenced with a semantic version.
    # The build will fail if we reference it using the commit sha.
    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0
    with:
      base64-subjects: ${{ needs.build.outputs.artifacts-sha256 }}
    permissions:
[...]

PR #945, opened 2022-10-02
update verifier version in actions

asraa:

Pending https://github.com/slsa-framework/slsa-verifier/pull/288/files

to fix https://github.com/slsa-framework/slsa-github-generator/issues/942


asraa:
The hash was verified in https://github.com/slsa-framework/slsa-verifier/pull/288

asraa:
After this: a new patch version of the builders needs to be pushed

ianlewis (approved):

ianlewis (approved):

lukehinds:
when will a release be made with this fix?

asraa:
> when will a release be made with this fix?

Hi! We're just about done. We're working on some fixes in the release https://github.com/slsa-framework/slsa-github-generator/issues/968, and should definitely be done by EOW if not sooner.

Thank you so much for the patience here.

PR #946, opened 2022-10-03
Add doc on sigstore policy-controller

### LINKED ISSUE #491: [docs] Verification with sigstore-policy-controller
Document examples using [sigstore-policy-controller](https://docs.sigstore.dev/policy-controller/overview) to verify provenance.

ianlewis:
Fixes #491 

Adds doc on using sigstore [policy-controller](https://docs.sigstore.dev/policy-controller/overview) w/ Kubernetes to verify container provenance.


internal/builders/container/README.md
```diff
@@ -321,6 +322,84 @@ Certificate issuer URL:  https://token.actions.githubusercontent.com
...
+
+```yaml
+apiVersion: policy.sigstore.dev/v1alpha1
+kind: ClusterImagePolicy
+metadata:
+  name: image-is-signed-by-github-actions
+spec:
+  images:
+    # All images in example repository matched
+    - glob: "**"
```
ianlewis, line 344: Maybe update this to be more specific and match the kyverno policy below.

Maybe something like: `glob: "ghcr.io/ianlewis/actions-test:*"`

ianlewis:
@laurentsimon @asraa @joshuagl friendly ping

@hectorj2f @vaikas I wrote a short example for how to use sigstore policy-controller to verify images created by our container workflow. We would kindly appreciate feedback.

internal/builders/container/README.md
```diff
@@ -321,6 +322,86 @@ Certificate issuer URL:  https://token.actions.githubusercontent.com
...
+                    // repository.
+                    uri: =~"^git\\+https://github.com/ianlewis/actions-test@refs/tags/v[0-9]+.[0-9]+.[0-9]+$"
+                  }
+                }
+              }
+```
+
+When applied the `ClusterImagePolicy` will be evaluated when a `Pod` is
+created. If the `Pod` fulfills the policy conditions then the `Pod` can be
+created. If the `Pod` does not fulfill one or more of the policy conditions
```
hectorj2f, line 394: This behaviour can be configured whenever a Pod doesn't match to any policy to deny, allow or warn:

```
In addition to that, the policy controller offers a configurable behavior defining whether to allow, deny or warn whenever an image does not match a policy. This behavior can be configured using the config-policy-controller ConfigMap created under the release namespace, and by adding an entry with the property no-match-policy and its value warn|allow|deny. By default, any image that does not match a policy is rejected whenever no-match-policy is not configured in the ConfigMap.
```

You can find additional info here: https://github.com/sigstore/docs/blob/main/content/en/policy-controller/overview.md. Perhaps you could simply add the following https://github.com/sigstore/docs/blob/main/content/en/policy-controller/overview.md#admission-of-images here.

laurentsimon (approved):
LGTM, thanks!

hectorj2f:
@ianlewis Thanks for doing this 🤗 ! It looks good to me, I just let some comments in relation to the last paragraph.

internal/builders/container/README.md
ianlewis: > This behaviour can be configured whenever a Pod doesn't match to any policy to deny, allow or warn:
> 
> ```
> In addition to that, the policy controller offers a configurable behavior defining whether to allow, deny or warn whenever an image does not match a policy. This behavior can be configured using the config-policy-controller ConfigMap created under the release namespace, and by adding an entry with the property no-match-policy and its value warn|allow|deny. By default, any image that does not match a policy is rejected whenever no-match-policy is not configured in the ConfigMap.
> ```
> 
> You can find additional info here: https://github.com/sigstore/docs/blob/main/content/en/policy-controller/overview.md. Perhaps you could simply add the following https://github.com/sigstore/docs/blob/main/content/en/policy-controller/overview.md#admission-of-images here.

Thanks for the feedback! The content you linked to is the same as on the website right? Perhaps that's a better link?
https://docs.sigstore.dev/policy-controller/overview/#admission-of-images

hectorj2f: It is related to the latest available version. I'm gonna solve the problem on the website, it should get those changes in there.

hectorj2f: Yes, I'll use [https://docs.sigstore.dev/policy-controller](https://docs.sigstore.dev/policy-controller) for any related info about the policy-controller.

ianlewis: Cool thanks.

hectorj2f: Fixed now in https://docs.sigstore.dev/policy-controller/overview/#admission-of-images!

PR #957, opened 2022-10-04
Update READMEs to clarify that SLSA generators and builders must be referred by tag

diogoteles08:
Previously, at some points across the different READMEs there were explanations on why it's necessary to refer the SLSA generators and builders using tags in the format @vX.Y.Z, but some of them were incomplete. This change adds a subtopic to the base README that contains all the explanation, and this subtopic is linked on other parts of READMEs that needs the explanation.

This issue was previously discussed with @laurentsimon

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ This repository contains the code, examples and technical design for system desc
 
 - [Roadmap](#roadmap)
 - [Generation of provenance](#generation-of-provenance)
+  - [Refering SLSA builders and generators](#referring-slsa-builders-and-generators)
   - [Builders](#builders)
   - [Provenance-only generators](#provenance-only-generators)
 - [Verification of provenance](#verification-of-provenance)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -14,6 +14,7 @@ This repository contains the code, examples and technical design for system desc
 
 - [Roadmap](#roadmap)
 - [Generation of provenance](#generation-of-provenance)
+  - [Refering SLSA builders and generators](#referring-slsa-builders-and-generators)
```
asraa, line 17:
```suggestion
  -[Referencing SLSA builders and generators](#referencing-slsa-builders-and-generators)
```

internal/builders/generic/README.md
```diff
@@ -80,8 +80,8 @@ provenance:
     base64-subjects: "${{ needs.build.outputs.hashes }}"
 ```
 
-**Note**: Make sure that you reference the generator with a semantic version of the form `@vX.Y.Z`. The build will fail
-if you reference it via a shorter tag like `@vX.Y` or `@vX`.
+**Note**: Make sure that you reference the generator with a semantic version of the form `@vX.Y.Z`. 
+More information [here](/README.md#referring-slsa-builders-and-generators).
```
asraa, line 84: referencing-slsa-builders-and-generators here and below like above if you make that change!

asraa (commented):
Thanks for helping!

README.md
diogoteles08: Thanks for the suggestion! I applied it

README.md
```diff
@@ -42,9 +43,13 @@ timeline for completion.
...
 
-**Note**: At present the GitHub Actions provided in this repository as builders and generators **MUST** be referenced by
-a tag that correpsonds to a semantic version of the form `@vX.Y.Z`. The build will fail
-if you reference it via a shorter tag like `@vX.Y` or `@vX` or if you reference it by a tag of a different form (e.g., `@main`).
+### Referencing SLSA builders and generators
+At present, the GitHub Actions provided in this repository as builders and generators **MUST** be referenced 
+by tag in order for the `slsa-verifier` to be able to verify the ref of the trusted builder/generator's 
+reusable workflow. It also needs to be referred as `@vX.Y.Z`, because the build will fail if you reference it via a shorter tag like `@vX.Y` or `@vX`.
+
+This is contrary to the [best practice](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) which recommends referencing by digest, but intentional due to limits in GitHub Actions.
```
ianlewis, line 51:
```suggestion
This is contrary to the [GitHub best practice for third-party actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) which recommends referencing by digest, but intentional due to limits in GitHub Actions.
```

ianlewis (approved):

PR #970, opened 2022-10-07
Update to use Go 1.19

### LINKED ISSUE #782: [refactor] Update to use Go 1.19

### LINKED ISSUE #969: [bug] second argument to errors.As should not be *error
This message pops up when running unit tests.
```
github/oidc_test.go:237:10: second argument to errors.As should not be *error
```

ianlewis:
Fixes #782 
Fixes #969


ianlewis:
The version of golangci-lint we are using doesn't support 1.19. Need to update it as well.

ianlewis:
Hmm, it seems Docker Content Trust hasn't been updated for the Go 1.19.x images.

ianlewis:
> Hmm, it seems Docker Content Trust hasn't been updated for the Go 1.19.x images.

Nope, it seems it was the wrong digest.

internal/runner/runner.go
```diff
@@ -128,10 +118,38 @@ func (r *CommandRunner) runStep(ctx context.Context, step *CommandStep, dry bool
...
-	// env = cmd.Environ()
+	// is needed to capture the actual environment used. However, we will not
+	// include posix environment variables in the captured environment variables as
+	// they are environment specific and inhibit reproducibility.
+	// See: https://github.com/slsa-framework/slsa-github-generator/issues/822
+
+	// First set the default environment variables.
+	env := []string{
+		"PWD=" + pwd,
+		"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
```
asraa, line 130: Why include PATH here? Does it get used internally before being stripped from `finalEnv`?

ianlewis: Yeah, a PATH needs to be set at least in order to find commands to be run. Without a `PATH` lots of stuff will break.

Currently, if your command is something like `["npm", "run", "build"]` the runner uses the builder's `PATH` to look up the location of `npm` but any commands run by npm will use this `PATH`. Alternatively, I could just use the builder's `PATH` here.

internal/runner/runner.go
```diff
@@ -128,10 +118,38 @@ func (r *CommandRunner) runStep(ctx context.Context, step *CommandStep, dry bool
...
+	cmd.Env = env
+
+	// Strip common POSIX env vars from provenance.
+	posixVars := []string{"PATH", "PWD", "HOME", "USER", "TERM", "SHELL", "EDITOR"}
+	finalEnv := []string{}
+	// cmd.Environ will dedup and get final environment variables.
+	for _, s := range cmd.Environ() {
+		k, _, _ := strings.Cut(s, "=")
+		isPosix := false
+		for _, pVar := range posixVars {
```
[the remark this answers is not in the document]
ianlewis: Good call.

internal/runner/runner.go
asraa: Ahhh gotcha. Since it's an isolated environment, there's likely no trickery someone could do even if you did use the builders PATH, so I guess either are fine.

laurentsimon: `PATH` is probably better so we don't inadvertently use a PATH that breaks the command?

ianlewis: > `PATH` is probably better so we don't inadvertently use a PATH that breaks the command?

I don't think I was consistent with the backticks. Just to be clear, you're advocating for using the builder's `PATH` right?

Side Note: it is possible for callers to override these Env vars and the Go builder already does it because it copies over the builder's environment variables itself.
https://github.com/slsa-framework/slsa-github-generator/blob/10cb4de28b1e0f50c3af6c494a088a155c87162e/internal/builders/go/pkg/build.go#L267

But only the env vars the user sets in their config + the Go env vars are recorded in proveance
https://github.com/slsa-framework/slsa-github-generator/blob/10cb4de28b1e0f50c3af6c494a088a155c87162e/internal/builders/go/pkg/build.go#L241

I resisted doing it at first but I think maybe I'll just adopt this and merge the env vars from `CommandRunner`'s caller with the GitHub job/builder's env but we'll only record the former. That way POSIX env vars like `PATH`, `PWD` etc. will be handled by the builder's env but will be ignored in provenance for reproducibility.

laurentsimon: I'm fine with that.

ianlewis: On the other hand the builder/GitHub environment is going to have a bunch of env vars like "GITHUB_OUTPUT", "GITHUB_ACTIONS", "GITHUB_EVENT" etc. that we don't really care about from a reproducibility standpoint... Maybe "GITHUB_*" is specific enough for us to ignore all those env vars?

asraa: GITHUB_, RUNNER_, and CI maybe?
https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables

ianlewis: Yeah, Those we perhaps won't want to record in provenance.

I think some more sensitive variables we set as part of the workflow we don't really want to pass along at all. Like `GITHUB_TOKEN` or `GITHUB_CONTEXT` (the [github context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) includes the github token).

ianlewis: I updated the runner to copy over the builder's environment variables but leave out the following from the provenance.
1. Common set of POSIX env vars
2. `GITHUB_*`, `RUNNER_*`, and `CI` environment variables.

ianlewis: @asraa @laurentsimon 
Given, that just about everything that comes from the builder's environment is basically just noise, I went with just including the Env that is explicitly given to the CommandRunner in provenance. That way the `GO*` env vars are captured but things like POSIX, GITHUB_*, RUNNER_, CI, TERM etc. env vars are excluded.

Apologies for taking a long route to get here and for doing all this on the same PR as the Go update. LMK if I should split it out.

Please take another look.

ianlewis: You know what, I'm just going to split the changes into separate PRs since they don't have anything to do with each other anymore.

Issue #968, opened 2022-10-07
[bug] Blocker release v1.2.1: verify-checkout compares SHA digests from different repos


ianlewis:
The `verify-checkout` action seems to compare `$GITHUB_SHA` with the locally checked out version. However, `$GITHUB_SHA` refers to the digest of the commit that triggered the workflow, which might be from a different repo than the one we are checking out.

`verify-checkout` seems to have been created on the [premise](https://github.com/slsa-framework/slsa-github-generator/issues/626#issuecomment-1238324548) that one could update the tag between the time that it was pushed and when it was checked out by the workflow.

> I tested a different attack. Say, there's a new tag pushed by maintainers. The checkout Action seems to fetch and do `git checkout --progress --force refs/tags/<tag>` - as per the logs. So one idea is to force push a different tag before the checkout happens, to trick the builder into fetching different code for the same tag. I was not able to make it work. It seems that the tag is "locked", it the old tag is still accessible until the workflow finishes.

I'm not sure this is actually the case since `git checkout` will only ever checkout the local tag reference and won't pull a new tag reference (that could be changed).

/cc @laurentsimon

ianlewis:
#993 finishes changes made to add the `secure-checkout` action. I still need to update the pre-submits that check that we don't use checkout.

ianlewis:
Reopening this because `secure-checkout` is still broken.

ianlewis:
> There are 2 "types" of checkout:
> 
> * project checkout: these are subject to the attacks with force pushing a tag with a different hash than reported in the Git±Hub event. The GITHUB_SHA should match for most events, except for PRs where the PR sha should match.

To be clear, this is a situation where the tag is being updated between the time the event was generated and the time the checkout actually happens? Right now `secure-checkout` checks out `github.sha` by default so it doesn't use the complicated logic in `actions/checkout`.

I also noticed something else. Since you mentioned that `actions/checkout` pulls a tar file, the tar could actually be a checkout with files modified. In that case `git log -format '%H'` could show that the git tree is correct and report the correct sha but not that the checkout is dirty.

> * builder checkout: we can harden by verifying the ref from detect-env

[...]

PR #971, opened 2022-10-07
Add secure-checkout action

ianlewis:
Updates #968 
Updates #795 

Adds a `secure-checkout` action to replace the existing `verify-checkout` action.

`verify-checkout` checked if the checked out commit matched `GITHUB_SHA` which doesn't work if checking out the builder repo like the `generate-builder` action does which causes failures.

`secure-checkout` instead wraps the `actions/checkout` and forces the user to specify a git sha1 digest to check out and verifies that the checked out commit is the expected commit.


.github/workflows/scripts/pre-submit.actions/checkout.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# Verify that no internal Actions are using `actions/checkout`
+# See reasoning in ./github/actions/README.md
+
+set -euo pipefail
+
+# NOTE: All actions and workflows should not use actions/checkout. The should
```
asraa, line 8: nit: s/the/they

asraa (approved):
LGTM!

Assuming you'll move over checkouts in checkout-* in another PR and update the presubmit

ianlewis:
> LGTM!
> 
> Assuming you'll move over checkouts in checkout-* in another PR and update the presubmit

Yep. I need to get the action checked in before I can reference it so I'll send a follow up PR. Thanks for the review!

PR #993, opened 2022-10-11
Update ref for checkout-go

ianlewis:


asraa (approved):

ianlewis:
I didn't test enough so secure-checkout was broken. Need #994 to go in and then need to update checkout-go/node again :(

PR #994, opened 2022-10-11
Fix secure-checkout bugs

ianlewis:


asraa (approved):
:|

as a follow-up I'll add actionlint to our presubmits. but it still doesn't catch everything

PR #1005, opened 2022-10-13
Support ref in secure-checkout

ianlewis:
Updates #968

This PR changes the `sha` input for `secure-checkout` to `ref` and allows passing git a `ref` or `sha`. The action now performs validation by looking up the `ref` in the API and checking that the sha digest returned by the API and the sha checked out locally match. This may not necessarily satisfy the TLS interception issue brought up on #626 but at least makes it a little harder to fake a checkout.

We need to support `ref` in `secure-checkout` because we need to be able to checkout the builder using a tag reference. This is because we recommend users call our workflows using a release tag.

e.g.
```yaml
uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.0
```

This results in `detect-env` action returning something like `refs/tags/v1.2.0` as the detected reference.

### DIFF
--- a/.github/actions/secure-checkout/action.yaml
+++ b/.github/actions/secure-checkout/action.yaml
@@ -7,7 +7,7 @@ inputs:
     required: false
     default: ${{ github.repository }}
   ref:
-    description: "The ref or SHA to checkout."
+    description: "A fullly formed ref (refs/...) or SHA to checkout."
     required: false
     # NOTE: different from actions/checkout which defaults to GITHUB_REF, GITHUB_SHA, or default branch.
     default: ${{ github.sha }}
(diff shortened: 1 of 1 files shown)


.github/actions/secure-checkout/action.yaml
```diff
@@ -6,12 +6,10 @@ inputs:
     description: "The repository to check out."
     required: false
     default: ${{ github.repository }}
-  sha:
-    description: "The SHA to checkout."
+  ref:
+    description: "The ref or SHA to checkout."
```
asraa, line 10: nit: document that ref should be fully-formed in the form `refs/*, and that it can handle branches and tags.
currently it also handles pull requests. I'm not sure why that would be used in practice

asraa: Documenting for later: we'll keep in branches, technically I think we only need it for tags and otherwise we're using sha

asraa (approved):

PR #1048, opened 2022-10-14
fix: update refs for checkout-go

asraa:

Uses the updated commit https://github.com/slsa-framework/slsa-github-generator/commit/b447920901e060ac2459d2e87fee02652033593e for `checkout-go`


mihaimaruseac (approved):

asraa:
I think @mihaimaruseac has write but not maintain access, so the reviews don't count to the merge requirement

bobcallaway (approved):

PR #1050, opened 2022-10-14
update refs for generate-builder

asraa:

Last one! Update the reference for `generate-builder` action, from https://github.com/slsa-framework/slsa-github-generator/commit/325c27afeab313babb4ab9265fa123957da01ac2


asraa:
The rate limiting monster came after me! I'll refresh the failed jobs in 10-15

Issue #1004, opened 2022-10-13
[bug] Migrate uses of set-output


ianlewis:
We need to migrate the use of `set-output` commands and use environment files.

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

ianlewis:
We also need to upgrade some actions which use the `save-state` command. This message was emitted by `actions/checkout`.

```
Warning: The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

PR #1061, opened 2022-10-16
fix: use GITHUB_OUTPUT instead of deprecated set-output command

suzuki-shunsuke:
- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

`set-output` was deprecated.

AS IS

```
echo '::set-output name={name}::{value}'
```

TO BE

```
echo "{name}={value}" >> $GITHUB_OUTPUT
```

### DIFF
--- a/.github/actions/rng/action.yml
+++ b/.github/actions/rng/action.yml
@@ -28,4 +28,4 @@ runs:
         # -l: the number of bytes
         # -c: the number of bytes displayed per column
         value=$(xxd -p -l "$LENGTH" -c "$LENGTH" /dev/urandom)
-        echo "::set-output name=result::$value"
+        echo "result=$value" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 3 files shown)


ianlewis:
Updates #1004

suzuki-shunsuke:
I thought [github.com/sethvargo/go-githubactions](https://pkg.go.dev/github.com/sethvargo/go-githubactions) is useful because this library has already migrated to GITHUB_OUTPUT in [v1.1.0](https://github.com/sethvargo/go-githubactions/releases/tag/v1.1.0), but I found the proposal was rejected. https://github.com/slsa-framework/slsa-github-generator/issues/153

ianlewis:
Yeah, I think it's a useful library as well but I think we would only be using it for reading the context and setting outputs so it's not really worth adding another dependency.

suzuki-shunsuke:
Yeah, I agree.

ianlewis (approved):

PR #1066, opened 2022-10-17
fix: use GITHUB_OUTPUT instead of deprecated set-output command

suzuki-shunsuke:
Same as https://github.com/slsa-framework/slsa-github-generator/pull/1061

- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

https://github.com/slsa-framework/slsa-github-generator/issues/1004


ianlewis:
I merged #1061 but we should probably wait to merge this until [v1.2.1](https://github.com/slsa-framework/slsa-github-generator/releases/tag/v1.2.1) is released.

laurentsimon (approved):

ianlewis:
@suzuki-shunsuke I think we can merge this but there is a conflict in one of the files and I can't fix it myself. Can you fix it? then we can merge.

suzuki-shunsuke:
Sure. I'm fixing.
The action `.github/actions/secure-checkout` seems to be removed.

ianlewis:
> Sure. I'm fixing. The action `.github/actions/secure-checkout` seems to be removed in the main branch.

Yeah, you can probably just ignore the changes there.

PR #1072, opened 2022-10-18
Add pre-submit for secure-checkout

ianlewis:
Adds pre-submits for `secure-checkout` that runs the code included in the PR. 

These pre-submits need to be added as required to the protected branch after merge.


ianlewis:
I'll close this since #1075 removes it anyway.

PR #1074, opened 2022-10-18
Document renovate exception for tags over digest.

datosh:
As proposed in https://github.com/slsa-framework/slsa-verifier/issues/12#issuecomment-1280619107 it might be helpful for other users to have a handy renovate snippet to follow the README's guidance on using tags instead of digests.


laurentsimon (approved):
Thanks!

PR #1075, opened 2022-10-18
Remove ref for internal action calls

laurentsimon:
Remove the ref we use to call internal actions, Instead, checkout the builder repo and calls them locally.

closes https://github.com/slsa-framework/slsa-github-generator/issues/880


.github/actions/generate-builder/action.yml
```diff
@@ -33,11 +33,16 @@ outputs:
 runs:
   using: "composite"
   steps:
-    - name: Checkout the Go builder repository
-      uses: slsa-framework/slsa-github-generator/.github/actions/checkout-go@c2e7da4c53f2703d9af77f8d1483078c8fd3477e
+    - name: Checkout builder repository
```
laurentsimon, line 36: I will update `@main` to `@hash` once the PR is merged.

laurentsimon:
@ianlewis appreciate feedback and questions, since this seems to undo some of the secure-checkout code. I want to be sure I'm not mis-understanding something.

.github/actions/secure-project-checkout-go/action.yml
```diff
@@ -22,8 +22,8 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Checkout the repository with user SHA
-      uses: slsa-framework/slsa-github-generator/.github/actions/secure-checkout@ea457948e6a989e9818304dcb4f15fb4ce6765d6
+    - name: Checkout the repository with user ref
+      uses: ./__BUILDER_CHECKOUT_DIR__/.github/actions/secure-project-checkout
```
asraa, line 26: doesn't this need a secure builder dir checkout before use?
same below for node. or maybe I'm missing something

.github/actions/secure-project-checkout/action.yaml
```diff
@@ -0,0 +1,46 @@
+name: "secure-project-checkout"
+description: "Checkout a project and verify its commit sha"
+
+inputs:
+  # the token is not available to actions by defaulrts, so we need to
```
asraa, line 5: nit: defaults

.github/workflows/builder_go_slsa3.yml
```diff
@@ -232,7 +245,7 @@ jobs:
 
       - name: Upload generated binary
         id: upload
-        uses: slsa-framework/slsa-github-generator/.github/actions/secure-upload-artifact@e3220805577deb9d193f64e519abcb3b50851df5
+        uses: ./__BUILDER_CHECKOUT_DIR__/.github/actions/secure-upload-artifact
```
asraa, line 248: my earlier comment: after the buidl procedure, do we trust `./__BUILDER_CHECKOUT_DIR__`?
https://github.com/slsa-framework/slsa-github-generator/issues/880#issuecomment-1282771342

.github/actions/secure-project-checkout-go/action.yml
asraa: maybe it propogates since it's an action and run in the same environment? if so, maybe this action should validate that BUILDER_CHECKOUT_DIR is populated?

laurentsimon: you're correct, it assumes the builder repo is checkout'ed. The checkout is done in the main re-usable workflow by calling `uses: slsa-framework/slsa-github-generator/.github/actions/secure-builder-checkout`.

The actions run in the same context as the job, so it's present.

What do you mean by "maybe this action should validate that BUILDER_CHECKOUT_DIR is populated"? The action will fail if the directory is not present, so the validation is implicit. Do you have something else in mind?

Btw, BUILDER_CHECKOUT_DIR is not an env variable, because `uses:` does not accept them.

Let me know if this clarifies or still confusing.

asraa: > The action will fail if the directory is not present, so the validation is implicit.

Only for a clear error message: maybe can you add a comment to the action.yml saying it requires that the repository is populated? Just in case it hits an error when creating a new builder

.github/workflows/builder_go_slsa3.yml
asraa: Got it, resolved in the issue!

asraa (approved):
lgtm, after I understand now the user risk and interaction here.

.github/actions/secure-builder-checkout/action.yaml
```diff
@@ -0,0 +1,33 @@
+name: "secure-builder-checkout"
+description: "Checkout the builder repository"
+
+inputs:
+  repository:
```
ianlewis, line 5: I feel like secure-builder-checkout could call `detect-workflow` itself since the `repository` and `ref` here are always the results of `detect-workflow`.

That way we won't even need the `repository` and `ref` inputs.

.github/actions/secure-builder-checkout/action.yaml
```diff
@@ -0,0 +1,33 @@
...
+    description: "The repository to check out."
+    required: true
+  ref:
+    description: "A fully formed ref (refs/...) or SHA to checkout."
```
ianlewis, line 9: If we pass this directly to `actions/checkout` it doesn't need to be fully formed. It could just be a tag or branch name.

.github/actions/secure-project-checkout/action.yaml
ianlewis:
```suggestion
  # the token is not available to actions by default, so we need to
```

.github/actions/secure-project-checkout/action.yaml
```diff
@@ -0,0 +1,46 @@
...
+  # share it explicitely. The token is needed to checkout private repositories.
```
ianlewis, line 6:
```suggestion
  # share it explicitly. The token is needed to checkout private repositories.
```

.github/actions/secure-project-checkout/action.yaml
```diff
@@ -0,0 +1,46 @@
...
+      run: |
+        set -euo pipefail
+
+        git_sha="$(git log -1 --format='%H')"
+        github_sha="$GITHUB_SHA"
+
+        if [[ "$GITHUB_EVENT_NAME" == "pull_request" ]]; then
+          github_sha="$PULL_REQUEST_SHA"
+        fi
+
```
ianlewis, line 37: @laurentsimon Since you mentioned `actions/checkout` pulls a tar, I wondered if that tar could somehow be overwritten with a dirty checkout such that `git log -1 --format='%H'` returns the right SHA but the checkout itself it dirty and contains different code.

Maybe not necessary for this PR but we could maybe add a check for `git status` or `git diff` in the future.

ianlewis (approved):

ianlewis:
> @ianlewis appreciate feedback and questions, since this seems to undo some of the secure-checkout code. I want to be sure I'm not mis-understanding something.

I think this is fine. I wasn't totally confident `secure-checkout` was solving the issue properly anyway.

.github/workflows/builder_go_slsa3.yml
```diff
@@ -182,13 +188,20 @@ jobs:
     runs-on: ubuntu-latest
     needs: [privacy-check, builder, build-dry, rng]
     steps:
+      - name: Checkout builder repository
+        uses: slsa-framework/slsa-github-generator/.github/actions/secure-builder-checkout@main
```
ianlewis, line 192: This unfortunately won't pass pre-submits because `secure-builder-checkout` only exists in this PR and not at `main` currently. We can either temporarily remove the required check in order to merge, or we'll need to add the secure-builder-checkout to the repo first before using it in the top level reusable workflows.

ianlewis:
Related #795

.github/actions/secure-builder-checkout/action.yaml
laurentsimon: discussed offline: until now we could have done that. But this PR requires the top re-usable workflow to checkout the builder code, so we need the repository / ref information to be available.

.github/actions/secure-builder-checkout/action.yaml
laurentsimon: Thanks. I updated the description.

PR #1081, opened 2022-10-19
Update CodeQL workflow

### LINKED ISSUE #1047: [CI] Disable CodeQL for changes only on YAML
**Describe the bug**
CodeQL takes a long time to run, and has no meaningful output when changes are only done on YAML or MD. 

Using `paths-ignore` doesn't quite work, because this will result in a stalemate situation where the check is Required to run, but never does. It doesn't seem to be "skipped",

One solution is maybe to create a job as a pre-step that exits early when the diff context (somewhat dificult to grab) is only YAML or MD

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

ianlewis:
Fixes #1047 

I added "Analyze" as a required check on our protected branch rule. This should match both the no-op CodeQL workflow and the real CodeQL workflow.

/cc @asraa 

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -18,21 +18,21 @@ on:
       # If more paths are added here, then they must be added to the paths
       # of codeql-analysis-noop.yml.
       - "**/*.yml"
-      - '**/*.md'
+      - "**/*.md"
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [main]
     paths-ignore:
       - "**/*.yml"
-      - '**/*.md'
+      - "**/*.md"
   schedule:
     - cron: "21 0 * * 2"
 
 permissions: read-all
 
 jobs:
-  analyze:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -79,3 +79,19 @@ jobs:
...
+  # This is necessary because the matrix strategy generates new jobs with
+  # different names and throws off the required checks on protected
+  # branches which are set by job name.
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    needs: [analyze-matrix]
+    if: ${{ always() }}
+    env:
+      ANALYZE_RESULT: ${{ needs.analyze-matrix.result }}
```
laurentsimon, line 93: so this will be "success" only if all matrix jobs succeed, correct? If so LGTM.

ianlewis: correct

PR #1107, opened 2022-10-20
fix: workingDir

laurentsimon:


ianlewis (approved):
Do we even use this file anywhere?

asraa:
> Do we even use this file anywhere?

In the go presubmit!

asraa:
I thought about filing an issue to consolidate with example-package: and just gate on whether we're running in this repository for avoiding the sig verification

asraa (approved):

ianlewis:
> > Do we even use this file anywhere?
> 
> In the go presubmit!
Yeah! I found it. Searching for it on GitHub yielded zero results so I was confused initially.

Strange that this breakage got in in the first place :-/

PR #1113, opened 2022-10-21
fix: checkout uses the wrong repository

laurentsimon:
This will fix the current error that affects all e2e tests.
The PR also uses `@main` for all the actions we use. We will set these to the tag release before release. I have added pre-release workflows to verify the release tag and the reference of the Actions is the same' and amended the RELEASE.md accordingly.

### DIFF
--- a/.github/workflows/builder_go_slsa3.yml
+++ b/.github/workflows/builder_go_slsa3.yml
@@ -142,7 +142,7 @@ jobs:
       go-env: ${{ steps.build-dry.outputs.go-env }}
       go-working-dir: ${{ steps.build-dry.outputs.go-working-dir }}
     runs-on: ubuntu-latest
-    needs: [privacy-check, builder, rng]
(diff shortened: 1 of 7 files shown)


asraa (commented):
could you quickly explain the end-state? 
these actions would eventually be checked out at a resulting ref from detect-env?

laurentsimon:
> could you quickly explain the end-state? these actions would eventually be checked out at a resulting ref from detect-env?

correct. During development, we always reference them at `@main`. This ensure we're using the most recent code.

At release time, the "entry Actions" are pinned by version. Example: someone calls our generator at `@v1.2.3`. The entry Actions will be referenced with the same `@v1.2.3`. This ensure the code that is run is the code from a particular tag, not a combination of different hashes.

The pre-release verifies that.

Let me know if you have questions.

asraa (approved):

Issue #959, opened 2022-10-05
[feature] Improve handling of transient errors


ianlewis:
Rekor and Fulcio may have transient errors and we can perhaps handle those better by retrying requests.

ianlewis:
Per https://github.com/sigstore/rekor/issues/1115 if we update to Rekor 1.0rc or later then we can use the rekor client that performs retries (apparently 4 times by default).

ianlewis:
Updating to Rekor 1.0 should fix this once we do that.

ianlewis:
We haven't had any e2e test failures due to transient errors with Rekor for a while. Closing.

PR #1121, opened 2022-10-23
Update Rekor to v1.0.0

ianlewis:
Updates #959 
Related https://github.com/sigstore/rekor/issues/1115

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/in-toto/in-toto-golang v0.3.4-0.20220709202702-fa494aaa0add
 	github.com/secure-systems-lab/go-securesystemslib v0.4.0
 	github.com/sigstore/cosign v1.12.1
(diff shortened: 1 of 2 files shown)


laurentsimon (approved):

asraa:
QQ: Can you update this line https://github.com/slsa-framework/slsa-github-generator/blob/dfea9e3a4de9e3aa7babf15f3553a73994a797b8/signing/sigstore/rekor.go#L63

to pull `client.GetRekorClient` instead? Otherwise cosign hasn't necessarily updated with the retryable client. And we might as well get rid of the cosign dep and use it directly.
https://github.com/sigstore/rekor/blob/7807d1e5a3fbc9b8fdd17053da8163463e87a5ae/pkg/client/rekor_client.go#L28

ianlewis:
Oops, didn't realize auto-merge was enabled. Will submit a follow up PR.

PR #1126, opened 2022-10-24
feat: improve refs

laurentsimon:
This PR does:
- update references and binary paths that were not referenced properly
- update the action doc
- update the release doc
- add pre-submit to verify references use `@main`


.github/workflows/generator_generic_slsa3.yml
```diff
@@ -108,7 +108,7 @@ jobs:
       actions: read # Needed to read workflow info.
     steps:
       - name: Generate builder
-        uses: slsa-framework/slsa-github-generator/.github/actions/generate-builder@923a5a249d2ea820614a80486c9ef0857ced5321
+        uses: slsa-framework/slsa-github-generator/.github/actions/generate-builder@main
```
ianlewis, line 111: Not sure if it's worth it but I think we could technically unpin `generate-builder` if we bring the `secure-builder-checkout` out into the reusable worflows.

In that case the rule would be that all actions are unpinned except `detect-env` and `secure-builder-checkout`.

.github/workflows/pre-submit.actions.yml
```diff
@@ -117,3 +123,117 @@ jobs:
...
+    if: ${{ always() }}
+    env:
+      BUILDER_NAME_RESULT: ${{ needs.secure-download-artifact-builder-name.result }}
+      BUILDER_FOLDER_RESULT: ${{ needs.secure-download-artifact-builder-repo-folder.result }}
+      BUILDER_FILE_RESULT: ${{ needs.secure-download-artifact-builder-repo-file.result }}
+      DOWNLOAD_RESULT: ${{ needs.secure-download.result }}
+    steps:
+      - run: |
+          # exit 0 if checks were successful.
+          [ "${BUILDER_NAME_RESULT}" != "success" ] && \
```
ianlewis, line 236: nit: instead of using '&&' you could just have one of these checks on each line since they will continue to execute as long as the exit codes are zero.

.github/workflows/pre-submit.actions.yml
```diff
@@ -117,3 +123,117 @@ jobs:
...
+    needs: [secure-download-artifact, secure-download-artifact-builder-repo-name, secure-download-artifact-builder-repo-folder, secure-download-artifact-builder-repo-file]
+    if: ${{ always() }}
+    env:
+      BUILDER_NAME_RESULT: ${{ needs.secure-download-artifact-builder-name.result }}
+      BUILDER_FOLDER_RESULT: ${{ needs.secure-download-artifact-builder-repo-folder.result }}
+      BUILDER_FILE_RESULT: ${{ needs.secure-download-artifact-builder-repo-file.result }}
+      DOWNLOAD_RESULT: ${{ needs.secure-download.result }}
+    steps:
+      - run: |
+          # exit 0 if checks were successful.
```
ianlewis, line 235: I think `-e` is on by by default but add `set -euo pipefail` just to make it explicit.

.github/workflows/scripts/pre-submit.actions/references.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Verify that Action calls that are pinned (not locally called)
+# use the same referemnce as the relealse tag.
+# See reasoning in ./github/actions/README.md
+
+set -euo pipefail
+
+results=$(
+    find .github/workflows/ -name '*.yaml' -o -name '*.yml' \
```
ianlewis, line 10: nit: We'll probably block on warnings at some point so ideally fix the shellcheck warning.

RELEASE.md
```diff
@@ -240,6 +246,14 @@ Update the documentation to recommend using the new version:
 $ find . -name "*.md" -exec sed -i "s/v1.0.0/v1.1.1/g" {} +
 ```
 
+## Send a PR to reference Actions at main
+
+Send a PR to reference the internal Actions at `@main`. You can use:
+
+```shell
+find .github/ -name '*.yaml' -o -name '*.yml' | xargs sed -i 's/uses: slsa-framework\/slsa-github-generator\/\.github\/actions\/\(.*\)@_YOUR_RELEASE_TAG_*/uses: slsa-framework\/slsa-github-generator\/.github\/actions\/\1@main/'
```
ianlewis, line 254: Is this sed replacement not backwards? It looks like it replaces `uses: slsa-framework/slsa-github-generator/.github/actions/action_name@vX.Y.Z` with `uses: slsa-framework/slsa-github-generator/.github/actions/action_name@main` but shouldn't it be the other way around for release PRs?

laurentsimon:
> @ianlewis any ideas why some pre-submit are not running?

there was a yaml error.

ianlewis:
> > @ianlewis any ideas why some pre-submit are not running?
> 
> there was a yaml error.

Yeah, if there is a yaml error they don't show up in the list unfortunately. It's very unintuitive.

.github/workflows/generator_generic_slsa3.yml
```diff
@@ -68,7 +68,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Check private repos
-        uses: slsa-framework/slsa-github-generator/.github/actions/privacy-check@a3c7a56c8749c2c423f01bbcfd063315efc07a22
+        uses: slsa-framework/slsa-github-generator/.github/actions/privacy-check@main
```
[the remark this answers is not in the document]
laurentsimon: good point. let me do that.

RELEASE.md
ianlewis: Ah, I see. I missed the change above. This one is to set it back to main.

.github/workflows/scripts/pre-submit.actions/references.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# Verify that Action calls that are pinned (not locally called)
+# use the same referemnce as the relealse tag.
+# See reasoning in ./github/actions/README.md
+
+set -euo pipefail
+
+results=$(
```
ianlewis, line 9: nit: maybe we should check that it's either a semver tag if the PR has 'label#release' or `main` if not. You can maybe copy over the semver utilities from example-package.

.github/workflows/generator_generic_slsa3.yml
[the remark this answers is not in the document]
laurentsimon: done. I could not add it inside `secure-builder-checkout` because the `path` is dynamic and I cannot use a dynamic value in the `uses:`. I added it in `generate-builder` instead. Wdut?

Note that the changes sort-of assume other jobs in the workflow depend on the `build` job

.github/workflows/generator_generic_slsa3.yml
laurentsimon: we could. But then we pass the same arguments to both `secure-builder-checkout` and `generate-builder` which seems a little weird. wdut?

.github/workflows/pre-submit.actions.yml
laurentsimon: done. I added an different exit code to simply troubleshooting.

.github/workflows/pre-submit.actions.yml
laurentsimon: +1, done.

.github/workflows/scripts/pre-submit.actions/references.sh
laurentsimon: I tried to but the shell run was complaining about invalid command :/

RELEASE.md
laurentsimon: After the release, we revert to using `@main` for development until the next release. The reasoning is that during development, we want to use the most recent code to catch problems before the next release.

.github/workflows/scripts/pre-submit.actions/references.sh
laurentsimon: done. PTAL

laurentsimon:
Note: some adversarial tests (3 of them) will always show as failing after this PR.

.github/workflows/scripts/pre-submit.actions/references.sh
ianlewis: I think you should be able to use `find ... -print0 | xargs -0 ...`. Did that not work for you? Maybe it's a flag not available on MacOS or something?

laurentsimon: mistake from my end. Done.

RELEASE.md
```diff
@@ -33,7 +33,13 @@ $ export GH=/path/to/gh
 
 ## Pre-release tests
 
-Verify the references to the internal Actions by manually running the [release workflow](https://github.com/slsa-framework/slsa-github-generator/actions/workflows/release.yml). Ensure this workflow succeeds.
+Verify the references to the internal Actions by manually running the [release workflow](https://github.com/slsa-framework/slsa-github-generator/actions/workflows/release.yml). Ensure this workflow succeeds. If the references are still `@main`,update them with the following command:
```
asraa, line 36:
```suggestion
Verify the references to the internal Actions by manually running the [release workflow](https://github.com/slsa-framework/slsa-github-generator/actions/workflows/release.yml). Ensure this workflow succeeds. If the references are still `@main`, update them with the following command:
```

.github/actions/README.md
```diff
@@ -35,7 +35,7 @@ These Actions are *composite actions*. They invoke scripts and also call other A
 
 ## Development
 
-To create or update an internal Action, follow the 2 following steps:
+To create or update an internal Action, reference them at `@main`. For a releasem reference them at the release tag.
```
asraa, line 38:
```suggestion
To create or update an internal Action, reference them at `@main`. For a release, reference them at the release tag.
```

asraa (approved):

PR #1160, opened 2022-10-26
Fix privacy-check checkout

ianlewis:
We need to temporarily need to disable the reusable workflow pre-submits because they use `generate-builder@main` and it's broken at `main`. Will follow up with a later PR to re-enable them.

Also added pre-submits so we hopefully don't break `generate-builder` at `main` again.

Also updates adversarial pre-submits so that the check is done as part of each job. This is so that checks don't show up as failed in PRs.

### DIFF
--- a/.github/actions/generate-builder/action.yml
+++ b/.github/actions/generate-builder/action.yml
@@ -46,7 +46,7 @@ runs:
         path: __BUILDER_CHECKOUT_DIR__
 
     - name: Check private repos
-      uses: ./__BUILDER_CHECKOUT_DIR__/.github/actions/privacy-check@main
+      uses: ./__BUILDER_CHECKOUT_DIR__/.github/actions/privacy-check
       with:
         error_message: "Repository is private. The workflow has halted in order to keep the repository name from being exposed in the public transparency log. Set 'private-repository' to override."
         override: ${{ inputs.allow-private-repository }}
--- a/.github/workflows/pre-submit.actions.yml
+++ b/.github/workflows/pre-submit.actions.yml
@@ -193,10 +193,10 @@ jobs:
       - name: fail check
         if: always()
         env:
-          RESULT: ${{ steps.download-artifact.result }}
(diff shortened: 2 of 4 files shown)


laurentsimon (approved):
Thanks!

.github/actions/generate-builder/action.yml
```diff
@@ -46,7 +46,7 @@ runs:
         path: __BUILDER_CHECKOUT_DIR__
 
     - name: Check private repos
-      uses: ./__BUILDER_CHECKOUT_DIR__/.github/actions/privacy-check@main
+      uses: ./__BUILDER_CHECKOUT_DIR__/.github/actions/privacy-check
```
laurentsimon, line 49: good catch.

.github/workflows/pre-submit.actions.yml
```diff
@@ -182,12 +182,21 @@ jobs:
...
 
       - name: Download artifact
+        id: download-artifact
         uses: ./__BUILDER_CHECKOUT_DIR__/.github/actions/secure-download-artifact
         with:
           name: artifact2
           path: path/to/__BUILDER_CHECKOUT_DIR__/artifact2
           sha256: 5b3513f580c8397212ff2c8f459c199efc0c90e4354a5f3533adf0a3fff3a530
 
+      - name: fail check
```
laurentsimon, line 192: I tested this and it still showed up as failed pre-submit in the GitHub UI. I probably made a mistake during my tests.

ianlewis: Yeah, I'm playing with it a bit now.

laurentsimon: actually it still shows in this PR's pre-submit.

ianlewis: I think it should work if I use `continue-on-error` and then check the step `outcome`.

laurentsimon: cool, do you confirm it works?

PR #1161, opened 2022-10-26
re-enable pre-submits

ianlewis:
Re-enable reusable workflow pre-submits after disabling them in #1160


laurentsimon (approved):
Thanks!

PR #1170, opened 2022-10-27
Add known issues to docs

ianlewis:
Add a section on known issues to workflow docs.


internal/builders/go/README.md
```diff
@@ -91,19 +92,19 @@ In the meantime, you can use both GoReleaser and this builder in the same reposi
...
-    - darwin
-  goarch:
-    - amd64
-    - arm64
-    - s390x
-  # This instructs GoReleaser to not build for linux amd64.
-  ignore:
-    - goos: linux
-      goarch: amd64
+---
```
ianlewis, line 95: need to fix this. seems like an issue with the markdown formatter I use.

asraa (commented):
Thank you! 

Should we add a note about `v1.2.0`s issue or is it clear that a patch version supersedes an issue?

This one was the issue with an incompatible change when Rekor sharded, "After Rekor sharded (rotated trees) in production, the tree head no longer corresponds to the same tree that the entry was in, which caused a verification error."

https://github.com/slsa-framework/slsa-github-generator/issues/942

ianlewis:
> Thank you!
> 
> Should we add a note about `v1.2.0`s issue or is it clear that a patch version supersedes an issue?
> 
> This one was the issue with an incompatible change when Rekor sharded, "After Rekor sharded (rotated trees) in production, the tree head no longer corresponds to the same tree that the entry was in, which caused a verification error."
> 
> #942

I made the note say '1.2.x' in the sub header. AFAICT you currently get the same error for both v1.2.1 and v1.2.0

ianlewis:
Maybe I can make it clearer what the affected versions are.

PR #1176, opened 2022-10-28
Container build type

### LINKED ISSUE #638: [discussion] buildType for container generator
The container workflow currently just uses the generic generator binary. This means that the SLSA `buildType` for provenance generated by the container workflow is `"https://github.com/slsa-framework/slsa-github-generator/generic@v1"` which is the same as the generic workflow.

I'm thinking that the container workflow should have it's own unique `buildType` but not sure it's totally necessary since the dependent fields like `buildConfig` etc. are all the same.

ianlewis:
Fixes #638

Updates provenance generated by the generic container worflow to have it's own `buildType`.

Before:

```json
{
  "predicate": {
    "buildType": "https://github.com/slsa-framework/slsa-github-generator/generic@v1"
  }
}
```

After:

```json
{
  "predicate": {
    "buildType": "https://github.com/slsa-framework/slsa-github-generator/container@v1"
  }
}
```

This adds a new builder binary. Currently it has mostly the same logic as the generic builder's `generate` command had but may require more specialized logic in the future.


internal/builders/generic/generate.go
```diff
@@ -15,14 +15,10 @@
 package main
```
ianlewis, line 15: Probably we don't even need the `generate` command on the generic builder anymore? I think it was only used by tho container workflow anyway.

internal/builders/common/buildtype.go
```diff
@@ -0,0 +1,14 @@
...
+
+// GenericBuild is a very generic build type where build type can be specified.
+type GenericBuild struct {
+	*slsa.GithubActionsBuild
+	BuildTypeURI string
+}
+
+// URI implements BuildType.URI.
+func (b *GenericBuild) URI() string {
+	return b.BuildTypeURI
```
ianlewis, line 13: example provenance docs should be updated to reflect the new build type URI.

internal/builders/container/README.md
```diff
@@ -246,7 +246,7 @@ generated as an [in-toto](https://in-toto.io/) statement with a SLSA predicate.
     "builder": {
       "id": "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/generator_container_slsa3.yml@refs/tags/v1.1.1"
     },
-    "buildType": "https://github.com/slsa-framework/slsa-github-generator/generic@v1",
+    "buildType": "https://github.com/slsa-framework/slsa-github-generator/container@v1",
```
laurentsimon, line 249: the container builder and this one will have the same type, correct?
Do we want to align it with the Dockerfile-based builder TDD? I think the doc used `https://github.com/slsa-framework/slsa-github-generator/docker-based-builder@v1`, but the one you have LGT: as well.

@asraa wdut?

laurentsimon (approved):

internal/builders/container/README.md
ianlewis: I think they can have separate buildTypes for now. We could maybe revisit how these should work and maybe combine them when we support SLSA v1.0.

ianlewis: I'm giving the generic container generator the name 'container' here. The convention we have been using so far is to use the same name as the directory name for the builder in `internal/builders`.

ianlewis: I'm curious to get @asraa's thoughts too though.

asraa: > I think they can have separate buildTypes for now. We could maybe revisit how these should work and maybe combine them when we support [SLSA v1.0](https://github.com/MarkLodato/slsa/blob/provenance-refactor/docs/provenance/v1.0.proto).

It seems like the 1.0 spec will make it so that buildType will be agnostic of the environment that it was run in. I'm fine to have it be the same convention as the builder, but part of me thinks that in 1.0 these *should* be the same: the type of "build" performed here was a "no-op" observation, and mostly fills the details on the attestor/github environment, and not much about the process the artifact was built.

ianlewis: Agreed, I think that's likely the direction we'll want to go. I'm mostly doing this to cover bases before a GA of the generic container workflow because we'll likely want to do that before SLSA v1.0 gets finalized.

asraa: That's good with me!

asraa: Mostly because no one really is relying on this buildType field, and I think it'll be fine to move over -- people know SLSA isn't 1.0 stable yet

PR #1178, opened 2022-10-28
Update README.md

wietse-gmail:
Update the generic provenance generator's code example to use generator_generic_slsa3 v1.2.1. The code example uses v1.2.0, but the text describes the inputs 'provenance-name' and 'private-repository' which exist only in v1.2.1. 

Instead, we could update the text, and add "Available in version 1.2.1" to the definitions of  'provenance-name' and 'private-repository'.

Signed-off-by: Wietse Z Venema <72045954+wietse-gmail@users.noreply.github.com>

### DIFF
--- a/internal/builders/generic/README.md
+++ b/internal/builders/generic/README.md
@@ -134,7 +134,7 @@ jobs:
       actions: read
       id-token: write
       contents: write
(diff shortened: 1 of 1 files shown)


ianlewis:
@wietse-gmail Thanks for doing this. I think I updated this on a separate PR and that's why this was closed?

I like the idea of noting when inputs became available. Could you maybe submit a PR for that?

PR #1185, opened 2022-11-01
Group updates for renovate

ianlewis:
This should hopefully group up some updates so that they are easier to merge.

The intention of the update is as follows:
- Go updates that are minor and patch version updates should be grouped into a single PR. Major updates would not be grouped.
- npm dependencies that are minor and patch version update should be grouped into a single PR. Major updates would not be grouped.
- npm devDependencies should be grouped into a single PR.
- GitHub actions dependencies should be grouped.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,23 +1,27 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": [
-    "config:base"
-  ],
+  "extends": ["config:base"],
   "schedule": ["every weekend"],
   "packageRules": [
     {
       "matchManagers": ["github-actions"],
-      "groupName": "github-actions",
-      "commitMessageTopic": "github-actions"
+      "groupName": "github-actions"
     },
     {
-      "extends": "group:nodeJs",
-      "matchUpdateTypes": [
-        "minor",
-        "patch"
-      ],
-      "groupName": "nodeJs",
-      "commitMessageTopic": "NodeJs"
+      "matchManagers": ["gomod"],
+      "matchUpdateTypes": ["minor", "patch"],
+      "groupName": "go"
+    },
+    {
+      "matchManagers": ["npm"],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -1,23 +1,27 @@
...
-      ],
-      "groupName": "nodeJs",
-      "commitMessageTopic": "NodeJs"
+      "matchManagers": ["gomod"],
+      "matchUpdateTypes": ["minor", "patch"],
+      "groupName": "go"
+    },
+    {
+      "matchManagers": ["npm"],
+      "matchDepTypes": ["dependencies"],
```
ianlewis, line 17: NOTE: I'm not 100% sure about this depType since there don't seem to be any docs on available depTypes in renovate. I'm making a guess at this given other examples in the docs and npm terminology.

laurentsimon: @rarkins could you advise? Thanks!

renovate.json
```diff
@@ -1,23 +1,27 @@
 {
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": [
-    "config:base"
-  ],
+  "extends": ["config:base"],
   "schedule": ["every weekend"],
```
laurentsimon, line 4: Can we move this to monthly, except for cosign and its dependencies?

laurentsimon (approved):

renovate.json
rarkins: `depTypes` varies per manager. For some managers it can be dynamic from the user too. I the case of `npm`, `dependencies` and `devDependencies` are static terms and correct, as you've done

laurentsimon: Thank you!

renovate.json
ianlewis: I think this will lighten the burden significantly already. We can move to monthly updates if it's still too much.

PR #1189, opened 2022-11-01
feat: add commands to nodejs builder

laurentsimon:


internal/builders/nodejs/ci.go
```diff
@@ -0,0 +1,42 @@
...
+
+// ciCmd runs the 'ci' command.
+func ciCmd(check func(error)) *cobra.Command {
+	var ciArguments []string
+	var ciDirectory string
+
+	c := &cobra.Command{
+		Use:   "ci",
+		Short: "Run ci command",
+		Long:  `Run ci command to install dependencies of a node project.`,
```
asraa, line 31: nit: add an example usage with the arguments

internal/builders/nodejs/main.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func rootCmd() *cobra.Command {
+	c := &cobra.Command{
+		Use:   "slsa-github-generator",
+		Short: "Generate SLSA provenance for Github Actions",
+		Long: `Generate SLSA provenance for Github Actions.
+For more information on SLSA, visit https://slsa.dev`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return errors.New("expected command")
```
asraa, line 43: maybe print Usage string provided by cobra?

asraa (commented):
I'm guessing this is just template work right now? If so LGTM, maybe just add some example usage or start the README.md with the expected commands.

laurentsimon:
> I'm guessing this is just template work right now? If so LGTM, maybe just add some example usage or start the README.md with the expected commands.

correct, just the template for the command we need. I'll add README info.

internal/builders/nodejs/main.go
laurentsimon: it does it automatically, it seems

ianlewis (approved):

PR #1193, opened 2022-11-02
🌱 Ensure documentation is up-to-date with latest release

### LINKED ISSUE #1188: [feature] Builder ref in documentation should match the latest stable release
We have struggled to keep the ref of the builder in the documentation in sync with the latest stable builder we'd like people to use.

Let's:
1. Add it to RELEASE.md process
2. Add an automated pre-release test to verify the refs are set properly. We do already have some pre-release tests but not for docs (https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/release.yml#L19)

pnacht:
Closes #1188 

- Adds pre-release test to ensure all documentation points to the new release tag
- If not, lists all cases in the format:
  `filename:line-number: [...]/action-name.yml@_wrong-tag_`
  i.e.
  `/internal/builders/go/README.md:205: [...]/builder_go_slsa3.yml@v1.2.0`
  This removes the clutter that comes from the entire `uses: ...` path, as well as adding the line numbers of all infractions
- Also adopts this format for the internal Actions test
- Change to .github/actions/README.md to conform with other placeholders (would otherwise fail the documentation check)
- Modifies RELEASE.md, moving documentation-update up to the "pre-release" stage, to align with new pre-release test

As-is, the documentation check only detects tags within a "workflow chunk" (that is, containing a `uses:` term).

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -6,6 +6,7 @@ This is a document to describe the release process for the Go builder. Since all
 
 - [Pre-release](#pre-release-tests)
 - [Tagging](#tagging)
+- [Verify version references & code freeze](#verify-version-references--code-freeze)
 - [Post-release tests](#post-release-tests)
 - [Update Verifier](#update-verifier)
 - [Finalize release](#finalize-release)
@@ -31,19 +32,36 @@ $ export BUILDER_REPOSITORY="$GITHUB_USERNAME/slsa-github-generator"
 $ export GH=/path/to/gh
 ```
 
-## Pre-release tests
+## Tagging
 
-Verify the references to the internal Actions by manually running the [release workflow](https://github.com/slsa-framework/slsa-github-generator/actions/workflows/release.yml). Ensure this workflow succeeds. If the references are still `@main`, update them with the following command:
+Create a new tag for the official generator via [slsa-framework/slsa-github-generator/releases/new](https://github.com/slsa-framework/slsa-github-generator/releases/new).
+The tag _MUST_ be a "canonical" semantic version without metadata (`$BUILDER_TAG`). Shorter versions are not accepted by the builder's and verifier's code.
+
+Set the title to `$BUILDER_TAG`.
+
+Tick the `This is a pre-release` option.
+
+Click `Publish release`.
+
+## Verify version references & code freeze
+
+Update version references with the following command:
 
 ```shell
-find .github/workflows/ -name '*.yaml' -o -name '*.yml' | xargs sed -i 's/uses: slsa-framework\/slsa-github-generator\/\.github\/actions\/\(.*\)@main*/uses: slsa-framework\/slsa-github-generator\/.github\/actions\/\1@_YOUR_RELEASE_TAG_/'
+find .github/workflows/ -name '*.yaml' -o -name '*.yml' | xargs sed -i "s/uses: slsa-framework\/slsa-github-generator\/\.github\/actions\/\(.*\)@\(main\|v[0-9]\+\.[0-9]\+\.[0-9]\+\)/uses: slsa-framework\/slsa-github-generator\/.github\/actions\/\1@$BUILDER_TAG/"
 ```
 
 Send a PR with this update and add `#label:release ${BUILDER_TAG}` in the PR description.
 
+Once the PR is merged, update the tag to point to HEAD.
+
 Code freeze the repository for 1-2 days.
 
 Verify all the e2e tests in [github.com/slsa-framework/example-package/.github/workflows/](github.com/slsa-framework/example-package/.github/workflows/) are passing. (They run daily).
+
+## Pre-release tests
+
+Verify the references to the internal Actions by manually running the [release workflow](https://github.com/slsa-framework/slsa-github-generator/actions/workflows/release.yml). Ensure this workflow succeeds.
 
 There is one integration test we cannot easily test "live", so we need to simulate it by changing the code: malicious verifier binary in assets. We want to be sure the builder fails if the verifier's binary is tampered with. For this:
 
@@ -115,23 +133,6 @@ There is one integration test we cannot easily test "live", so we need to simula
    verifier hash computed is 5891b5b522d5df086d0ff0b110fbd9d21bb4fc7163af34d08286a2e846f6be03
    Error: Process completed with exit code 4.
    ```
-
-## Tagging
-
-Create a new tag for the official generator via [slsa-framework/slsa-github-generator/releases/new](https://github.com/slsa-framework/slsa-github-generator/releases/new).
-The tag _MUST_ be a "canonical" semantic version without metadata (`$BUILDER_TAG`). Shorter versions are not accepted by the builder's and verifier's code.
-
-Set the title to `$BUILDER_TAG`.
-
-Tick the `This is a pre-release` option.
-
-Click `Publish release`.
-
-Download the generated binary `slsa-builder-go-linux-amd64` locally on your machine:
-
-```
-$ "$GH" release -R slsa-framework/slsa-github-generator download "$BUILDER_TAG" -p "slsa-builder-go-linux-amd64"
-```
 
 ## Post-release tests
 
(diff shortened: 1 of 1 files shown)


.github/workflows/scripts/pre-release/references.sh
```diff
@@ -32,3 +33,16 @@ if [[ "$results" != "" ]]; then
     exit 1
 fi
 
+# Verify documentation refers to the most recent release tag
+results=$(
+    find . -name "*.md" -print0 \
+    | xargs -0 grep -Pn "uses: slsa-framework/.*@(?!<|main|$RELEASE_TAG)" \
```
asraa, line 39: could we qualify w/ the repo name slsa-github-generator too? just in case we start using slsa-verifier action

asraa, line 39: do we currently allow @main for something? i know the internal actions reference main, but I would expect those only to be used in the .yml's not in the documentation

asraa (commented):
Thanks @pnacht !

.github/workflows/scripts/pre-release/references.sh
pnacht: Good catch. Done.

pnacht: `@main` is currently used in two documentation pages.

One is for the reuseable `detect-workflow` workflow. 

https://github.com/slsa-framework/slsa-github-generator/blob/9a62ed287f4ffc9ab9c45aa40264c92f10784f51/.github/actions/detect-workflow/README.md?plain=1#L27

The other is for the generic container builder, which doesn't have a tagged release yet.

https://github.com/slsa-framework/slsa-github-generator/blob/9a62ed287f4ffc9ab9c45aa40264c92f10784f51/internal/builders/container/README.md?plain=1#L70

https://github.com/slsa-framework/slsa-github-generator/blob/9a62ed287f4ffc9ab9c45aa40264c92f10784f51/internal/builders/container/README.md?plain=1#L143

I can shrink the search path to ignore the `.github` folder (and therefore the `detect-workflow` "infraction") if you wish.

But not much to be done about the container builder, I don't think. The search will need to handle it somehow, for now: either by allowing main or by omitting that directory as well.

I'd argue temporarily allowing `@main` is better simply because the threat (keeping an Action `@main` after a release exists) is far more obvious and therefore easier to detect, at which point the Action can be patched. Omitting the directory risks no one remembering it's omitted and therefore it remains with old tags for longer than it should.

I've added a comment stating `@main` should be forbidden once the container builder is released.

pnacht:
Also improved the error if the documentation is obsolete. I'd previously just modified the error from the other check, but documentation also allows `<pseudo_tags>`.

.github/workflows/scripts/pre-release/references.sh
pnacht: I also feel really sorry for user `@main`, who must get a thousand notifications a day. 😂

.github/workflows/scripts/pre-release/references.sh
```diff
@@ -32,3 +33,19 @@ if [[ "$results" != "" ]]; then
     exit 1
 fi
 
+# Verify documentation refers to the most recent release tag
+# TODO(https://github.com/slsa-framework/slsa-github-generator/milestone/3):
```
ianlewis, line 37: Maybe we should add something like `-not -path "./internal/builders/container/*"` to the find command instead of allowing `main`?

Also, let's create & link to an issue rather than the milestone.

.github/workflows/scripts/pre-release/references.sh
ianlewis: > `@main` is currently used in two documentation pages.
> 
> One is for the reuseable `detect-workflow` workflow.
> 
> https://github.com/slsa-framework/slsa-github-generator/blob/9a62ed287f4ffc9ab9c45aa40264c92f10784f51/.github/actions/detect-workflow/README.md?plain=1#L27

We should make this a more generic reusable workflow example like `uses: path/to/reusable_workflow.yml@v1.0`

> The other is for the generic container builder, which doesn't have a tagged release yet.

See my other comment.

laurentsimon:
> filename:line-number

This will be brittle when we add code: line number will change. Can we instead use some sort of annotation? This [https://stackoverflow.com/questions/4823468/comments-in-markdown] shows that:
```
<empty line>
[whatever comment text]::
```

is not rendered. So we could use something like:
```
<empty line>
[pre-submit]::no-latest-tag
```

(Note: I don't think this works inside code snippet)

RELEASE.md
```diff
@@ -33,12 +33,20 @@ $ export GH=/path/to/gh
...
+If workflows still refer to `@main`, update them with the following command:
 
 ```shell
 find .github/workflows/ -name '*.yaml' -o -name '*.yml' | xargs sed -i 's/uses: slsa-framework\/slsa-github-generator\/\.github\/actions\/\(.*\)@main*/uses: slsa-framework\/slsa-github-generator\/.github\/actions\/\1@_YOUR_RELEASE_TAG_/'
 ```
 
+Likewise, if documentation refers to a previous release, update it with the following command:
+
+```shell
+$ find . -name "*.md" -exec sed -i "s/_PREVIOUS_RELEASE_TAG_/_YOUR_RELEASE_TAG_/g" {} +
```
laurentsimon, line 47: Maybe (not sure) we could add a PREVIOUS_RELEASE_TAG as env variables, so that here we could say:
```shell
$ find . -name "*.md" -exec sed -i "s/$PREVIOUS_RELEASE_TAG/$RELEASE_TAG/g" {} +
```

pnacht: Not sure I understand. This is run locally as a throwaway command, no? We wouldn't even have access to the Action's `$RELEASE_TAG` variable.

.github/workflows/scripts/pre-release/references.sh
pnacht: On `detect-workflow`, to clarify, are you suggesting creating another repo for this Action (and maybe others)? Would you suggest making this change before merging this PR?

.github/workflows/scripts/pre-release/references.sh
pnacht: On the `-not -path`: this wouldn't remove the alert on the `detect-workflow` workflow, though that might be solved by what you suggested in the other comment.

On the issue: would #409 be alright? I assume it'll only be considered closed when the container builder is actually released.

pnacht:
> > filename:line-number
> 
> This will be brittle when we update README text, because the line number will change. Can we instead use some sort of annotation?

I'm not sure I understand. That format is only printed to the workflow logs in case of errors. The line numbers are only shown to help find the outdated lines. What are you thinking of annotating?

RELEASE.md
laurentsimon: we run these commands on our local machines. When doing the release process, we set up a series of env variables (https://github.com/slsa-framework/slsa-github-generator/blob/3c5599ba46a13926b6daf4f67202a28c4d6183d3/RELEASE.md#prerequisites). I was suggesting we add another one `PREVIOUS_RELEASE_TAG` to avoid mistakes.

pnacht: Ah yes, of course. 🤦‍♂️

But we'd then need both RELEASE_TAG and PREVIOUS_RELEASE_TAG, no?

laurentsimon: yes. REQLSE_TAG is already set, it's called BUILDER_TAG. So maybe call the previous one BUILDER_PREVIOUS_TAG?

.github/workflows/scripts/pre-release/references.sh
```diff
@@ -32,3 +33,19 @@ if [[ "$results" != "" ]]; then
...
+results=$(
+    find . -name "*.md" -print0 \
+    | xargs -0 grep -Pn "uses: slsa-framework/slsa-github-generator/.*@(?!<|main|$RELEASE_TAG)" \
+    | sed 's/\(.*:\) *uses:.*\(\/.*\)/\1 [...]\2/' \
+    || true
+)
+
+if [[ "$results" != "" ]]; then
+    echo "Some documentation refers to an incorrect release tag"
+    echo "Allowed tags are <pseudo_tags> or \"$RELEASE_TAG\""
```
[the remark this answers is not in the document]
pnacht: Done.

RELEASE.md
pnacht: Done. See if my implementation is ok.

pnacht: If of interest, I could also simplify the `sed` commands by using another separator, i.e.:

```shell
sed -i "s/uses: slsa-framework\/slsa-github-generator\/\.github\/actions\/\(.*\)@main*/uses: slsa-framework\/slsa-github-generator\/.github\/actions\/\1@$BUILDER_TAG/"

# becomes

sed -i "s#uses: slsa-framework/slsa-github-generator/\.github/actions/\(.*\)@main*#uses: slsa-framework/slsa-github-generator/.github/actions/\1@$BUILDER_TAG#"
```

.github/workflows/scripts/pre-release/references.sh
ianlewis: I don't think we need another repo. I'm just suggesting that we change the example reusable workflow used in the [.github/actions/detect-workflow/README.md](https://github.com/slsa-framework/slsa-github-generator/blob/9a62ed287f4ffc9ab9c45aa40264c92f10784f51/.github/actions/detect-workflow/README.md?plain=1#L27) from

```yaml
uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@main 
```

to

```yaml
uses: path/to/reusable_workflow.yml@v1.0
```

so that it doesn't need to be changed during releases. It's a generic action so the example in the doc doesn't need to use `slsa-framework/slsa-github-generator/.github/...`

.github/workflows/scripts/pre-release/references.sh
ianlewis: > On the `-not -path`: this wouldn't remove the alert on the `detect-workflow` workflow, though that might be solved by what you suggested in the other comment.

IIUC, It should only exclude the `./internal/builders/container/` directory which contains the generic container builder's code and docs. I don't think `detect-workflow` would be affected.

> On the issue: would #409 be alright? I assume it'll only be considered closed when the container builder is actually released.

SG

RELEASE.md
```diff
@@ -33,10 +35,18 @@ $ export GH=/path/to/gh
...
+
+```shell
+find .github/workflows/ -name "*.yaml" -o -name "*.yml" | xargs sed -i "s/uses: slsa-framework\/slsa-github-generator\/\.github\/actions\/\(.*\)@main*/uses: slsa-framework\/slsa-github-generator\/.github\/actions\/\1@$BUILDER_TAG/"
+```
+
+Likewise, if documentation refers to a previous release, update it with the following command:
 
 ```shell
-find .github/workflows/ -name '*.yaml' -o -name '*.yml' | xargs sed -i 's/uses: slsa-framework\/slsa-github-generator\/\.github\/actions\/\(.*\)@main*/uses: slsa-framework\/slsa-github-generator\/.github\/actions\/\1@_YOUR_RELEASE_TAG_/'
+$ find . -name "*.md" -exec sed -i "s/$BUILDER_PREVIOUS_TAG/$BUILDER_TAG/g" {} +
```
ianlewis, line 49: You may want to rebase with main as I merged #1212 which mostly obviates the need for these changes by adding a more flexible find command.

pnacht: Done.

PR #1194, opened 2022-11-03
cleanup: remove more set-outputs

asraa:

Fixes https://github.com/slsa-framework/slsa-github-generator/issues/1004


.github/workflows/builder_go_slsa3.yml
```diff
@@ -230,7 +230,6 @@ jobs:
         run: |
           set -euo pipefail
 
-          # Disable set-output command.
```
asraa, line 233: stop-commands is NOT deprecated

laurentsimon (approved):

.github/workflows/builder_node_slsa3.yml
```diff
@@ -268,7 +268,7 @@ jobs:
 
           # cp output into upper folder to make the tarball accessible to
           # next step.
-          echo '::set-output name=filename::$TARBALL'
+          echo 'filename=$TARBALL >> $GITHUB_OUTPUT'
```
ianlewis, line 271: I don't think this actually works (and was broken before this too). You need to double quote this or the environment variables are not expanded.
```suggestion
          echo "filename=$TARBALL >> $GITHUB_OUTPUT"
```

github/set_output.go
```diff
@@ -0,0 +1,35 @@
...
+		f, err := os.OpenFile(filename, os.O_APPEND|os.O_WRONLY, 0o755)
+		if err != nil {
+			return err
+		}
+		if _, err := f.WriteString(name + "=" + value + "\n"); err != nil {
+			return err
+		}
+		return f.Close()
+	}
+	return errors.New("file GITHUB_OUTPUT not found")
```
ianlewis, line 34: nit: I think general consensus for readability nowadays is to exit early for errors like this rather than at the end.
```suggestion
	filename := os.Getenv("GITHUB_OUTPUT")
	if filename == "" {
		return errors.New("file GITHUB_OUTPUT not found")
	}
	f, err := os.OpenFile(filename, os.O_APPEND|os.O_WRONLY, 0o755)
	if err != nil {
		return err
	}
	if _, err := f.WriteString(name + "=" + value + "\n"); err != nil {
		return err
	}
	return f.Close()
```

ianlewis: nit: Alternatively you could fall back to using `set-output` if `GITHUB_OUTPUT` is not defined as they do in the Github Actions toolkit
https://github.com/actions/toolkit/blob/b36e70495fbee083eb20f600eafa9091d832577d/packages/core/src/core.ts#L192-L200

github/set_output.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	"errors"
+	"os"
+)
+
+// SetOutput writes a name value pair to a file located at GITHUB_OUTPUT.
+func SetOutput(name, value string) error {
+	if filename := os.Getenv("GITHUB_OUTPUT"); filename != "" {
+		f, err := os.OpenFile(filename, os.O_APPEND|os.O_WRONLY, 0o755)
```
ianlewis, line 25: nit: It's hard to say what the appropriate  #mode should be for the newly created file. The Github Actions toolkit uses the `fs.appendFileSync` default of `0o666` but I doubt they thought about it much.
https://github.com/actions/toolkit/blob/b36e70495fbee083eb20f600eafa9091d832577d/packages/core/src/file-command.ts#L22
https://nodejs.org/api/fs.html#fsappendfilesyncpath-data-options

```suggestion
		f, err := os.OpenFile(filename, os.O_APPEND|os.O_WRONLY, 0o666)
```

ianlewis (approved):

.github/actions/detect-workflow/main.go
```diff
@@ -157,6 +157,6 @@ func main() {
 	fmt.Printf("ref:%s\n", ref)
 
 	// Output of the Action.
-	fmt.Println(fmt.Sprintf(`::set-output name=repository::%s`, repository))
-	fmt.Println(fmt.Sprintf(`::set-output name=ref::%s`, ref))
+	github.SetOutput("repository", repository)
+	github.SetOutput("ref", ref)
```
ianlewis, line 161: Looks like you'll need to add this on a later PR after the code gets merged since `detect-env` uses a separate `go.mod`.

github/set_output.go
asraa: I'll add a TODO to remove this when it's EOL for set-output

.github/actions/detect-workflow/main.go
asraa: OHHH I see.

PR #1206, opened 2022-11-05
feat: npm builder updates

laurentsimon:
This PR:
1. Simplifies the builder binary
2. Tries to make job outputs more generic to simplify building new builders (possibly with BYOB framework)


.github/workflows/builder_nodejs_slsa3.yml
```diff
@@ -220,62 +218,64 @@ jobs:
...
-            --directory "$UNTRUSTED_DIR"
+          "$GITHUB_WORKSPACE/$BUILDER_BINARY" build
 
-      # TODO(hermeticity) Enable OS-level hermeticity.
+      - name: Upload generated tarball
+        id: upload
+        uses: ./__BUILDER_CHECKOUT_DIR__/.github/actions/secure-upload-artifact
+        with:
+          name: "<TODO>"
+          path: "<TODO>"
```
ianlewis, line 236: Were you planning on setting these?

ianlewis (approved):

.github/workflows/builder_nodejs_slsa3.yml
laurentsimon: not in this PR. The output of the builder can theoretically contain multiple artifacts, so I think we need to write code to process a number of artifacts at once. For nodejs, I think it will always be a single artifact. But in general I'm trying to se what's needed to build a "generic workflow" that generalizes for any builder (nodejs, Java, goreleaser, etc)

PR #1210, opened 2022-11-07
Update release tag for v1.2.2

ianlewis:
#label:release v1.2.2

Signed-off-by: Ian Lewis <ianlewis@google.com>

PR #1216, opened 2022-11-07
fix release tag parsing from body

asraa:
This is a test!

### DIFF
--- a/.github/workflows/pre-submit.actions.yml
+++ b/.github/workflows/pre-submit.actions.yml
@@ -110,7 +110,7 @@ jobs:
       - name: Verify version references
         if: ${{ contains(github.event.pull_request.body, '#label:release') }}
         run: |
-          echo "${{ github.event.pull_request.body }}" | awk '{print $2}'
+          echo "${{ github.event.pull_request.body }}" | grep -o -e 'v[0-9].[0-9].[0-9]'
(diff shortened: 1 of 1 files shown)


asraa:
Just used a grep match for a version string: https://github.com/slsa-framework/slsa-github-generator/actions/runs/3413607704/jobs/5680520567

Worked here

.github/workflows/pre-submit.actions.yml
```diff
@@ -110,7 +110,7 @@ jobs:
       - name: Verify version references
         if: ${{ contains(github.event.pull_request.body, '#label:release') }}
         run: |
-          echo ${{ github.event.pull_request.body }} | awk '{print $2}'
+          RELEASE_TAG=$(echo "${{ github.event.pull_request.body }}" | grep -o -e 'v[0-9].[0-9].[0-9]')
```
ianlewis, line 113: '.' will get interpreted as regex so it probably should be escaped. otherwise it matches strings like 'v1a2b3'.
```suggestion
          RELEASE_TAG=$(echo "${{ github.event.pull_request.body }}" | grep -o -e 'v[0-9]\.[0-9]\.[0-9]')
```

ianlewis (approved):

.github/workflows/pre-submit.actions.yml
asraa: Noted, I think it got copied into https://github.com/slsa-framework/slsa-github-generator/pull/1210/files, let's merge yours which does a code freeze, but I'll fix that next release

PR #1217, opened 2022-11-07
wip: docker-based template

asraa:

@rbehjati this is a quick docker-based template outlining some inputs and outputs and the steps of the reuseable workflow.

There are some TODOs. Let me know if you have some feedback about the options for the steps. I'd like to set up some pre-submits to start, and then start creating the builder binary.


.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -0,0 +1,191 @@
...
+        required: true 
+        type: string
+      builder-digest:
+        description: >
+          The OCI image digest of the builder-image. 
+          This must be specified in order to fully specify the builder 
+          image. The image digest is of the form '<alg>:<digest>'
+        required: true 
+        type: string
+      config:
```
laurentsimon, line 46: nit: config-path?

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -0,0 +1,191 @@
...
+        description: >
+          Path to a configuration file containing a command that the builder
+          image should be invoked with and optionally, docker options
+          and environment variables.
+        required: true 
+        type: string
+      output:
```
laurentsimon, line 53: Let's use a common output across all builders. We need to decide the format, and what we include: artifacts, provenance files, etc including where to download them from (thru download-artifact Action).

We can align this format with the BYOB framework as well: once we have it, all our internal builders can also use the framework.

Not needed for this PR.

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -0,0 +1,191 @@
...
+        description: >
+          Path to the output artifact(s) relative to the root of the Git
+          repository.
+          Users may specify a single artifact or use a shell-style wildcard
+          pattern to match multiple artifacts.
+        required: true 
+        type: string
+      source-repository:
```
laurentsimon, line 61: do you think the source-repository and source-sha should be a single option, repo/name@sha1?
Just curious.

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -0,0 +1,191 @@
...
+        description: >
+          Source repository to checkout for the build. For example, octocat/hello-world.
+          The builder will mount the root of this source repository to the workspace.
+
+          By default, uses the source repository of the caller workflow.
+        required: false 
+        type: string
+      source-digest:
```
laurentsimon, line 69: how do we support different hashes? GitHub only support sha1 today, but will support sha256 in the future. Can we acommodate it? Maybe `source-sha1` or `source-commit-sha1`?

Or keep source-digest and use a format like `sha1:bla`

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -0,0 +1,191 @@
...
+  #           Optional: Verify builder-image provenance             #
+  #                                                                 #
+  ###################################################################
+  verify-builder:
+    # This is a no-op right now.
+    outputs:
+      verified: ${{ steps.verify.outputs.verified }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: run no-op slsa-verifier
```
laurentsimon, line 96: Will the container download require authentication? I think it might, we we may need secret in workflow arguments.

Think about how we can express this in the BYOB Option 1. At the moment we only support binaries to download. We can either add an option for container images in the BYOB config file, or let the "callback Action" pull the image. (Probably a better option if secrets are necessary)

laurentsimon (approved):

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -0,0 +1,191 @@
...
+
+###################################################################
+#                                                                 #
+#            Input and output argument definitions                #
+#                                                                 #
+###################################################################
+on:
+  workflow_call:
+    inputs:
+      builder-image:
```
laurentsimon, line 33: you also need a `private-repository` input.

.github/workflows/builder_docker-based_slsa3.yml
asraa: This is the build output produced by the container image execution, i.e. the subject that needs to be measured. (Not the uploaded artifacts). Maybe I should rename this to builder-output?

laurentsimon: what do you mean by "measure"? Is it the list of artifacts to include in the attestation? If so, shall we remove this option to start with (image run should be able to make a cp of artifacts it wants attestation for in a predefined directory).

asraa: Yeah exactly -- we can start with pre-declaring where the output should be, e.g. an output folder like `output`.

laurentsimon: SG.

.github/workflows/builder_docker-based_slsa3.yml
ianlewis: I think Asra copied what I did in the container generator workflow for container images with the idea that it was clearer for 
folks to understand they *need* to use a digest rather than a tag, branch, etc. I don't have really strong opinions so I could go either way. It is maybe more natural for users the full `repo@sha`.

ianlewis: In the case of the container generator at least it makes sense to split them up because actions like `docker/build-push-action` often return the image `digest` they built as a separate output.

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -0,0 +1,191 @@
...
+          The SHA of the source repository to checkout. 
+
+          By default, uses the source repository of the caller workflow.
+        required: false
+        type: string
+      # TODO: Options for requesting and verifying provenance of builder image
+      # This will use slsa-verifier and slsa-verifier option.
+      #      * request-builder-verification
+
+      # TODO: Authentication options for the builder-image.
```
rbehjati, line 80: Ideally the image should be publicly available, no?

rbehjati (commented):
Thanks for putting this together.

.github/workflows/builder_docker-based_slsa3.yml
laurentsimon: not if it's used by a private repository. We've had several users reach out to us who used private repos.

rbehjati: Yes. I agree. I hope it is possible to specify this kind of dependency between input option on GitHub workflows.

.github/workflows/builder_docker-based_slsa3.yml
asraa: Yeah, the builder image and digest I'll definitely keep separate. Right now the buidlder binary in https://github.com/slsa-framework/slsa-github-generator/pull/1267 ingests source repo and sha separately. I could update this later though if we change both

.github/workflows/builder_docker-based_slsa3.yml
asraa: I like the format of sha1:blah

.github/workflows/builder_docker-based_slsa3.yml
asraa: The container download will happen when performing the build, so the `build` step may require authentication. I'll leave a TODO there. For now, Oak's is public, but when we test we can add in secret workflow arguments (also labeled TODO)

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -0,0 +1,191 @@
...
+      builder-digest:
+        description: >
+          The OCI image digest of the builder-image. 
+          This must be specified in order to fully specify the builder 
+          image. The image digest is of the form '<alg>:<digest>'
+        required: true 
+        type: string
+      config:
+        description: >
+          Path to a configuration file containing a command that the builder
```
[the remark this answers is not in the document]
asraa: Yes! Updated

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -0,0 +1,191 @@
...
+        description: >
+          Source repository to checkout for the build. For example, octocat/hello-world.
+          The builder will mount the root of this source repository to the workspace.
+
+          By default, uses the source repository of the caller workflow.
+        required: false 
+        type: string
+      source-digest:
+        description: >
+          The SHA of the source repository to checkout.
```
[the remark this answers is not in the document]
asraa: I allowed specifying alg:digest in case that changes, so keeping as is

.github/workflows/builder_docker-based_slsa3.yml
asraa: I added TODOs here to specify secrets and also a TODO step in the build to authenticate first if needed. For now we can start with public options as we test

asraa:
I'll merge here and do an iteration on this as well.

PR #1221, opened 2022-11-09
Fix builder-fetch.sh path

ianlewis:

### DIFF
--- a/.github/actions/generate-builder/action.yml
+++ b/.github/actions/generate-builder/action.yml
@@ -72,11 +72,10 @@ runs:
 
         COMPILE_BUILDER: "${{ inputs.compile-builder }}"
         BUILDER_REF: "${{ inputs.ref }}"
-        BUILDER_DIR: "${{ inputs.directory }}"
+        BUILDER_DIR: "./__BUILDER_CHECKOUT_DIR__/${{ inputs.directory }}"
         # Needed for the gh CLI used in builder-fetch.sh.
         GH_TOKEN: "${{ inputs.token }}"
-      working-directory: ./__BUILDER_CHECKOUT_DIR__
-      run: ./.github/actions/generate-builder/generate-builder.sh
+      run: ./__BUILDER_CHECKOUT_DIR__/.github/actions/generate-builder/generate-builder.sh
 
     - name: Compute sha256 of builder
       uses: ./__BUILDER_CHECKOUT_DIR__/.github/actions/compute-sha256
--- a/.github/actions/generate-builder/action.yml
+++ b/.github/actions/generate-builder/action.yml
@@ -72,10 +72,11 @@ runs:
 
         COMPILE_BUILDER: "${{ inputs.compile-builder }}"
         BUILDER_REF: "${{ inputs.ref }}"
-        BUILDER_DIR: "./__BUILDER_CHECKOUT_DIR__/${{ inputs.directory }}"
+        BUILDER_DIR: "${{ inputs.directory }}"
         # Needed for the gh CLI used in builder-fetch.sh.
         GH_TOKEN: "${{ inputs.token }}"
-      run: ./__BUILDER_CHECKOUT_DIR__/.github/actions/generate-builder/generate-builder.sh
+      working-directory: ./__BUILDER_CHECKOUT_DIR__
+      run: ./.github/actions/generate-builder/generate-builder.sh
 
     - name: Compute sha256 of builder
       uses: ./__BUILDER_CHECKOUT_DIR__/.github/actions/compute-sha256
--- a/.github/actions/generate-builder/generate-builder.sh
+++ b/.github/actions/generate-builder/generate-builder.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-if [[ "$COMPILE_BUILDER" = true ]]; then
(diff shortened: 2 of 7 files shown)


asraa (commented):
somewhat naive question but could this script have been invoked after changing to the ./__BUILDER_CHECKOUT__DIR__/ here https://github.com/slsa-framework/slsa-github-generator/blob/55d5ea5f73a1bd4b82b9faeb7e614d4fc6bd222d/.github/actions/generate-builder/action.yml#L78?

(that way the script doesn't need to know the outer context)

ianlewis:
> somewhat naive question but could this script have been invoked after changing to the ./**BUILDER_CHECKOUT__DIR**/ here
> 
> https://github.com/slsa-framework/slsa-github-generator/blob/55d5ea5f73a1bd4b82b9faeb7e614d4fc6bd222d/.github/actions/generate-builder/action.yml#L78
> 
> ?
> (that way the script doesn't need to know the outer context)

Yeah, I thought about that. We could probably set the working dir.

asraa:
> Yeah, I thought about that. We could probably set the working dir.

I think it would be somewhat clearer, but feel free to push back: we know this one works

asraa (approved):

ianlewis:
> > Yeah, I thought about that. We could probably set the working dir.
> 
> I think it would be somewhat clearer, but feel free to push back: we know this one works

I agree. I would like to keep the script context free if possible.

laurentsimon:
> > somewhat naive question but could this script have been invoked after changing to the ./**BUILDER_CHECKOUT__DIR**/ here
> > https://github.com/slsa-framework/slsa-github-generator/blob/55d5ea5f73a1bd4b82b9faeb7e614d4fc6bd222d/.github/actions/generate-builder/action.yml#L78
> > 
> > ?
> > (that way the script doesn't need to know the outer context)
> 
> Yeah, I thought about that. We could probably set the working dir.

what do you mean by outer context? I think we need to agree on the way *all* actions invoke scripts for consistency. Otherwise we'll start running into subtle problems: the folders to access the source code (__PROJECT_CHECKOUT_DIR__) won't be accessible the same way, etc. Keeping the invocation from the default director simplifies all these complications

asraa:
> what do you mean by outer context?

The outer context being the action YAML that's invoking it. 

OK thinking about it from that angle, I'd actually assume that scripts in internal actions should be running relative to that action, not the root of the slsa-github-generator repository

laurentsimon:
> > > Yeah, I thought about that. We could probably set the working dir.
> > 
> > 
> > I think it would be somewhat clearer, but feel free to push back: we know this one works
> 
> I agree. I would like to keep the script context free if possible.

replied too late, but totally agre

> > what do you mean by outer context?
> 
> The outer context being the action YAML that's invoking it.
> 
> OK thinking about it from that angle, I'd actually assume that scripts in internal actions should be running relative to that action, not the root of the slsa-github-generator repository

I don't think we can assume that. Sometimes the internal Actions need to access the __PROJECT_CHECKOUT_DIR__. Having the working directory the top-level root means we can keep the access the same and can easily identify what's invoking: something in the project repo or the builder repo. If we remove this, I'm afraid we're going to start wondering what the working directory is, and make mistakes overwriting things we should not. We want to be sure we don't overwrite builder's code, for example.

ianlewis:
> > what do you mean by outer context?
> 
> The outer context being the action YAML that's invoking it.
> 
> OK thinking about it from that angle, I'd actually assume that scripts in internal actions should be running relative to that action, not the root of the slsa-github-generator repository



> > > what do you mean by outer context?
> > 
> > 
> > The outer context being the action YAML that's invoking it.
> > OK thinking about it from that angle, I'd actually assume that scripts in internal actions should be running relative to that action, not the root of the slsa-github-generator repository
> 
> I don't think we can assume that. Sometimes the internal Actions need to access the **PROJECT_CHECKOUT_DIR**. Having the working directory the top-level root means we can keep the access the same and can easily identify what's invoking: something in the project repo or the builder repo. If we remove this, I'm afraid we're going to start wondering what the working directory is, and make mistakes overwriting things we should not. We want to be sure we don't overwrite builder's code, for example.

Right, Unfortunately a lot of actions need to do something outside of their own directory in the workspace. If there is any real consistency to be had, it's probably, as @laurentsimon is suggesting, always running scripts from the root of the workspace.

laurentsimon:
> > Yeah, I thought about that. We could probably set the working dir.
> 
> I think it would be somewhat clearer, but feel free to push back: we know this one works

I'd like to push back for the reasons in other comments :)

ianlewis:
> Fyi, Actions don't support working-directory, so setting it will add inconsistency across scripts and Actions well

it's moot to the argument at this point. but [actions do support working-directory](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsworking-directory)

laurentsimon:
> > Fyi, Actions don't support working-directory, so setting it will add inconsistency across scripts and Actions well
> 
> it's moot to the argument at this point. but [actions do support working-directory](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsstepsworking-directory)

Last time I tried it was an invalid Action, IIRC. I think it only works for the run command. I think that's what the document hints at with `where the command is run`.

PR #1223, opened 2022-11-09
feat: [DRAFT - DO NOT REVIEW] start building the nodejs binary

laurentsimon:
This PR adds the build command the the nodejs builder binary.
It follows a "generic" design to validate BYOB design.


laurentsimon:
@asraa any early feedback before I start adding some unit tests?

internal/builders/nodejs/build.go
```diff
@@ -15,95 +15,226 @@
...
+
+	// Generate the provenance metadata.
+	artifact := integration.Inputs.Artifacts[0]
+	name := strings.TrimSuffix(filepath.Base(artifact.Path), ".tgz")
+	metadata := slsaDryRunOutput{
+		name + ".intoto.jsonl": []slsaDryMetadata{
+			{
+				Name:    name,
+				Digests: artifact.Digests,
+				Steps:   steps,
```
asraa, line 99: This is intentionally one possible format for the buildConfig right? i agree this is fine to start with, but i'm curious if people want to customize their own, since they're BYO-ing anyway

asraa (commented):
one meta level comment, but likely better put in the design is: 
because users are writing both the run and run-dry, how do we know that these implementations actually match? the answer is probably that the users trusted these workflows (the outer layer) as much as they trust the inner.
the stronger way would be that run-dry provides the steps that the trusted builder can execute

internal/builders/nodejs/build.go
laurentsimon: good point. I hope we can have a recommended one by default, which is consistent with the buildType.The. we can update the version and let builders change the buildConfig / buildType.

I could rename Steps to buildConfig maybe? This way we can easily update in the next version by keeping the same format but allowing custom format. Wdut?

Note that all this code may become moot if the v1 specs only want the top-level inputs to be defined. If the specs go in this direction, we won't even need the dry run anymore: our reusable workflow will just provide the inputs it received.

laurentsimon:
> one meta level comment, but likely better put in the design is: because users are writing both the run and run-dry, how do we know that these implementations actually match? the answer is probably that the users trusted these workflows (the outer layer) as much as they trust the inner. the stronger way would be that run-dry provides the steps that the trusted builder can execute

The builder binary is the same for dry and not dry so its behavior should "match". Are you suggesting dry-run provides the commands for our reusable workflow to run or to the BYOB builder to run? It would work to let our reusable workflow run the command, but then we need to support builders that may want to run containers, or GH Actions, etc. Not everything is a binary.

Maybe we can make it work if builders provide a single command "entryPoint" which we can run as a shell command. The script provided by the builder can do whatever it wants (call the binary, container, etc). But if we do that, the script is just the same as the existing binary and still need to be in sync.

Let's brainstorm

internal/builders/nodejs/build.go
laurentsimon: so I will make the buildConfig be a format the dry-run the dry run decide.

internal/runner/runner.go
```diff
@@ -114,11 +114,11 @@ func (r *CommandRunner) runStep(ctx context.Context, step *CommandStep, dry bool
 	if err != nil {
 		return nil, err
 	}
-	posixEnv = append(posixEnv, "PWD="+pwd)
+	// posixEnv = append(posixEnv, "PWD="+pwd)
```
ianlewis, line 117: It might be a good idea to pass through the `PATH` environment variable here otherwise builders would need to set it.
```suggestion
	posixEnv = append(posixEnv, "PWD="+pwd)
	posixEnv = append(posixEnv, "PATH="+os.Getenv("PATH"))
```

internal/runner/runner.go
```diff
@@ -114,11 +114,11 @@ func (r *CommandRunner) runStep(ctx context.Context, step *CommandStep, dry bool
...
 
 	cmd := exec.CommandContext(ctx, name, args...)
 	cmd.Dir = pwd
-	cmd.Env = posixEnv
+	// cmd.Env = posixEnv
```
ianlewis, line 121: You need to set `cmd.Env` otherwise `exec.Command` adds `PWD` and [copies the current process' env vars](https://cs.opensource.google/go/go/+/refs/tags/go1.18.8:src/os/exec/exec.go;l=1162-1191;drc=3a41094107237535673bdbbdaabc10e54cb2b67c). It's hard to know what those env vars are as the logic changes a bit from version to version of Go.

```suggestion
	cmd.Env = posixEnv
```

internal/runner/runner.go
ianlewis, line 117: Alternatively we could set a basic PATH statically.
```suggestion
	posixEnv = append(posixEnv, "PWD="+pwd)
	posixEnv = append(posixEnv, "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin")
```

ianlewis:
Can we close this since the npm builder will use byob?

Issue #1225, opened 2022-11-09
[bug] Generic generator: provenance filename includes subject's subdirectory


pnacht:
**Describe the bug**
If a subject is created within a subdirectory (i.e. `./target/foo.jar`) and hashes computed and stored from the root, the provenance generator attempts to create the provenance as `./target/foo.jar.intoto.jsonl`, which fails since `./target` no longer exists in the provenance job.

**To Reproduce**
See workflow: https://github.com/pnacht/jackson-core/blob/d6d0af665a0c9d842b07e4468a75d2b59828df99/.github/workflows/main.yml 

And failed job: https://github.com/pnacht/jackson-core/actions/runs/3431333783/jobs/5720700225

It successfully builds `./target/jackson-core-2.14.0-SNAPSHOT.jar` but then the provenance job throws:

```
os.OpenFile(): open target/jackson-core-2.14.0-SNAPSHOT.jar.intoto.jsonl: no such file or directory
```

**Expected behavior**
The provenance should be generated successfully.

**Additional context**
@asraa suggested I generate the hashes within `./target` instead of the root. I will try this later.

asraa:
@ianlewis @laurentsimon do you think we should clarify document here and punt to the next release or try to get in for this one?

asraa:
Pushing a PR

ianlewis:
Thanks. This also seems to affect the Go builder: https://github.com/slsa-framework/slsa-github-generator/pull/1226#issuecomment-1309536515

We'll need to do a separate PR for that.

behnazh-w:
@asraa Looking at your [fix](https://github.com/slsa-framework/slsa-github-generator/pull/1226), what happens if two artifacts have the same base name but different paths, e.g., `a/b/pom.xml` and `a/c/pom.xml`?

ianlewis:
> @asraa Looking at your [fix](https://github.com/slsa-framework/slsa-github-generator/pull/1226), what happens if two artifacts have the same base name but different paths, e.g., `a/b/pom.xml` and `a/c/pom.xml`?

Currently there is only one provenance file written for multiple subjects so the default is to call it `multiple.intoto.jsonl` otherwise the name is taken from the [`provenance-name` input](https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/README.md#workflow-inputs)

behnazh-w:
[...]

PR #1226, opened 2022-11-09
[generic] fix attestation file creation when subject names are in subdirectories

asraa:

Updates https://github.com/slsa-framework/slsa-github-generator/issues/1225

The generic SLSA 3 generator attests to subjects given as input and their digest. However, if the subject is in a sub-directory, then it will fail creating the attestation because the sub-directory doesn't exist.

This retrieves the Base for the subject name, instead of the full path name.
@ianlewis I could also have created the sub-directory, but that made a lot less sense to me.

### DIFF
--- a/internal/builders/generic/attest.go
+++ b/internal/builders/generic/attest.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path"
 
 	intoto "github.com/in-toto/in-toto-golang/in_toto"
 	"github.com/spf13/cobra"
@@ -59,7 +60,8 @@ run in the context of a Github Actions workflow.`,
 			// validated. This is done by CreateNewFileUnderCurrentDirectory.
 			if attPath == "" {
 				if len(parsedSubjects) == 1 {
-					attPath = fmt.Sprintf("%s.intoto.jsonl", parsedSubjects[0].Name)
+					filename := path.Base(parsedSubjects[0].Name)
+					attPath = fmt.Sprintf("%s.intoto.jsonl", filename)
 				} else {
 					// len(parsedSubjects) > 1
 					attPath = "multiple.intoto.jsonl"
--- a/internal/builders/generic/attest_test.go
+++ b/internal/builders/generic/attest_test.go
@@ -343,3 +343,38 @@ func Test_attestCmd_invalid_path(t *testing.T) {
 	// If no error occurs we catch it here. SkipNow will exit the test process so this code should be unreachable.
 	t.Errorf("expected an error to occur.")
 }
+
+// Test_attestCmd_subdirectory_artifact tests the attest command when provided
+// subjects in subdirectories.
+func Test_attestCmd_subdirectory_artifact(t *testing.T) {
+	t.Setenv("GITHUB_CONTEXT", "{}")
+
+	// Change to temporary dir
+	currentDir, err := os.Getwd()
+	if err != nil {
+		t.Errorf("unexpected failure: %v", err)
+	}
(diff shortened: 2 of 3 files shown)


internal/builders/generic/attest.go
```diff
@@ -59,7 +60,8 @@ run in the context of a Github Actions workflow.`,
 			// validated. This is done by CreateNewFileUnderCurrentDirectory.
 			if attPath == "" {
 				if len(parsedSubjects) == 1 {
-					attPath = fmt.Sprintf("%s.intoto.jsonl", parsedSubjects[0].Name)
+					filename := path.Base(parsedSubjects[0].Name)
```
ianlewis, line 63: Adding a simple test in attest_test.go might also be good.

ianlewis (approved):
Does the Go builder handle this ok? I guess it controls the subject's output location when it's compiled.

internal/builders/generic/attest.go
asraa: done, thanks!

asraa:
> Does the Go builder handle this ok? I guess it controls the subject's output location when it's compiled.

Good call, checking on this. It doesn't currently enforce that even when setting the output from the config (it checks if it's under the current working directory, but even a sub-directory would count). Double-checking more.

ianlewis:
> > Does the Go builder handle this ok? I guess it controls the subject's output location when it's compiled.
> 
> Good call, checking on this. It doesn't currently enforce that even when setting the output from the config during dry run (it checks if it's under the current working directory, but even a sub-directory would count). Double-checking more.

If it's an issue I can follow up later in my day today.

asraa:
OK, I don't think so. The config could end up setting
```
binary: ./subdirectory/binary-{{ .Os }}-{{ .Arch }}
```

and this would be allowed, and would fail on `CreateNewFileUnderCurrentDirectory`

ianlewis:
> OK, I don't think so. The config could end up setting
> 
> ```
> binary: ./subdirectory/binary-{{ .Os }}-{{ .Arch }}
> ```
> 
> and this would be allowed, and would fail on `CreateNewFileUnderCurrentDirectory`

Yeah, lets go a separate PR for the Go builder.

internal/builders/generic/attest.go
```diff
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path"
```
laurentsimon, line 23: why not set a default value for the provenance file in https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/generator_generic_slsa3.yml#L37, and let users provide the name they want?

laurentsimon: I think we may run into problems with the secure-upload-artifact Actions. They currently assume files are in the root folder, I think

laurentsimon: edit: we use the upload-artifact, not the secure one so my last comment is wrong

laurentsimon (approved):

internal/builders/generic/attest.go
ianlewis: > why not set a default value for the provenance file in https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/generator_generic_slsa3.yml#L37?

I'd argue against this, if just for the fact that users would have to specify the input when the didn't before so it adds friction to upgrading and folks have complained previously about having to add new inputs.

ianlewis:
> Let's amend the doc to say the files don't support folders. I have not seen anyone need this feature when using Goreleaser.

I'd say it's reasonable to support it if goreleaser supports it, but we can prioritize it when someone complains.

Issue #1231, opened 2022-11-15
[bug] Environment variable value cannot contain =


cfergeau:
**Describe the bug**
I'm following the steps from https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/go/README.md in order to use SLSA in my project
When building [my project](https://gtihub.com/cfergeau/vfkit/tree/slsa] I want to set `CGO_CFLAGS=-mmacosx-version-min=11.0` in the environment. 
I added this to my `.slsa-goreleaser.yml`:
```
env:
  - CGO_ENABLED=1
  - CGO_CFLAGS=-mmacosx-version-min=11.0
```

but the 'build dry project' step fails with `invalid environment variable: CGO_CFLAGS=-mmacosx-version-min=11.0`
https://github.com/cfergeau/vfkit/actions/runs/3469089554/jobs/5795684619

This seems to be caused by this code https://github.com/slsa-framework/slsa-github-generator/blob/eb79b8b82d0161a56fb9892a15443abc1cd9aae5/internal/builders/go/pkg/config.go#L170-L173 which do not expect the env var line to contain 2 `=` signs.

**To Reproduce**

1. Add a `env: -CGO_CFLAGS=a=b` line to your `.slsa-goreleaser-yml`
2. Trigger the slsa github actions workflow

**Expected behavior**
I think a change such as this one would let me use `=` in env vars.
```
diff --git a/internal/builders/go/pkg/config.go b/internal/builders/go/pkg/config.go
index 1df1bf7..eba3e39 100644
--- a/internal/builders/go/pkg/config.go
+++ b/internal/builders/go/pkg/config.go
@@ -167,11 +167,11 @@ func validateVersion(cf *goReleaserConfigFile) error {
 func (r *GoReleaserConfig) setEnvs(cf *goReleaserConfigFile) error {
 	m := make(map[string]string)
 	for _, e := range cf.Env {
-		es := strings.Split(e, "=")
-		if len(es) != 2 {
+		name, value, present := strings.Cut(e, "=")
+		if value == "" || !present {
 			return fmt.Errorf("%w: %s", ErrorInvalidEnvironmentVariable, e)
 		}
-		m[es[0]] = es[1]
+		m[name] = value
 	}
 
 	if len(m) > 0 {
```

laurentsimon:
+1, good catch and thanks for filing the bug. Do you want to send a PR for this?

cfergeau:
Sure, I'll send a PR!

PR #1232, opened 2022-11-16
builder: go: Allow equal signs in env vars

cfergeau:
If .slsa-goreleaser.yml contains an env var with multiple '=' signs:
```
env:
  - CGO_ENABLED=1
  - CGO_CFLAGS=-mmacosx-version-min=11.0
```

the 'build dry project' GH workflow step fails with "invalid environment
variable: CGO_CFLAGS=-mmacosx-version-min=11.0"
This is caused by the way the env vars are parsed in GoReleaserConfig:setEnvs
which only allow for a single '=' sign.

This commit fixes this by splitting the env string at the first equal sign,
first part is the env var name and won't contain '=' signs, second part
is its value and can contain any number of '='.

This also adds unit tests for this situation, and for env vars with no
values (`CGO_CFLAGS=`).

This fixes https://github.com/slsa-framework/slsa-github-generator/issues/1231

### DIFF
--- a/internal/builders/go/pkg/config.go
+++ b/internal/builders/go/pkg/config.go
@@ -167,11 +167,11 @@ func validateVersion(cf *goReleaserConfigFile) error {
 func (r *GoReleaserConfig) setEnvs(cf *goReleaserConfigFile) error {
 	m := make(map[string]string)
 	for _, e := range cf.Env {
-		es := strings.Split(e, "=")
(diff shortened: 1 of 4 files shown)


laurentsimon (approved):
Thank you!

ianlewis:
Good, you allowed no value. 👍  Thanks!

PR #1235, opened 2022-11-17
docs(generic-generator): clarify that created provenance is encapsulated

### LINKED ISSUE #1186: [docs] More explicit instructions for verifying provenance content
We include examples of the provenance format in the docs but when generating an `.intoto.jsonl` file the provenance is wrapped in a DSSE envelope and base64 encoded so it's not totally clear how to view it.

We should provide a simple command users can run on the file to view the provenance json.

diogoteles08:
After exposing the expected provenance format, clarify that the actual result will be encapsulated in a DSSA envelope and encoded in base64. Also includes quick command to decode the human readable JSON.

Closes #1186

Signed-off-by: Diogo Teles Sant'Anna <diogoteles@google.com>


asraa (approved):
Thank you!

Issue #1177, opened 2022-10-28
[feature] Return exit code instead of failing jobs


behnazh-w:
**Is your feature request related to a problem? Please describe.**
Because GHA does not support [`continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error ) at `job` level, it's hard to add constraints based on the reusable workflow SLSA generator's success or failure. At the moment when the generator job fails, the caller workflow is marked as failed too. 

It would be useful to return the exit code as output instead of failing the job and allow the caller to decide to fail the workflow and take action.

ianlewis:
I think there is a job level [`jobs.<job_id>.continue-on-error`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error). I believe you can check the result via the [`needs`](https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context) context. Does that solve your problem?

behnazh-w:
Do you know how to make it work with reusable workflows? This is not a valid GHA:

```yaml
  provenance:
    needs: [build, release]
    continue-on-error: true
    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.1
    with:
      base64-subjects: ${{ needs.build.outputs.artifacts-sha256 }}
      private-repository: true
      compile-generator: true
    permissions:
      actions: read # To read the workflow path.
      id-token: write # To sign the provenance.
      contents: write # To add assets to a release.
```

ianlewis:
Ah, right. I see what you mean. Maybe we can consider something like a `continue-on-error` input or something that wouldn't make it outright fail but would return a result in the output.

behnazh-w:
Yes, that would be great.

laurentsimon:
So we need to set `return-on-error: true` for all steps of the generic generators, then collect the results for each in a "final" step. Here is an example https://github.com/laurentsimon/slsa-on-github-test/blob/main/.github/workflows/run-error.yml

If everyone is OK with this change, I can sent a PR. I think we only needs this change for generic generators, since builder failures means the build failed (not just the provenance generation).

behnazh-w:
>So we need to set return-on-error: true for all steps of the generic generators,

I guess you meant `continue-on-error: true`.

[...]

PR #1237, opened 2022-11-18
feat: add `continue-on-error` for generic generators

laurentsimon:
Updates #1177 

Instead of failing the workflow, users can pass a `continue-on-error` to ignore failure.

Note: let's not include this in the container generator release. Let's wait until we have e2e tests + validated it works as expected.


.github/workflows/generator_container_slsa3.yml
```diff
@@ -48,12 +52,17 @@ on:
         required: false
         type: boolean
         default: false
```
ianlewis, line 54: I think for #1177 we want to add our own `continue-on-error` input to the workflow that defaults to `false` (for backwards compatibility). This gives the user more control over what they want to do and doesn't force them to check `outcome` if they want to fail fast.

This input would be passed to the `continue-on-error` inputs further down. `outcome` would thus only be valid if the user passed `continue-on-error: true` to the reusable workflow.

laurentsimon: OK. So `continue-on-error` cannot take a variable as input. So I'll set them all to `true` and add a final step in each job that will exit (or not) based on the value of `continue-on-error`. Correct?

ianlewis: Lovely. Yeah, I suppose that will have to be how it's implemented...

laurentsimon: PTAL. I still needed the additional `final` job with `always()`, otherwise the overall run would be reported as a failure.

.github/workflows/generator_container_slsa3.yml
```diff
@@ -134,3 +160,23 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [detect-env, generator]
+    # Note: always run even if nedded jobs are skipped.
+    if: always()
+    steps:
+      - name: Final outcome
+        env:
+          SUCCESS: ${{ needs.detect-env.outputs.outcome != 'failure' && steps.generator.outputs.outcome != 'failure' }}
+          CONTINUE: ${{ inputs.continue-on-error }}
+        run: [ "$CONTINUE" == "true" ] || [ "$SUCCESS" == "true" ] || exit 27
```
laurentsimon, line 182: @ianlewis any ida why this is failing but the other similar occurrences (e.g. line 168) are not flagged?

.github/workflows/generator_container_slsa3.yml
```diff
@@ -62,7 +72,14 @@ jobs:
     steps:
       - name: Detect the generator ref
         id: detect
+        continue-on-error: true
         uses: slsa-framework/slsa-github-generator/.github/actions/detect-workflow@main
+      
+      - name: Final outcome
+        id: final
+        env:
+          SUCCESS: ${{ steps.detect.outcome != 'failure' }}
```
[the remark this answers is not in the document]
laurentsimon: this is built-in and defined by GitHub. All steps have this `outcome` property set to `success` or `failure`

.github/workflows/generator_container_slsa3.yml
```diff
@@ -74,6 +91,8 @@ jobs:
   # with the release at the ref detected by `detect-env`. This must be a tag
   # reference.
   generator:
```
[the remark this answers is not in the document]
laurentsimon: the last job in the workflow (`final`) always runs using `if: always()` and collects all the `outcome`s from the jobs it depends on. So if `generator` job is skipped because of a an error of a previous job, the final job will still collect at least one of the failing job's `outcome` and fail accordingly. The `generator` job's `outcome` would be empty, but the previous job's won't be so `final` will still find the failure. 

Let me know if this makes sense or if you think there's a logic error.

.github/workflows/generator_container_slsa3.yml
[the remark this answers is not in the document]
asraa: aah thanks! i did a quick search but i was just wrong. it's part of step context indeed

.github/workflows/generator_container_slsa3.yml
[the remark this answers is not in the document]
asraa: My question was that wouldn't `generator` still attempt to run if `detect-env` had a `continue-on-error`? Maybe the `needs` was actually the key (is `needs` requiring that the previous job succeeds?)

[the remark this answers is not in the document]
asraa: Got it: yes, `needs` requires success https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

.github/workflows/generator_container_slsa3.yml
ianlewis: Not sure why 168 isn't picked up but I think you'll want to do something like:

```yaml
run: |
```

So it's interpreted as a multi-line string rather than as via YAML's type inference. Probably yamllint thinks you're trying to create a yaml list. e.g.

```yaml
run: ["foo", "bar"]
```

This holds for all the other places where you use `run`.

.github/workflows/generator_container_slsa3.yml
ianlewis: BTW, the docs seem to imply that `inputs` can be used in `jobs.<job_id>.steps.continue-on-error`. I wonder if the docs are wrong or is it just not available for reusable workflows? I assume you tried it and got an error?
https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability

.github/workflows/generator_container_slsa3.yml
```diff
@@ -48,12 +52,18 @@ on:
...
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   # detect-env detects the reusable workflow's repository and ref for use later
   # in the workflow.
   detect-env:
     outputs:
+      outcome: ${{ steps.final.outputs.outcome }}
```
ianlewis, line 66: nit: Does this not work?
```suggestion
      outcome: ${{ steps.detect.outcome }}
```

.github/workflows/generator_container_slsa3.yml
```diff
@@ -82,6 +101,8 @@ jobs:
       packages: write # Needed to login and upload attestations to ghcr.io.
     steps:
       - name: Generate builder
+        id: generate-builder
+        continue-on-error: true
```
ianlewis, line 105: I wonder if we can pass the `continue-on-error` input here?
https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability

Also on `cosign-install`, `login`, and `sign-prov`. I assume `login` and `sign-prov` should work at least since those are just basic job steps.

.github/workflows/generator_container_slsa3.yml
```diff
@@ -134,3 +160,23 @@ jobs:
...
             --force \
             "${UNTRUSTED_IMAGE}@${UNTRUSTED_DIGEST}"
+
+      - name: Final outcome
+        id: final
+        env:
+          SUCCESS: ${{ steps.generate-builder.outcome != 'failure' && steps.cosign-install.outcome != 'failure' && steps.login.outcome != 'failure' && steps.sign-prov.outcome != 'failure' }}
+        run: echo "outcome=$([ "$SUCCESS" == "true" ] && echo "success" || echo "failure")" >> $GITHUB_OUTPUT
+
+  # final fails or suceeds based on the value of `inputs.continue-on-error`
```
ianlewis, line 170:
```suggestion
  # final fails or succeeds based on the value of `inputs.continue-on-error`
```

.github/workflows/generator_container_slsa3.yml
```diff
@@ -134,3 +160,23 @@ jobs:
...
+  # and the outcome of previous jobs.
+  final:
+    runs-on: ubuntu-latest
+    needs: [detect-env, generator]
+    # Note: always run even if nedded jobs are skipped.
```
ianlewis, line 175:
```suggestion
    # Note: always run even if needed jobs are skipped.
```

.github/workflows/generator_generic_slsa3.yml
```diff
@@ -161,22 +190,47 @@ jobs:
...
           files: |
             ${{ needs.generator.outputs.provenance-name }}
+
+      - name: Final outcome
+        id: final
+        env:
+          SUCCESS: ${{ steps.checkout-builder.outcome != 'failure' && steps.download-prov.outcome != 'failure' && steps.release.outcome != 'failure' }}
+        run: echo "outcome=$([ "$SUCCESS" == "true" ] && echo "success" || echo "failure")" >> $GITHUB_OUTPUT
+
+  # final fails or suceeds based on the value of `inputs.continue-on-error`
```
ianlewis, line 224:
```suggestion
  # final fails or succeeds based on the value of `inputs.continue-on-error`
```

.github/workflows/generator_generic_slsa3.yml
```diff
@@ -161,22 +190,47 @@ jobs:
...
+  # and the outcome of previous jobs.
+  final:
+    runs-on: ubuntu-latest
+    needs: [detect-env, generator, create-release]
+    # Note: always run even if nedded jobs are skipped.
```
ianlewis, line 229:
```suggestion
    # Note: always run even if needed jobs are skipped.
```

ianlewis (approved):
Comments on `generator_container_workflow.yml` generally apply to `generator_generic_workflow.yml` as well.

ianlewis:
You might also consider adding a pre-submit with `continue-on-error: true`. It will also just succeed but we'd like to make sure it doesn't just fail outright in this mode.

.github/workflows/generator_container_slsa3.yml
laurentsimon: Should work. Shall I use you suggestion or keep as-is so that all the jobs have the same pattern and is easier to review / update?

.github/workflows/generator_container_slsa3.yml
laurentsimon: You can pass it, but it's interpreted as a string rather than a boolean and the workflow is deemed invalid. I tried `${{ inputs.continue-on-error == 'true' }}` but still was complaining about not being a boolean. Let me know if you have other ideas.

.github/workflows/generator_container_slsa3.yml
laurentsimon: no working because it's interpreted as a string. `${{ inputs.continue-on-error == 'true' }}` also interpreted as string. See https://github.com/laurentsimon/slsa-on-github-test/blob/main/.github/workflows/run-error.yml#L6

.github/workflows/generator_container_slsa3.yml
laurentsimon: Thanks!

laurentsimon:
> You might also consider adding a pre-submit with `continue-on-error: true`. It will also just succeed but we'd like to make sure it doesn't just fail outright in this mode.

+1. Done, PTAL

.github/workflows/generator_container_slsa3.yml
ianlewis: huh. Strange... It's not that important so we could clean it up later if we find a way.

.github/workflows/generator_container_slsa3.yml
ianlewis: I think its ok as is.

.github/workflows/generator_generic_slsa3.yml
```diff
@@ -76,8 +86,15 @@ jobs:
...
       - name: Detect the generator ref
         id: detect
+        continue-on-error: true
         uses: slsa-framework/slsa-github-generator/.github/actions/detect-workflow@v1.3.0
 
+      - name: Final outcome
+        id: final
+        env:
+          SUCCESS: ${{ steps.detect.outcome != 'failure' }}
+        run: echo "outcome=$([ "$SUCCESS" == "true" ] && echo "success" || echo "failure")" >> $GITHUB_OUTPUT
```
ianlewis, line 96:
```suggestion
        run: |
          echo "outcome=$([ "$SUCCESS" == "true" ] && echo "success" || echo "failure")" >> $GITHUB_OUTPUT
```

.github/workflows/generator_container_slsa3.yml
```diff
@@ -62,7 +72,14 @@ jobs:
...
       - name: Detect the generator ref
         id: detect
+        continue-on-error: true
         uses: slsa-framework/slsa-github-generator/.github/actions/detect-workflow@v1.3.0
+      
+      - name: Final outcome
+        id: final
+        env:
+          SUCCESS: ${{ steps.detect.outcome != 'failure' }}
+        run: echo "outcome=$([ "$SUCCESS" == "true" ] && echo "success" || echo "failure")" >> $GITHUB_OUTPUT
```
ianlewis, line 82:
```suggestion
        run: |
          echo "outcome=$([ "$SUCCESS" == "true" ] && echo "success" || echo "failure")" >> $GITHUB_OUTPUT
```

.github/workflows/pre-submit.e2e.generic.default.yml
```diff
@@ -39,3 +39,33 @@ jobs:
           BINARY: "binary-name"
           PROVENANCE: ${{ needs.build.outputs.provenance-name }}
         run: ./.github/workflows/scripts/pre-submit.e2e.generic.default.sh
+
+  build-continue-error:
```
ianlewis, line 43: We probably want a pre-submit for the container workflow too since we are trying to GA it.

.github/workflows/pre-submit.e2e.generic.default.yml
```diff
@@ -39,3 +39,33 @@ jobs:
...
+      # echo "2e0390eb024a52963db7b95e84a9c2b12c004054a7bad9a97ec0c7c89d4681d2    binary-name" | base64 -w0
+      base64-subjects: "MmUwMzkwZWIwMjRhNTI5NjNkYjdiOTVlODRhOWMyYjEyYzAwNDA1NGE3YmFkOWE5N2VjMGM3Yzg5ZDQ2ODFkMiAgICBiaW5hcnktbmFtZQo="
+      compile-generator: true
+      continue-on-error: true
+
+  verify-continue-error:
+    # NOTE: this name is used as the status check name and by protected
+    # branches for required status checks. It should have a unique name among
+    # other pre-submits.
+    name: verify generic provenance
```
ianlewis, line 59: This job name needs to be unique among all pre-submit jobs. Give this a different name so we can add it to protected branch checks.
```suggestion
    name: verify generic provenance (continue-on-error)
```

.github/workflows/pre-submit.e2e.generic.default.yml
```diff
@@ -39,3 +39,33 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [build-continue-error]
+    if: ${{ always() }}
+    steps:
```
ianlewis, line 63: Add a job step to verify the `outcome` output.

.github/workflows/pre-submit.e2e.generic.default.yml
```diff
@@ -39,3 +39,33 @@ jobs:
...
+
+  build-continue-error:
+    permissions:
+      id-token: write # For signing.
+      contents: write # For asset uploads.
+      actions: read # For reading workflow info.
+    uses: ./.github/workflows/generator_generic_slsa3.yml
+    with:
+      # echo "2e0390eb024a52963db7b95e84a9c2b12c004054a7bad9a97ec0c7c89d4681d2    binary-name" | base64 -w0
+      base64-subjects: "MmUwMzkwZWIwMjRhNTI5NjNkYjdiOTVlODRhOWMyYjEyYzAwNDA1NGE3YmFkOWE5N2VjMGM3Yzg5ZDQ2ODFkMiAgICBiaW5hcnktbmFtZQo="
```
ianlewis, line 51: nit: We could probably add another pre-submit to check 'failure' by passing an invalid base64 subjects line.

.github/workflows/pre-submit.e2e.generic.default.yml
laurentsimon: we removed the outcome from the reusable workflow, and use `continue-on-error` instead. So I don't think I can add the check. Right?

laurentsimon: done. PTAL

.github/workflows/pre-submit.e2e.generic.default.yml
laurentsimon: done. PTAL

.github/workflows/pre-submit.e2e.generic.default.yml
ianlewis: Ah, I forgot we can't push to ghcr.io on pull_request. We'll need to make an e2e test.

laurentsimon:
I've added the final job's `outcome` back, because there is not built-in field to get the failure / success status. There is a `result` field, but it's always set to 'success' if `continue-on-error: true`.

Issue #358, opened 2022-06-20
Support for pull_request


ianlewis:
This is a tracking issue for supporting `pull_request` events. Please comment regarding your use case.

asraa:
Another use-case: is to test the configuration.

For example, I didn't realize that env vars are case-sensitive and ran into this problem
```
env variable empty or not set: {{ .Env.Version }}
```

because I had been using `VERSION` in my `evaluted-envs`.

See https://github.com/slsa-framework/slsa-verifier/pull/298

ianlewis:
The OpenZepplin folks mentioned that they would like support for pull requests
https://github.com/OpenZeppelin/defender-client/issues/277

asraa:
Is there anything stopping delegator from PR events? I believe it should be able to output unsigned attestations if htere's a guard on the sign-attestations step.

ianlewis:
Right. I think signing is the biggest thing though there are some other subtle differences as well. I think we had issues with which git sha we pick up when generating the provenance? I can't exactly remember but I think `detect-workflow-js` already supports it: https://github.com/slsa-framework/slsa-github-generator/blob/942ce4027d93050826def3a8677d51c851419e97/.github/actions/detect-workflow-js/src/detect.ts#L84

pdxjohnny:
It would be nice to have `id-token` scoped to reflect it being issued within the context of a pull request so that test artifacts generated during pull request runs such as container images could be uploaded to appropriately access controlled registries.

Issue #131, opened 2022-05-30
`id-token: write` not available on pull_requests


ianlewis:
Adding `id-token: write` to workflows triggered by `pull_request` events doesn't seem to be allowed.

This seems a bit problematic that we would expect that folks only run our workflow when a tagged. I can imagine folks would run into issues if they only exercise the workflows when releasing new versions as there would be significant time and changes to their repo in between releases. If the builders don't run on `pull_request` it will be hard to ensure that new code works with the reusable workflows.

For comparison, go-releaser's example workflow runs on `pull_request`. 
https://github.com/goreleaser/goreleaser-action#workflow

laurentsimon:
Goreleaser works on PR but they cannot sign with keys, because the keys need to be stored in secrets that are not available on PRs.

To have access to the secrets or OIDC tokens, we could support pull_request_target instead. 

@ianlewis are you suggesting we support PRs and not output provenance in this case (to be in par with Goreleaser)?

ianlewis:
@laurentsimon I'm not actually sure what's best right now. I just think we probably need to support PRs in some capacity even if we don't sign provenance because it would introduce differences between how building in PR pre-submits work and how it works on releases. I think it may be a future point of user friction.

Like for example, maybe just running the builder on PRs is the right way to go?

laurentsimon:
I think that's fine, yes. I considered this to be low priority for v1, and wanted to see if we get push back from people. Wdut?

ianlewis:
> I think that's fine, yes. I considered this to be low priority for v1, and wanted to see if we get push back from people. We can tackle this in the next iteration. Wdut?

I agree. I haven't included it in the v1 milestone.

ianlewis:
As an aside, we currently use the `NilClientProvider` for pull requests which returns nil for both the OIDC and GitHub clients, but we should still be able to use the GitHub API in pull requests as it uses `github.token` for auth.
https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

That should allow us to set the invocation entry point properly and not have special code for that.

ianlewis:
We can also look into supporting the `[pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target)` event which may allow us to get the right permissions because it runs using workflow code at the base ref of the PR.

laurentsimon:
agreed, that's the way to support pull requests if we need to.

[...]

PR #1238, opened 2022-11-18
[go] [generic] add pull request support without signing

asraa:

1. Should we display some github log warning?
2. This sort of makes the check `IsPresubmitTest` not necessary. Is that OK? I'm not sure if we want to continue gating on that, or maybe we want to add an explicit flag to users like "snapshot" that is required to allow pull_requests to generate the predicate (that way users don't accidentally generate unsigned provenance)

### DIFF
--- a/internal/builders/generic/attest.go
+++ b/internal/builders/generic/attest.go
@@ -56,6 +56,11 @@ run in the context of a Github Actions workflow.`,
 				check(errors.New("expected at least one subject"))
 			}
 
+			// If the environment does not have access to an OIDC provider, use a nil one.
+			if provider == nil && !github.HasOIDCClient() {
+				provider = &slsa.NilClientProvider{}
+			}
+
 			// NOTE: The provenance file path is untrusted and should be
 			// validated. This is done by CreateNewFileUnderCurrentDirectory.
 			if attPath == "" {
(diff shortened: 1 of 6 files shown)


internal/builders/go/pkg/provenance.go
```diff
@@ -70,6 +70,11 @@ func GenerateProvenance(name, digest, command, envs, workingDir string, s signin
 		return nil, err
 	}
 
+	// If the environment does not have access to an OIDC provider, use a nil one.
+	if provider == nil && !github.HasOIDCClient() {
```
laurentsimon, line 74: are there cases where provider is not `nil` and has not OIDC client? Existing code suggest no.

internal/builders/generic/attest.go
```diff
@@ -80,29 +85,19 @@ run in the context of a Github Actions workflow.`,
 			}
 			if provider != nil {
```
laurentsimon, line 86: Can we remove the condition `if provider != nil`?

laurentsimon (approved):
LGTM. I agree with you it'd be nice for users to specify an input acknowledging they want to opt-in for dry run. This will force them to read the doc that explains that on triggers with no OIDC clients, they should not expect it. Otherwise I suspect some user will miss it and complain?

if we don't want to commit to a new input argument, we could start by having this clearly in the documentation and re-visit in a few months.

ianlewis (approved):

ianlewis:
Updates #131

ianlewis:
Updates #358

ianlewis:
@asraa This is pretty old. Should we close it?

asraa:
> @asraa This is pretty old. Should we close it?

I think so. If we achieve this the better route to go is to refactor them to use the signing actions

PR #1240, opened 2022-11-18
Fix reference Gradle workflow.

DanAlbert:
runs-on is required: https://github.com/DanAlbert/slsa-gradle-example/actions/runs/3500410886
contents write permission is required: https://github.com/DanAlbert/slsa-gradle-example/actions/runs/3500417144

PR #1247, opened 2022-11-21
Fix permissions in doc

ianlewis:
I thought for sure we had fixed all these.

Related: #1240 


laurentsimon (approved):
I thought too!

PR #1252, opened 2022-11-21
Fix docs for goreleaser with the generic generator to include docker di…

gal-legit:
goreleaser [latest release](https://github.com/goreleaser/goreleaser/releases/tag/v1.13.0) includes a couple of new features that are useful for provenance generation:

* https://github.com/goreleaser/goreleaser/commit/5eb1e4ad0d41c6ee06db34183689a42e02a77fe0: feat: add digest to artifacts info of published docker images (https://github.com/goreleaser/goreleaser/pull/3540) (@gal-legit)
* https://github.com/goreleaser/goreleaser/commit/76dc0b559e073eca268700cc5799c1f00ffdf466: feat: output checksums to artifacts info (https://github.com/goreleaser/goreleaser/pull/3548) (@gal-legit)

The first one (published docker images) actually enables users to include the built dockers in their provenance documents.
The second one (checksums) is added mainly to simplify the usage.

This pull request updates the instructions in the documentation for the generic builder with goreleaser, to take advantage of these new features.

### DIFF
--- a/internal/builders/generic/README.md
+++ b/internal/builders/generic/README.md
@@ -331,7 +331,7 @@ jobs:
       hashes: ${{ steps.hash.outputs.hashes }}
 ```
 
-2. Add an `id: run-goreleaser` field to your goreleaser step:
+2. Add an `id: run-goreleaser` field to your goreleaser step. Use goreleaser version >= v1.13.0 to enable provenance generation for dockers.
 
 ```yaml
     steps:
@@ -352,6 +352,10 @@ jobs:
   run: |
     set -euo pipefail
     hashes=$(echo $ARTIFACTS | jq --raw-output '.[] | {name, "digest": (.extra.Digest // .extra.Checksum)} | select(.digest) | {digest} + {name} | join("  ") | sub("^sha256:";"")' | base64 -w0)
+    if test "$hashes" = ""; then # goreleaser < v1.13.0
+      checksum_file=$(echo "$ARTIFACTS" | jq -r '.[] | select (.type=="Checksum") | .path')
+      hashes=$(cat $checksum_file | base64 -w0)
+    fi
     echo "hashes=$hashes" >> $GITHUB_OUTPUT
 ```
 
--- a/internal/builders/generic/README.md
+++ b/internal/builders/generic/README.md
@@ -325,7 +325,7 @@ generate SLSA3 provenance by updating your existing workflow with the steps indi
 **Notes**:
 - Make sure you did not disable checksum generation in the goreleaser yml.
 - Make sure you specified sha256 as the algorithm for the checksum or left it empty (sha256 is the default).
-- To enable provenance generation for dockers (as well as artifacts), use goreleaser version >= v1.13.0.
+- To enable provenance generation for dockers (as well as artifacts), use [goreleaser version >= v1.13.0](https://github.com/goreleaser/goreleaser/releases/tag/v1.13.0).
 
 1. Declare an `outputs` for the GoReleaser job:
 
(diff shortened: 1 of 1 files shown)


internal/builders/generic/README.md
```diff
@@ -351,9 +351,8 @@ jobs:
     ARTIFACTS: "${{ steps.run-goreleaser.outputs.artifacts }}"
   run: |
     set -euo pipefail
-
```
laurentsimon, line 354: Just to be clear: the new feature does not break backward compatibility, correct?

If we update the code, will users of the older goreleaser versions break?

laurentsimon (commented):
Thanks for the PR!

internal/builders/generic/README.md
gofri: I'm afraid it will; it won't work for older versions. 
I can edit the PR to also include the instructions for older versions (the current instructions) if that's preferred.

laurentsimon: Yes, please edits the PR to account for older version.

Btw, since it's breaking backward compatibility, should you not bump the Action tag to a major change, like v4.0.0 instead of v3.2.0?

gofri: np, will do. 
Regarding the version bump -
1. The github action actually isn't relevant here. The version of the goreleaser is the one affecting the output (the version of goreleaser is passed as an argument to the action).
2. The chnage in goreleaser is backwards compatible, i.e. any code written for older versions should still work. The code that breaks is the code I added in this PR, which uses the new features that weren't available in older goreleaser versions (specifically, using the checksum file still works, so the current instructions won't break).

laurentsimon: Thanks for the clarification

gal-legit: I pushed a new commit.
btw, I didn't want to amend (& force-push) the old commit to avoid noise, so note that I accidentally wrote "generic builder" rather than "generic generator" in the commit message, so you might want to fix that when you squash&merge.

gal-legit: p.s. the instructions currently assume that the user didn't disable checksum generation in the goreleaser (see https://goreleaser.com/customization/checksum/).
It might be an obvious assumption, but disabling it would cause the job to hang (because of that `cat $checksum_file` when `checksum_file` is an empty string).
We might want to add a clarification for that. wdyt?

laurentsimon: Yes, please add a note. Thank you.

internal/builders/generic/README.md
```diff
@@ -322,6 +322,11 @@ This section explains how to generate non-forgeable SLSA provenance with existin
 If you use [GoReleaser](https://github.com/goreleaser/goreleaser-action) to generate your build, you can easily
 generate SLSA3 provenance by updating your existing workflow with the steps indicated in the workflow below:
 
+**Notes**:
+- Make sure you did not disable checksum generation in the goreleaser yml.
+- Make sure you specified sha256 as the algorithm for the checksum or left it empty (sha256 is the default).
+- To enable provenance generation for dockers (as well as artifacts), use goreleaser version >= v1.13.0.
```
laurentsimon, line 328: Do you want to add a link to the new goreleaser feature / doc?

laurentsimon (approved):
Thanks. I proposed adding a link to goreleaser feature / doc for the docker. Let me know what you think.

internal/builders/generic/README.md
gal-legit: actually it's transparent so there's no doc for it, but I'll add a link to the release there.

Issue #547, opened 2022-07-14
[bug] 


melba-lopez:
**Describe the bug**
Improve repository's OpenSSF Scorecard score (currently at 7.1)

**To Reproduce**
`docker run -e GITHUB_AUTH_TOKEN gcr.io/openssf/scorecard:stable --show-details --repo=https://github.com/slsa-framework/slsa-github-generator --format=json > scorecard_slsa-framework_slsa-github-generator.json`

**Expected behavior**

- Branch Protections could be improved
- CII-Best-Practices Badge could be obtained
- Project should be Fuzzed
- Security Policy should be created
- Token Permissions should follow principle of least priveledge

**Screenshots**
![image](https://user-images.githubusercontent.com/101211710/179085462-1b9e1eef-4a10-4bbc-89e8-d7b2c79c0d91.png)
![image](https://user-images.githubusercontent.com/101211710/179085517-21910b84-a436-4cc7-92ee-39ad5e09787e.png)
![image](https://user-images.githubusercontent.com/101211710/179086284-7c96bd7c-a6c3-49f1-ae82-0a6d5aaedb07.png)
![image](https://user-images.githubusercontent.com/101211710/179086133-4b000ab8-a769-4654-a57d-1f984a4b07f0.png)
![image](https://user-images.githubusercontent.com/101211710/179085764-dd56ce86-4883-41c8-a5c8-b1123680e02e.png)

**Additional context**
Attempted to upload the JSON file, but github does not allow me to.  Related to recommendation of securing our repos: https://github.com/slsa-framework/slsa/issues/424

laurentsimon:
You know you can send us some PR and get rewarded at sos.dev, right?

melba-lopez:
I thought PRs were mostly meant for doc/code changes??  I know I am unable to make these changes myself since I don't have repo access.  So not sure how that would work?

laurentsimon:
The permission changes could be changes you propose via a PR. If you identify some critical projects and get them to accept your PR, you can use sos.dev to be rewarded

laurentsimon:
I thought we already set the permissions as read-only in all our workflows. If we don't for some, please feel free to send a PR

jspeed-meyers:
Huh, `scorecard` says, as of 10/19/22 for this repo, `0/10` for `Token-Permissions` and that `non read-only tokens detected`. Let me do a little digging.

jspeed-meyers:
[...]

PR #1258, opened 2022-11-24
Update GHA token permissions for generic container workflow

### LINKED ISSUE #1256: [bug] Set container workflow top level token permissions
Scorecard pointed out we have no top level permissions on the container workflow:
Warn: no topLevel permission defined: .github/workflows/generator_container_slsa3.yml

Related to #547

### LINKED ISSUE #1257: [bug] Don't use `package: write` permissions for container workflow
Using `package: write` on the workflow is convenient but is not setting least privilege in the case where users are not using ghcr.io. We should force users to set the username and password and log into ghcr.io manually.

Related to #547

ianlewis:
Fixes #1256 
Fixes #1257 
Updates #547 

1. Sets top level token permissions for the container generator workflow.
2. Updates container generator workflow to no longer require `package: write` permissions on the GHA token. Instead we require users of ghcr.io to login using a PAT.

### DIFF
--- a/.github/workflows/generator_container_slsa3.yml
+++ b/.github/workflows/generator_container_slsa3.yml
@@ -14,6 +14,9 @@
 
 name: SLSA container image provenance
 
+permissions:
+  contents: read
+
 env:
   # Generator
   BUILDER_BINARY: slsa-generator-container-linux-amd64 # Name of the binary in the release assets.
@@ -79,7 +82,7 @@ jobs:
     permissions:
       id-token: write # Needed to get OIDC token for keyless signing.
       actions: read # Needed to read workflow info.
-      packages: write # Needed to login and upload attestations to ghcr.io.
+      # packages: write # Needed to login and upload attestations to ghcr.io.
     steps:
       - name: Generate builder
         uses: slsa-framework/slsa-github-generator/.github/actions/generate-builder@v1.3.0
--- a/.github/workflows/generator_container_slsa3.yml
+++ b/.github/workflows/generator_container_slsa3.yml
@@ -14,8 +14,7 @@
 
 name: SLSA container image provenance
 
-permissions:
-  contents: read
+permissions: {}
 
 env:
   # Generator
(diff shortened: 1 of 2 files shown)


laurentsimon:
Note that it's debatable that long-lived PAT tokens are a more secure option to ephemeral GITHUB_TOKEN. But we can't have both. Starting as you suggest in this PR means we can later add a `package:write` if needed.

laurentsimon (approved):

ianlewis:
> Note that it's debatable that long-lived PAT tokens are a more secure option to ephemeral GITHUB_TOKEN. But we can't have both. Starting as you suggest in this PR means we can later add a `package:write` if needed. We need to follow-up whether they can compute the permissions for a reusable workflow dynamically at runtime - instead of statically.

Yeah, I understand that. This at least makes it so you don't have to give `packages: write` permissions even if you're using a repository other than ghcr.io. Hopefully this will be fixed by the more granular PAT tokens that are in beta.

ianlewis:
BTW, I tried to pass the `GITHUB_TOKEN` with `package:write` permission to the workflow as an input, basically giving the token more permissions than the workflow asked for, but that didn't seem to work. It still required the workflow to request `packages:write` for it to work.

laurentsimon:
> Hopefully this will be fixed by the more granular PAT tokens that are in beta. 

the fine-grained tokens seem to be scoped in time and always expire (at least in the beta), which makes them unsuitable for workflows. (We can't expect uses to update PATs every month or so)

> I'm pretty sure permissions can't be computed dynamically.

I just meant that GH should be able to check the permissions given by the caller at runtime. Technically they should be able to do that, without doing a validation statically.

ianlewis:
> the fine-grained tokens seem to be scoped in time and always expire (at least in the beta), which makes them unsuitable for workflows. (We can't expect uses to update PATs every month or so)

Yeah, that seems odd if they allow more granular scopes, users should be more comfortable with having long-term tokens. Maybe that's some feedback for GitHub...

> I just meant that GH should be able to check the permissions given by the caller at runtime on access to actions that need these permissions. Technically they should be able to do that, without doing a validation statically.

Yeah, I explicitly tried that and it didn't seem to work. I set `package: write` on the call to the reusable workflow and passed the `GITHUB_TOKEN` as an input, but still got permissions errors. It worked when the re-usable workflow requested `packages: write` so it seems like the workflow needs to request all the permissions it needs to run in all cases, even if it uses only a subset.

The only solution I can think of is to make a separate workflow just for ghcr.io that requests `packages: write` but that feels like overkill...

PR #1266, opened 2022-11-25
ci: Ensure documentation is up-to-date with latest release

### LINKED ISSUE #1188: [feature] Builder ref in documentation should match the latest stable release

pnacht:
Closes #1188. This is a resubmission of #1193 after I tripped over myself on a simple rebase 🤦‍♂️.

- Adds pre-release test to ensure all documentation points to the new release tag
- If not, logs all cases in the format:
  `filename:line-number: [...]/action-name.yml@_wrong-tag_`
  i.e.
  `/internal/builders/go/README.md:205: [...]/builder_go_slsa3.yml@v1.2.0`
  This removes the clutter that comes from the entire `uses: ...` path, as well as adding the line numbers of all infractions
- Also adopts this format for the internal Actions test
- Change to .github/actions/README.md to conform with other placeholders (would otherwise fail the documentation check)
- Modifies RELEASE.md, moving documentation-update up to the "pre-release" stage, to align with new pre-release test

As-is, the documentation check only detects tags within a "workflow chunk" (that is, containing a `uses:` term).


pnacht:
Not sure why the ["pre-submit actions / references" check](https://github.com/slsa-framework/slsa-github-generator/actions/runs/3549577486/jobs/5962121074#step:4:6) is failing. It's detecting references to @v1.3.0, but I intentionally rebased to a commit prior to the v1.3.0 freeze, there aren't any @v1.3.0 references here...

For instance, it's complaining about builder_nodejs_slsa3.yml, saying it uses `rng@v1.3.0`, but... the version in my pre-freeze fork is using [`rng@main`](https://github.com/pnacht/slsa-github-generator/blob/df619699d488c0e07bfe10958c76d685fd4bf696/.github/workflows/builder_go_slsa3.yml#L79)

.github/workflows/scripts/pre-release/references.sh
```diff
@@ -32,3 +33,19 @@ if [[ "$results" != "" ]]; then
     exit 1
 fi
 
+# Verify documentation refers to the most recent release tag
+# TODO(https://github.com/slsa-framework/slsa-github-generator/issues/409):
+#   Include "./internal/builders/container/*" in this check
+results=$(
+    find . -name "*.md" -not -path ./internal/builders/container/* -print0 \
```
ianlewis, line 40: We're basically about the GA the container builder so you can just remove the exception for that builder.
```suggestion
    find . -name "*.md" -print0 \
```

RELEASE.md
```diff
@@ -55,6 +55,12 @@ Update version references with the following command:
 find .github/workflows/ -name '*.yaml' -o -name '*.yml' | xargs sed -i "s/uses: slsa-framework\/slsa-github-generator\/\.github\/actions\/\(.*\)@\(main\|v[0-9]\+\.[0-9]\+\.[0-9]\+\)/uses: slsa-framework\/slsa-github-generator\/.github\/actions\/\1@$BUILDER_TAG/"
 ```
 
+Likewise, update documentation with the following command:
+
+```shell
+find . -name "*.md" -exec sed -i "s/@v[0-9]\+\.[0-9]\+\.[0-9]\+/@$BUILDER_TAG/g" {} +
```
ianlewis, line 61: Sometimes we may refer to old versions in the docs. For example in sections on "Known issues": https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/go#known-issues

Is there a way we can maybe get around the check and update the right version in those cases? Before we only updated references to the slsa-framework/slsa-github-generator workflows rather than all references to any semantic version.

Maybe we should wrap the versions we don't want to update in pseudo tags?

pnacht: I could modify the regex to only update tags prefixed with `uses: slsa-framework/slsa-github-generator`. So basically only updating tags within example workflow blocks.

Or, if you believe there could be cases where we'd want to reference the current version outside of code blocks, we could instead check that tags are prefixed with `@`. This would still detect all workflows and then within the broader text we could prefix the "current version" tags with `@` ("New features in @v1.2.3"), while leaving references to old versions alone ("Known issues with v1.2.1"). (Or visa-versa)

Or, if we want to use some markdown hacks, we can prefix old versions with `[]()` (i.e. `"Known issues with []()v1.2.1"`, which renders as "Known issues with []()v1.2.1". We can then filter out versions with this "prefix". For improved clarity, we can also add a descriptive identifier to the prefix: `[](skip_update)` or something of the sort (though it'd have to be the same everywhere, since we'd have to add it to the filter).

ianlewis: I think for now just checking for the `uses: slsa-framework/slsa-github-generator` prefix would work.

BTW, another instance where we reference old versions in the docs in where we say which version a builder/generator was released as GA.
e.x. https://github.com/slsa-framework/slsa-github-generator#builders

ianlewis:
@pnacht Apologies for not looking at this sooner. RELEASE.md looks like it has conflicts. Can you resolve?

RELEASE.md
```diff
@@ -8,32 +8,32 @@ This is a document to describe the release process for the slsa-github-generator
...
-  - [Release tagging](#release-tagging)
-  - [Verify final version references](#verify-final-version-references)
-  - [Final adversarial tests](#final-adversarial-tests)
-  - [Reference Actions at main](#reference-actions-at-main)
-  - [Update documentation](#update-documentation)
-  - [Update verifier](#update-verifier)
-  - [Finish the release.](#finish-the-release)
-  - [Update the starter workflows](#update-the-starter-workflows)
-  - [Announce](#announce)
+- [How to cut a release](#how-to-cut-a-release)
```
ianlewis, line 11: nit: I don't think the TOC needs to include the doc title.

ianlewis (approved):

RELEASE.md
pnacht: Fixed. I have a Markdown linter that auto-updates the TOC, missed this.

PR #1283, opened 2022-11-29
Revert package perms

ianlewis:
Reopens #1257 

Reverts most of #1258 but leaves in place top level permissions for the generic container workflow. This is because cosign seems to use the ambient token to retrieve data from the OCI registry for signing rather than using the provided registry password.

### DIFF
--- a/.github/workflows/generator_container_slsa3.yml
+++ b/.github/workflows/generator_container_slsa3.yml
@@ -14,8 +14,6 @@
 
 name: SLSA container image provenance
 
-permissions: {}
-
 env:
   # Generator
   BUILDER_BINARY: slsa-generator-container-linux-amd64 # Name of the binary in the release assets.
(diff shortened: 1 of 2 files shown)


joshuagl:
> This is because cosign seems to use the ambient token to retrieve data from the OCI registry for signing rather than using the provided registry password.

I was going to suggest we file an issue, but see you've already done that. Thanks!
https://github.com/sigstore/cosign/issues/2489

joshuagl (approved):
LGTM

ianlewis:
> > This is because cosign seems to use the ambient token to retrieve data from the OCI registry for signing rather than using the provided registry password.
> 
> I was going to suggest we file an issue, but see you've already done that. Thanks! [sigstore/cosign#2489](https://github.com/sigstore/cosign/issues/2489)

I'm really not 100% sure what the cause is but the fact that the docker login and push actions succeed but the cosign upload of the attestation doesn't makes me think cosign is at fault.

laurentsimon:
Shall we wait for the cosign issue to be resolved before releasing the container workflow?

ianlewis:
> Shall we wait for the cosign issue to be resolved before releasing the container workflow?

I don't think it's that big of an issue that we need to hold up releasing the container workflow GA. Though we will want to fix it at some point. I'm tracking on the reopened #1257

PR #1285, opened 2022-11-29
Code freeze for v1.4.0-rc.1

ianlewis:
Code freeze

#label:release v1.4.0-rc.1

### DIFF
--- a/.github/workflows/builder_go_slsa3.yml
+++ b/.github/workflows/builder_go_slsa3.yml
@@ -76,7 +76,7 @@ jobs:
     steps:
       - name: Generate random 16-byte value (32-char hex encoded)
         id: rng
-        uses: slsa-framework/slsa-github-generator/.github/actions/rng@main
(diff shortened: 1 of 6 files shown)


ianlewis:
Somehow references.sh is getting a RELEASE_TAG value with a newline at the end. Not sure why...

joshuagl:
The `[:space:]` POSIX class in the RELEASE_TAG regular expression also matches line breaks, so I assume that's just matched from the input by the regular expression.

It might be worth changing the `[:space:]` to a `[:blank:]`? Or, alternatively, piping the output through a `tr -d "\n"`?

ianlewis:
> [:space:]

Yeah, this must be different for different versions of grep or something. I can't reproduce it locally and most descriptions I've read indicate that grep shouldn't ever match newlines. It's also strange because I'm pretty sure this code hasn't changed since I did the rc.0 release.

laurentsimon (approved):

PR #1286, opened 2022-11-29
docs: Add documentation on matrix-strategy builds

### LINKED ISSUE #707: [doc] Add example for workflows that use strategy / matrix
Here is an example https://github.com/google/sentencepiece/blob/master/.github/workflows/wheel.yml

pnacht:
Closes #707.

This PR adds documentation on how to adding provenance for matrix-strategy builds. It handles both the single-file and multi-file scenarios (where maintainers want a single file for all artifacts from all iterations or a different file for each iteration, respectively).

The documentation is stored within `./internal/builders/generic/README.md`, and a reference has been added to `./internal/builders/container/README.md` (this method has not been tested with the container generator, but should work there as well).


laurentsimon (approved):
Thanks. @ianlewis PTAL

PR #1318, opened 2022-12-06
migrate deprecated set-output to GITHUB_OUTPUT

datosh:
I recently migrated my pipelines to v1.4.1 and noticed there are some warnings refering to `set-output` left. I have found #1004 which makes it sounds these last references were missed during that cleanup.

In case this was not on purpose, here is a suggestion on how we could migrate to `GITHUB_OUTPUT`. 🙂 

The only reference I have left (that I could find) is: https://github.com/datosh/slsa-github-generator/blob/2c6620ba8f96507604e3e76364a19129fbf76cc2/github/set_output.go#L36
I left that in for @asraa once set-output is EOL, as stated by comment.

### DIFF
--- a/.github/actions/detect-workflow/main.go
+++ b/.github/actions/detect-workflow/main.go
@@ -157,6 +157,6 @@ func main() {
 	fmt.Printf("ref:%s\n", ref)
 
 	// Output of the Action.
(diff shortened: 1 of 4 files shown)


laurentsimon (approved):
Thanks, this was on our backlog!

PR #1321, opened 2022-12-07
refactor: Update env vars included in provenance.

ianlewis:
This updates the `CommandRunner` in two ways.

1. Copy env vars from the builder's environment to the command being run. This allows important env vars like `PATH` and `PWD` to be set properly for the environment.
1. Only include in provenance the env vars specified to the `CommandStep` or `CommandRunner`. Env vars copied from the builder's environment are *not* included in provenance for reproducibility. See #822.

Changes split out from #970 


laurentsimon (approved):

internal/runner/runner_test.go
```diff
@@ -152,3 +250,59 @@ func TestCommandRunner_RunnerMulti(t *testing.T) {
...
+			expected: []string{
+				"FOO=bar",
+				"HOGE=fuga",
+			},
+		},
+	}
+
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
+			sorted := cmpopts.SortSlices(func(a, b string) bool { return a < b })
```
ianlewis, line 302: I shouldn't sort slices when comparing here. `dedupEnv` should output the vars in the original order they were in the list, removing the previous values in the list if they are overwritten.

internal/runner/runner.go
```diff
@@ -140,9 +143,21 @@ func (r *CommandRunner) runStep(ctx context.Context, step *CommandStep, dry bool
...
 	}, nil
 }
+
+func dedupEnv(env []string) []string {
+	var deduped []string
+	seen := map[string]bool{}
+	for i := len(env) - 1; i >= 0; i-- {
+		k, _, found := strings.Cut(env[i], "=")
+		if found && !seen[k] {
+			deduped = append(deduped, env[i])
```
ianlewis, line 158: This results in a slice that is in reverse of the order of input. We should preserve the ordering.

internal/runner/runner.go
```diff
@@ -140,9 +143,21 @@ func (r *CommandRunner) runStep(ctx context.Context, step *CommandStep, dry bool
...
 		WorkingDir: pwd,
 	}, nil
 }
+
+func dedupEnv(env []string) []string {
+	var deduped []string
+	seen := map[string]bool{}
+	for i := len(env) - 1; i >= 0; i-- {
+		k, _, found := strings.Cut(env[i], "=")
+		if found && !seen[k] {
```
ianlewis, line 157: The `exec` package [leaves in malformed env var strings](https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/os/exec/exec.go;l=1247-1252;drc=888047c31007c0ad96764254fa8afb1417ddfd40). Should we do that too? or produce some kind of error?

ianlewis: Yeah, I think going with what they did is safest.

Issue #1327, opened 2022-12-07
[e2e]: generic tag main assets slsa3 


ianlewis:
Repo: https://github.com/slsa-framework/example-package/tree/v18.0.181
Run: https://github.com/slsa-framework/example-package/actions/runs/3635396734
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.generic.tag.main.assets.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.generic.tag.main.assets.slsa3.yml
Trigger: push
Branch: v18.0.181
Date: Wed Dec  7 02:56:16 UTC 2022

ianlewis:
Seems like `SUCCESS` for `upload-assets` is `false`

An error occurred in the "Download the provenance" step.
```
Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/example-package/example-package/__BUILDER_CHECKOUT_DIR__/.github/actions/secure-download-artifact'. Did you forget to run actions/checkout before running your local action?
```

ianlewis:
Looks like it is supposed to check out the builder repository but it checked out the `example-package` repo at https://github.com/slsa-framework/example-package/commit/d4d7666556cd3709eb8f537963256174f623aa5c

ianlewis:
Repo: https://github.com/slsa-framework/example-package/tree/v18.0.182
Run: https://github.com/slsa-framework/example-package/actions/runs/3644766673
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.generic.tag.main.assets.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.generic.tag.main.assets.slsa3.yml
Trigger: push
Branch: v18.0.182
Date: Thu Dec  8 02:51:30 UTC 2022

asraa:
I find it REALLY weird that it's only for this one. You'd expect all jobs to fail then. See https://github.com/slsa-framework/example-package/actions/runs/3647348725/jobs/6159481273 for a recent run where the output is set correctly (and the main at the time includes that PR)

asraa:
Actually, you can see in both the failed runs above that the outputs were set in Complete Job https://github.com/slsa-framework/example-package/actions/runs/3635396734/jobs/6134470562

```
Set output 'outcome'
Set output 'repository'
[...]

PR #1331, opened 2022-12-08
fix(generic): fix bug in generic workflow to set pre-requisite job

asraa:

Fixes https://github.com/slsa-framework/slsa-github-generator/issues/1327

It seemed like the `repository` and `ref` were not set during the upload-assets job, and because of this, example-package was checked out instead of the builder repository. This was because that job did not require `detect-env` where the `repository` and `ref` come from, so they were probably resolved before that job completed.


laurentsimon (approved):
Thanks. We need to cut a release for the generators then?

asraa:
> Thanks. We need to cut a release for the generators then?

My feeling is probably yes @ianlewis. I think this problem is somewhat intermittent and so would likely be confusing for users.

laurentsimon:
+1, let's cut a release.

Issue #599, opened 2022-07-21
[feature] Share errors across builders


laurentsimon:
We use WrappableError in the generic builder, but not in the Go builder.

We should use the same construction and error definitions across builders

PR #1383, opened 2022-12-13
refactor: Migrate go builder to WrappableError

datosh:
Worked on this after discovering https://github.com/slsa-framework/slsa-github-generator/issues/599

I added two questions (as TODO prefixed code comments) where I was not sure which path is the better. Do you have any guidance?


internal/builders/go/main_test.go
```diff
@@ -237,35 +237,36 @@ func Test_runBuild(t *testing.T) {
 			workingDir: "./valid/path/",
 		},
 		// Below are the same tests we do in pkg/config_test.go
+		// TODO: Do they provide additional value then? Can we remove them instead of exporting errors?
```
laurentsimon, line 240: good point. I think they are used in main_test.go, and it's the only reason they are public. A while back, we discussed having all the files in the same folder, in which case we could make them private.

internal/builders/go/pkg/config.go
```diff
@@ -62,6 +53,22 @@ type GoReleaserConfig struct {
 	Binary  string
 }
 
+// ErrUnsupportedVersion indicates an unsupported GoReleaser version.
```
laurentsimon, line 56: note: this is not GoReleaser, it's our own Go builder

internal/builders/go/pkg/config.go
```diff
@@ -158,7 +164,7 @@ func convertPathError(e error, msg string) error {
 func validateVersion(cf *goReleaserConfigFile) error {
 	_, exists := supportedVersions[cf.Version]
 	if !exists {
-		return fmt.Errorf("%w:%d", ErrorUnsupportedVersion, cf.Version)
+		return errors.Errorf(&ErrUnsupportedVersion{}, "version '%d' not supported", cf.Version)
```
laurentsimon, line 167: I'm not super familiar with wrappable error. But how do we make sure the error message remains consistent across the codebase?

laurentsimon (commented):
Thanks for the PR. Very much appreciated!
@ianlewis PTAL

internal/builders/go/main_test.go
ianlewis: The only difference I can tell is that the config is passed through `runBuild` in `main_test.go`  and `ConfigFromFile` (which is called by `runBuild`) in `config_test.go`. Maybe these tests in `main_test.go` are the broader tests?

I think it's fine to leave this as is for now and tackle it in a later refactor.

ianlewis (commented):
Thanks for taking this on!

datosh:
I have reworked the tests as discussed @ianlewis based on the structure in `attestation_test.go`.

I have defined the functions to check errors outside of the tests to keep the cognitive complexity value of the checker low. Also this allowed to reuse the same functions in `build_test.go`.

ianlewis (approved):

PR #1389, opened 2022-12-16
[docker] update docker-based workflow and add pre-submit workflow

asraa:

This updates the docker-based workflow to be a working with placeholder outputs. Adds a pre-submit job that checks for workflow correctness right now (but performs no verification yet).

The builder output and custom build definition are hard-coded, and the predicate/attestation/envelope generation is placeholder. 

The placeholder actions will be added in separate PRs with tests. These will also be re-used in the delegator workflow. Once the delegator workflow is done, I will also test using the completed builder binary in the delegator.

@rbehjati

### DIFF
--- a/.github/workflows/builder_docker-based_slsa3.yml
+++ b/.github/workflows/builder_docker-based_slsa3.yml
@@ -345,7 +345,7 @@ jobs:
           mkdir attestations
           echo "{}" > attestations/attestation1.json
           echo "{}" > attestations/attestation2.json
-          echo "output-folder=attestations/" >> $GITHUB_OUTPUT
+          echo "output-folder=attestations" >> $GITHUB_OUTPUT
 
       - name: Upload unsigned intoto attestations file for pull request
         if: ${{ github.event_name == 'pull_request' }}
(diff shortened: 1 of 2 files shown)


laurentsimon (approved):

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -104,95 +119,265 @@ jobs:
...
+          echo "run resolution and output buildDefinition!"
+
+          SOURCE_ALG=${SOURCE_DIGEST%%:*}
+          SOURCE_DIGEST=${SOURCE_DIGEST#*:}
+
+          IMAGE_ALG=${BUILDER_DIGEST%%:*}
+          IMAGE_DIGEST=${BUILDER_DIGEST#*:}
+
+          cat <<EOF >DATA
+          {
```
rbehjati, line 186: The JSON format is not quite right, I think. The one I get from the Go structs is https://github.com/slsa-framework/slsa-github-generator/blob/c281078c3a9fe1d76451cca2f3605182d915d192/internal/builders/docker/testdata/build-definition.json.

asraa: Oops, I think ou're right: the source/builderImage needed to be inlined in artifacts. I'll send a quick correction PR

PR #1395, opened 2022-12-20
ci: Enable and fix lll

datosh:
Updates #450 

### DIFF
--- a/internal/builders/docker/pkg/common.go
+++ b/internal/builders/docker/pkg/common.go
@@ -85,6 +85,7 @@ type ArtifactReference struct {
 	// OPTIONAL.
 	LocalName string `json:"localName,omitempty"`
 
+	//nolint:lll
 	// [URI] identifying the location that this artifact was downloaded from, if
 	// different and not derivable from `uri`.
 	//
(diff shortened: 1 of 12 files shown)


internal/builders/docker/pkg/common.go
```diff
@@ -85,6 +85,7 @@ type ArtifactReference struct {
 	// OPTIONAL.
 	LocalName string `json:"localName,omitempty"`
 
+	//nolint:lll
```
datosh, line 88: I guess it makes no sense to shorten the link (https://files.pythonhosted.org/packages/36/2b/61d51a2c4f25ef062ae3f74576b01638bebad5e045f747ff12643df63844/PyYAML-6.0.tar.gz). Also putting it on its own line is not sufficient. Breaking up the link into multiple lines breaks the link, as in, you cannot follow the link when viewing the docs.

ianlewis (approved):

internal/builders/docker/pkg/common.go
ianlewis: Yeah, I agree. Thanks!

PR #1397, opened 2022-12-20
Add attestation generation action from layout and predicate

asraa:

cc @rbehjati this action takes the build's subjects output and combines with a predicate to turn it into attestations to sign


.github/actions/generate-attestations/__tests__/attestation.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    fs.readFileSync.mockReturnValueOnce(JSON.stringify(predicate));
+
+    const attestations: Record<string, string> =
+      await statement.writeAttestations(
+        "layoutMock",
+        predicateType,
+        "predicateMock"
+      );
+    expect(Object.keys(attestations)).toEqual(
+      expect.arrayContaining(["attestation1.intoto", "attestation2.intoto"])
```
laurentsimon, line 74: do you need to compare the predicate type?

.github/actions/generate-attestations/src/main.ts
```diff
@@ -0,0 +1,58 @@
+import * as core from "@actions/core";
+import fs from "fs";
+import path from "path";
+import { writeAttestations } from "./attestation";
+
+// Detect directory traversal for input file.
+function resolvePathInput(input: string, wd: string): string {
+  const safeJoin = path.join(wd, input);
+  if (!safeJoin.startsWith(wd)) {
```
laurentsimon, line 9: I think you need to normalize the wd to have a trailing `/` before calling startsWith`, otherwise:
`wd: /some/path`
`input: ../path-other/afile`
`safeJoin = /some/path-other/afile` would pass the verification

.github/actions/generate-attestations/src/main.ts
```diff
@@ -0,0 +1,58 @@
...
+    const attestations = await writeAttestations(
+      safeSlsaOutputs,
+      predicateType,
+      safePredicateFile
+    );
+
+    // Write attestations
+    fs.mkdirSync(outputFolder, { recursive: true });
+    for (const att in attestations) {
+      const outputFile = `${outputFolder}/${att}`;
```
laurentsimon, line 44: use `path.join()` and validate for path traversal?

.github/actions/generate-attestations/src/attestation.ts
```diff
@@ -0,0 +1,60 @@
...
+        JSON.stringify(layout[att])
+      );
+
+      const attestationJSON = createStatement(
+        subjectJson,
+        predicateType,
+        predicateJson
+      );
+
+      ret[att] = JSON.stringify(attestationJSON, null, "\t");
```
laurentsimon, line 56: curious if `\t` is needed. Do we need pretty-print?

.github/workflows/pre-submit.actions.yml
```diff
@@ -299,3 +299,52 @@ jobs:
...
+            "buildType": "https://github.com/slsa-framework/slsa-github-generator/go@v1",
+            "builder": {
+              "id": "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_go_slsa3.yml@refs/tags/v0.0.1"
+            }
+          }
+          EOF
+          cat PREDICATE > predicate.json
+      - name: Test generate attestations
+        id: generate
+        uses: asraa/slsa-github-generator/.github/actions/generate-attestations@attestation-action
```
laurentsimon, line 338: update asraa to this repo.

.github/workflows/pre-submit.actions.yml
```diff
@@ -299,3 +299,52 @@ jobs:
...
+          slsa-outputs-file: slsa-outputs.json
+          predicate-type: "https://slsa.dev/provenance/v0.2"
+          predicate-file: predicate.json
+      - name: Get outputs
+        env:
+          OUTPUT_FOLDER: ${{ steps.generate.outputs.output-folder }}
+        run: |
+          set -euo pipefail
+          ls $OUTPUT_FOLDER
+          cat $OUTPUT_FOLDER/attestation1.intoto | jq -r
```
laurentsimon, line 349: want to verify the name and digest?

laurentsimon (approved):

.github/actions/generate-attestations/__tests__/attestation.test.ts
```diff
@@ -0,0 +1,77 @@
+const fs = require("fs");
+const statement = require("../src/attestation");
```
laurentsimon, line 2: since you added unit tests, maybe we can add these as pre-submits.

.github/workflows/pre-submit.actions.yml
asraa: Yes! Will do before submitting. Without this though, the pre-submits wouldn't run correct (because the action isn't checked into main -- just wanted to validate that it will go through)

.github/actions/generate-attestations/src/attestation.ts
```diff
@@ -0,0 +1,60 @@
...
+      if (ret[att] !== undefined) {
+        throw Error(`duplicate attestation file ${att}`);
+      }
+
+      // TODO: How to cast directly into types.Subject[]?
+      const subjectJson: types.Subject[] = JSON.parse(
+        JSON.stringify(layout[att])
+      );
+
+      const attestationJSON = createStatement(
```
laurentsimon, line 50: IIUC, the delegated tool is in charge of setting env variables and other parameters from the GH run, correct?

asraa: Yes! The predicate would be created in a prior step of a trusted job -- but separating this allows me to modularize the predicate creation

(e.g. create v0.2, v1.0)

laurentsimon: OK. Will we create some guidance about *what* and *how* to record? So that we can be consistent with our own builders?

asraa: Hmm good point - yes. I could do some predicate validation on some of these fields. You mentioned predicate type checking earlier -- I wasn't sure if you meant some validation that the prediateType input matches the predicate format defined. I thought about adding that in, and if so, I could do make a `validatePredicate(type, predicate)` that checks (1) if the predicate conforms to the Type and (2) if the predicate contains certain expected fields (context, runner material, some source material, a builder ID, to start?)

.github/actions/generate-attestations/__tests__/attestation.test.ts
asraa: For testing, the createStatement tests assert on predicate type for the output attestations. this test just tests the filenames. I will add some more though -- there's some conditions I expect on the filenames produced (not in nested, no repeats)

.github/actions/generate-attestations/src/main.ts
asraa: adding tests, since this wasn't clear to me

.github/actions/generate-attestations/src/attestation.ts
laurentsimon: Sg. Maybe just create an issue for now.

.github/actions/generate-attestations/src/intoto.ts
```diff
@@ -0,0 +1,13 @@
+export const INTOTO_TYPE = "https://in-toto.io/Statement/v0.1";
+
+export type Subject = {
```
rbehjati, line 3: Thanks for the PR @asraa. I will generate a json file containing an array of subjects of this type as the output of the `build` command, in the Docker builder.

.github/actions/generate-attestations/src/attestation.ts
asraa: removed

.github/actions/generate-attestations/src/main.ts
asraa: done! factored the path validation out into a function

.github/workflows/pre-submit.actions.yml
asraa: done

.github/actions/generate-attestations/src/attestation.ts
asraa: linked the issue (https://github.com/slsa-framework/slsa-github-generator/issues/1422) in a TODO

.github/actions/generate-attestations/__tests__/attestation.test.ts
asraa: done! updated contirbuting docs to mention jest (node.js test framework) and added to presubmits and Makefile

asraa:
@laurentsimon anything else here? I have to craft some extra malicious SLSA subject layouts (build outputs)
https://github.com/slsa-framework/slsa-github-generator/issues/1423
and update with predicate validation when I have some predicate-aware code
https://github.com/slsa-framework/slsa-github-generator/issues/1422

.github/actions/generate-attestations/src/main.ts
laurentsimon: fyi, this is the similar Go function we use in case you want to compare https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/utils/path_test.go#L27

.github/actions/generate-attestations/src/attestation.ts
```diff
@@ -0,0 +1,67 @@
...
+): types.IntotoStatement {
+  return {
+    _type: types.INTOTO_TYPE,
+    subject: subjects,
+    predicateType: type,
+    predicate,
+  };
+}
+
+export async function writeAttestations(
```
laurentsimon, line 21: curious if `async` is needed or not.

.github/actions/generate-attestations/src/main.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    // Write attestations
+    fs.mkdirSync(outputFolder, { recursive: true });
+    for (const att in attestations) {
+      const outputFile = path.join(outputFolder, att);
+      const safeOutput = resolvePathInput(outputFile, wd);
+      fs.writeFileSync(safeOutput, attestations[att]);
+    }
+
+    core.setOutput("output-folder", outputFolder);
```
laurentsimon, line 41: curious why this is needed if it's the same as the `output-folder` input

.github/actions/generate-attestations/src/main.ts
```diff
@@ -0,0 +1,51 @@
...
+  } catch (error) {
+    if (error instanceof Error) {
+      core.setFailed(error.message);
+    } else {
+      core.info(`Unexpected error: ${error}`);
```
laurentsimon, line 46: change to `core.setFailed()`.

.github/workflows/pre-submit.actions.yml
```diff
@@ -319,3 +320,79 @@ jobs:
           directory: "internal/builders/generic"
           # NOTE: compile-builder explicitly set to false.
           compile-builder: false
+
+  generate-attestations:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create test inputs
```
laurentsimon, line 327: you could store these results in a file committed to the repo

.github/workflows/pre-submit.actions.yml
```diff
@@ -319,3 +320,79 @@ jobs:
...
+          cat <<EOF >PREDICATE
+          {
+            "buildType": "https://github.com/slsa-framework/slsa-github-generator/go@v1",
+            "builder": {
+              "id": "https://github.com/slsa-framework/slsa-github-generator/.github/workflows/builder_go_slsa3.yml@refs/tags/v0.0.1"
+            }
+          }
+          EOF
+          cat PREDICATE > predicate.json
+      # TODO(asraa): Enable generate-attestation action at main
```
laurentsimon, line 365: create a tracking issue if that helps not forget

CONTRIBUTING.md
```diff
@@ -105,6 +109,12 @@ unit tests locally using `make`. This requires that the Go runtime be installed.
 make unit-test
 ```
 
+Tests for GitHub Actions created with Typescript require Jest.
+
```
laurentsimon, line 113: want to add a link to their official repo?

laurentsimon (approved):

.github/actions/generate-attestations/src/main.ts
asraa: output-folder is an optional input with a default, so that callers would know the resolved one as output. It's not necessary though, since the caller would know what it used as input/default.

.github/actions/generate-attestations/action.yml
```diff
@@ -0,0 +1,22 @@
...
+  predicate-type:
+    description: 'A URI dedining the type of the predicate, for e.g. https://slsa.dev/provenance/v0.2'
+    required: true
+  predicate-file:
+    description: 'A JSON file describing the SLSA predicate to attach to the subjects'
+    required: true
+  output-folder:
+    description: 'Output folder to place attestations'
+    required: false
+    default: 'attestations'
```
laurentsimon, line 16: how will this work if the Action is called several times, e.g. if there are multiple `builder_labels` running in parallel or if the PW has calls the TRW multiple times in the same workflow? The name needs to be unique. Would it be better to remove this input and only provide the output, which this Action would randomize `attestations-xxxxxx`?
Maybe this will be done by the SRW which we control and we will re-use the rng Action for it? If that's the case, shall we mark the input as `required`, since the default is not safe?

asraa: > Maybe this will be done by the SRW which we control and we will re-use the rng Action for it? If that's the case, shall we mark the input as required, since the default is not safe?

Yes, this will be done by the SRW. Something like here during the trusted attestation creation/signing: https://github.com/laurentsimon/slsa-delegator/blob/main/.github/workflows/delegator_generic_slsa3.yml#L237-L244

Ah, sg, let's mark it as required. I agree it was confusing to have a default anyway. Having a required input also means that I no longer need to set an output.

.github/actions/generate-attestations/src/main.ts
asraa: resolved below, making it a required one, since it will likely need to be randomized to avoid multiple calls to an SRW

.github/actions/generate-attestations/action.yml
laurentsimon: SG.

laurentsimon:
no further comment from me

.github/workflows/pre-submit.actions.yml
asraa: done! i'll add more malicious layouts and predicate validations in these folders anyway.

.github/workflows/pre-submit.actions.yml
asraa: i removed, realized i can checkout and invoke as a local action :P

PR #1399, opened 2022-12-21
Enable and fix govet

datosh:
Updates #450 

As discussed in #1384 this is mostly changes to order of fields to (allegedly) improve garbage collection performance.  The idea is that all pointers are moved to the top of the struct so that GC can quit checking the struct as early as possible. I have not used this specific check before. So I am not aware how big the performance gain is. I don't think we have any benchmarks I could use to verify?

All these checks are enabled via `fieldalignment` so we could also decide to opt out of these specific checks in `govet`. I don't consider this project to be that performance critical, so it might make sense to not have the check and opt for clearer code, since finding the right order of elements requires some knowledge about how certain golang build-ins are implemented, e.g., slice, map, string, ...

Let me know what you think!


ianlewis (approved):

ianlewis:
@datosh A couple of conflicts were added. Could you fix those?

datosh:
> @datosh A couple of conflicts were added. Could you fix those?

Done :slightly_smiling_face: 

The "verify base images" step seems to be failing, but it does not seem to be specific to this PR.

ianlewis:
> > @datosh A couple of conflicts were added. Could you fix those?
> 
> Done 🙂
> 
> The "verify base images" step is failing, but it does not seem to be specific to this PR.

Yes we've made it non-blocking for now. See #1401

PR #1400, opened 2022-12-21
ci: Enable and fix gosec

datosh:
Updates #450 

As previously discussed in #1384 this is triggered by G601. More detailed background in this [SO post](https://stackoverflow.com/questions/62446118/implicit-memory-aliasing-in-for-loop).


internal/builders/go/pkg/build_test.go
```diff
@@ -1026,6 +1026,7 @@ func Test_generateCommand(t *testing.T) {
 			}
 
 			for _, cfg := range cfgs {
+				cfg := cfg
```
ianlewis, line 1029: I think that if the slice definition (at line 1016) is for a pointer then we can avoid this issue.

i.e.
```
cfgs := []*goReleaserConfigFile{
    ...
}
```

then line 1029 can just be

```golang
c, err := fromConfig(cfg)
```

ianlewis (approved):
Comment is nit so I'll approve and merge. Can update the test if we get to it.

ianlewis:
Looks like there are some new gosec errors.

@datosh Could you take a look?

internal/builders/go/pkg/build_test.go
datosh: Yes, good point!

datosh:
> Looks like there are some new gosec errors.

Yes, the new errors were related to the usage of `os/exec`. `gosec` states that ["G204 is really intended for assisting in code audits. It doesn't necessarily indicate a bad behaviour, but something that should be checked for potential command execution."](https://github.com/securego/gosec/issues/292#issuecomment-480089941)

I have checked that the provided input is from user config, therefore a trusted and controlled value.

An alternative would be to disable G204 for the whole project, if you think the workflow of vetting each finding by hand is not something this project needs/wants. Let me know what you think, @ianlewis

ianlewis:
> An alternative would be to disable G204 for the whole project, if you think the workflow of vetting each finding by hand is not something this project needs/wants. Let me know what you think, @ianlewis

Yeah, gosec seems to air heavily on the side of caution and complain about a lot of stuff that's not necessarily an issue. In this case I guess every call to `exec.Command` gets marked? Seems pretty arbitrary. It also seems to be fine with `exec.CommandContext` and doesn't complain about that. 🤷
https://github.com/slsa-framework/slsa-github-generator/blob/43936fefa3f76d28126da888dbf4736ea3f908c0/internal/runner/runner.go#L106

ianlewis (approved):

PR #1403, opened 2022-12-21
feat: enable and fix gocritic

datosh:
Updates #450 


internal/builders/go/pkg/build.go
```diff
@@ -229,7 +229,8 @@ func (b *GoBuild) getDir() (string, error) {
 }
 
 func (b *GoBuild) generateCommand(flags []string, binary string) []string {
-	command := append(flags, []string{"-o", binary}...)
+	command := flags
```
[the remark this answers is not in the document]
datosh: Makes sense! Thanks. 👍

ianlewis (approved):

internal/builders/docker/pkg/builder.go
```diff
@@ -69,15 +69,15 @@ type Builder struct {
...
+func NewBuilderWithGitFetcher(config *DockerBuildConfig, forceCheckout bool) (*Builder, error) {
+	gc, err := newGitClient(config, forceCheckout, 0 /* depth */)
 	if err != nil {
 		return nil, fmt.Errorf("could not create builder: %v", err)
 	}
 
 	return &Builder{
 		repoFetcher: gc,
-		config:      config,
+		config:      *config,
```
ianlewis, line 80: Should the `Builder.config` be a pointer too?

datosh: `gocritic` suggested to pass by pointer to copy less data. I think `Builder.config` is fine as value so that `Builder` can own its configuration? I think you can judge this better than I can.

ianlewis (approved):

PR #1404, opened 2022-12-21
Enable and fix errcheck

datosh:
Updates #450 


ianlewis:
> cognitive complexity 31 of func `Test_runBuild` is high (> 30) (gocognit)

https://github.com/slsa-framework/slsa-github-generator/pull/1404/files#diff-a4bc4326608ea5a0b21c45e77330c3dd26ab06449babf86c3601ac07b9d9c150

I'm not a big fan of the way that gocognit and others calculate "cognitive completity". Statements in an indented block get multiplied per indentation so every statement even just two blocks in are 3 points per statement, so you can write at most 10 lines of code or so.

```golang
func TestFoo(t *testing.T) {
    t.Run("foo", func(t *testing.T) {
        // every statement here is 3 points.
    })
}
```

This makes it really hard to write even fairly benign code like table-driven tests. Their default is 30, which by itself is pretty restrictive, but [golangci-lint recommends between 10-20](https://golangci-lint.run/usage/linters/#gocognit), which I think is pretty impractical.

Maybe we can increase our max complexity in .golangci-lint.yml?

```yaml
linters-settings:
  gocognit:
    # Minimal code complexity to report
    # Increased beyond default to support table-driven tests.
    min-complexity: 35
```

@laurentsimon Any opinions here?

ianlewis:
@datosh Can you go ahead and increase the `min-complexity` to 35?

```
linters-settings:
  gocognit:
    # Minimal code complexity to report
    # Increased beyond default to support table-driven tests.
    min-complexity: 35
```

ianlewis (approved):

PR #1420, opened 2022-12-27
feat: Add verify-token Action

laurentsimon:


laurentsimon:
note: setup-token pre-submit is being fixed in https://github.com/slsa-framework/slsa-github-generator/pull/1417 and should be merged before this PR

.github/actions/verify-token/src/main.ts
```diff
@@ -0,0 +1,284 @@
...
+    // Verify the intended recipient.
+    validateField(
+      "builder.audience",
+      rawTokenObj.builder.audience,
+      workflowRecipient
+    );
+
+    // Verify that the runner label is not empty.
+    validateNonEmptyField(
+      "builder.runner_label",
```
laurentsimon, line 120: we could also verify that it's one of the supported labels. Currently we only support `ubuntu-latest`. Shall we do a verification here?

.github/actions/verify-token/src/main.ts
```diff
@@ -0,0 +1,284 @@
...
+      throw new Error(`malformed token: ${parts.length}`);
+    }
+
+    const bundleStr = Buffer.from(parts[0], "base64").toString("utf-8");
+    const b64Token = parts[1];
+    const bundle = JSON.parse(bundleStr);
+
+    // First, verify the signature, i.e., that it is signed by a certificate that
+    // chains up to Fulcio.
+    await sigstore.sigstore.verify(bundle, Buffer.from(b64Token));
```
asraa, line 74: so the payload signed is the base64-encoded token? why not the rawToken?

.github/actions/verify-token/src/main.ts
asraa: sgtm

asraa (approved):
nice!

.github/actions/verify-token/src/main.ts
laurentsimon: done

.github/actions/verify-token/src/main.ts
laurentsimon: 1. less parsing before verification, so less attack surface
2. base64 decoding allows some bits to be flipped without changing the decoded results, which seems weird for signature verification to still pass

Not a big deal, I agree :)

PR #1434, opened 2023-01-03
feat: add attestation signing action

asraa:

This adds an attestation signing action that uses Sigstore signing to sign a folder of attestations with a given `payloadType`. It outputs them in a given `output-folder` (which will be invoked in trusted workflows with an RNG suffix). 

This action is trusted, in that it is used by the trusted workflows. The attestation & output folder is assumed to be in a trusted path.

Testing: Added a schedule/workflow_dispatch action that runs sigstore-js verify on the outputs. When slsa-verifier implements verification, we can use that to test the verification. A successful test run is here: https://github.com/asraa/slsa-github-generator/actions/runs/3832784934/jobs/6523496029


.github/actions/sign-attestations/src/main.ts
```diff
@@ -0,0 +1,53 @@
...
+    // Output folder
+    const outputFolder = core.getInput("output-folder");
+
+    const files = await fs.promises.readdir(attestationFolder);
+    fs.mkdirSync(outputFolder, { recursive: true });
+    for (const file of files) {
+      const fpath = path.join(attestationFolder, file);
+      const stat = await fs.promises.stat(fpath);
+      if (stat.isFile()) {
+        console.log("Signing '%s'.", fpath);
```
laurentsimon, line 27: bit: use `core.debug()/info()` instead of `console.log`?

.github/actions/sign-attestations/src/main.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+async function run(): Promise<void> {
+  try {
+    // Attestations
+    const attestationFolder = core.getInput("attestations");
+    const payloadType = core.getInput("payload-type");
+
+    // Output folder
+    const outputFolder = core.getInput("output-folder");
```
laurentsimon, line 19: output-folder is trusted. If not sure, we can addd a verification for dir traversal.

.github/actions/sign-attestations/src/main.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    const files = await fs.promises.readdir(attestationFolder);
+    fs.mkdirSync(outputFolder, { recursive: true });
+    for (const file of files) {
+      const fpath = path.join(attestationFolder, file);
```
laurentsimon, line 24: do we verify the `file` fr dir traversal?

laurentsimon (approved):
Nice!

.github/actions/sign-attestations/src/main.ts
```diff
@@ -0,0 +1,53 @@
...
+          payloadType,
+          signOptions
+        );
+        const bundleStr = JSON.stringify(bundle);
+        const outputPath = path.join(
+          outputFolder,
+          path.basename(file),
+          ".sigstore"
+        );
+        fs.writeFileSync(outputPath, bundleStr);
```
laurentsimon, line 40: nit: update the options like in https://github.com/slsa-framework/slsa-github-generator/pull/1437/files#diff-75682bfbd9f1896f11129f0cf37ca0ddd081e2bf109c221bf30fef94b0098664

.github/actions/sign-attestations/src/main.ts
asraa: Yes, it should be provided by the trusted workflow. Either way, I'll validate. Part of why I didn't add that in was because it required c/p the util. 

I'll file an issue about reducing code-dupe between GHA

PR #1437, opened 2023-01-03
feat: update file creation flags

laurentsimon:
Mirror what we do for Go https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/utils/path.go#L84

In particular, always fail if the file already exists, to ensure an attacker cannot overwrite existing files.
(May not be needed for the use case of generate-attestations, but good practice + the code will be copied / pasted in other places)


joshuagl (approved):
Seems like a sensible change 👍

asraa (approved):
Thank you!

PR #1466, opened 2023-01-04
fix: fix sign-attestation paths

asraa:

There was an error causing the path to be doubly-absolutely-resolved. Fixes so that the path is still verified for path traversal, but is not $(wd)/$(wd)/$fpath.


laurentsimon (approved):

.github/actions/sign-attestations/dist/index.js
```diff
@@ -37119,7 +37119,7 @@ function run() {
             fs.mkdirSync(safeOutputFolder, { recursive: true });
             const files = yield fs.promises.readdir(safeAttestationFolder);
             for (const file of files) {
-                const fpath = resolvePathInput(path.join(safeAttestationFolder, file), wd);
+                const fpath = resolvePathInput(path.join(attestationFolder, file), wd);
```
ianlewis, line 37122: nit: Since we've already resolved the attestation folder (and checked it isn't outside the working directory), isn't it safe to just do `path.join(safeAttestationFolder, file)`?
```suggestion
                const fpath = path.join(safeAttestationFolder, file);
```

ianlewis (approved):

.github/actions/sign-attestations/dist/index.js
laurentsimon: `file` is already validated, I think. But we could also verify the entire file path `path.join(attestationFolder, file)`. It could be hard in practice for a reviewer to track back everything that's going on and the assumptions about where validation took place.

asraa: > it safe to just do path.join(safeAttestationFolder, file)?

Yeah, I figured it might be. I wasn't totally sure whether there may be `file` names in the folder that had weird characters (somehow) though.

PR #1468, opened 2023-01-04
feat: add pre-submit for verify-token Action

laurentsimon:


.github/workflows/pre-submit.actions.yml
```diff
@@ -115,6 +115,50 @@ jobs:
...
+          slsa-unverified-token: ${{ steps.token.outputs.unverified }}
+          slsa-workflow-recipient: "elegator_generic_slsa3.yml"
+      - id: verify-mismatch-token
+        uses: ./.github/actions/verify-token
+        continue-on-error: true
+        with:
+          slsa-unverified-token: aGVsbG8K
+          slsa-workflow-recipient: "delegator_generic_slsa3.yml"
+      - env:
+          SUCCESS: ${{ steps.verify.outcome == 'failure' && steps.verify-mismatch-recipient.outcome == 'failure' && steps.verify-mismatch-token.outcome == 'failure' }}
```
ianlewis, line 157: These are all failure scenarios. Maybe add a successful test with the output of the `setup-token` test above?

.github/workflows/pre-submit.actions.yml
```diff
@@ -115,6 +115,50 @@ jobs:
...
+          TOOL_REF: "refs/heads/main"
+          TOOL_URI: "https://github.com/laurentsimon/slsa-delegated-tool/.github/workflows/tool1_slsa3.yml@refs/heads/main"
+      - id: verify
+        uses: ./.github/actions/verify-token
+        continue-on-error: true
+        with:
+          # Always fails due to the GitHub run mismatch.
+          slsa-unverified-token: ${{ steps.token.outputs.unverified }}
+          slsa-workflow-recipient: "delegator_generic_slsa3.yml"
+      - id: verify-mismatch-recipient
```
ianlewis, line 144: Is this just going to fail anyway due to the run mismatch? are we actually testing what we think we're testing?

.github/workflows/pre-submit.actions.yml
```diff
@@ -115,6 +115,50 @@ jobs:
...
+        uses: ./.github/actions/verify-token
+        continue-on-error: true
+        with:
+          slsa-unverified-token: ${{ steps.token.outputs.unverified }}
+          slsa-workflow-recipient: "elegator_generic_slsa3.yml"
+      - id: verify-mismatch-token
```
ianlewis, line 150: Is this testing a mismatch or just an invalid token?

.github/workflows/pre-submit.actions.yml
asraa: could you maybe leverage a core.debug/core.setFailed somehow (maybe by running it as a node script instead)?

.github/workflows/pre-submit.actions.yml
laurentsimon: I copied the tests from the schedule workflow. The verification will always fail no matter what because the token was generated *outside* this workflow, but the verify-token can only succeed if the token is generated in the same 'run'.

Maybe those pre-submits are not useful :/

.github/workflows/pre-submit.actions.yml
laurentsimon: I just copied from the schedule workflow. Agreed it's not very useful and maybe we should just rely on the scheduled runs... or do what Asra is suggesting. At this point maybe we should instead use unit test in the TS code itself?

.github/workflows/pre-submit.actions.yml
laurentsimon: cannot use setup-token in pre-submit, because it does not have the permission for signing (`id-token: write`). The end-to-end test is done in the scheduled workflow run [(already present in the codebase)](https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/e2e.verify-token.schedule.yml)

laurentsimon:
Based on the comment, it may be better to invest in unit test in TS code. Wdut?

.github/workflows/pre-submit.actions.yml
asraa: Yeah unit tests would be great if the code is modularized from signature checking to the policy.

I suggested a presubmit though  because this action was missing the action.yml! So long as *somewhere* we are running the action as an action in a test, we'd catch this. So maybe it's OK, since the actions will eventually be used in a workflow.

asraa:
> Based on the comment, it may be better to invest in unit test in TS code. Wdut?

I think this is OK -- let's do this. If there are action syntax errors somewhere we'll eventually catch them in post-submits or e2e tests even if there's a lag.

.github/workflows/pre-submit.actions.yml
laurentsimon: mhhh, a pre-submit will always fail verification. I need to differentiate between errors, but I don't know if I can read the logs / error messages. Running as a node application does not require the action.yml, so we would not catch the problem.

Any other ideas?

.github/workflows/pre-submit.actions.yml
```diff
@@ -115,6 +115,50 @@ jobs:
...
+        with:
+          # Always fails due to the GitHub run mismatch.
+          slsa-unverified-token: ${{ steps.token.outputs.unverified }}
+          slsa-workflow-recipient: "delegator_generic_slsa3.yml"
+      - id: verify-mismatch-recipient
+        uses: ./.github/actions/verify-token
+        continue-on-error: true
+        with:
+          slsa-unverified-token: ${{ steps.token.outputs.unverified }}
+          slsa-workflow-recipient: "elegator_generic_slsa3.yml"
```
ianlewis, line 149:
```suggestion
          slsa-workflow-recipient: "delegator_generic_slsa3.yml"
```

.github/workflows/pre-submit.actions.yml
ianlewis: Are things like the run ID things we can fake by just setting the right GITHUB_* environment variables? I think that probably only solves one of the issues so won't get us all the way there anyway.

otherwise yeah, maybe unit tests are a better way to test the `verify-token` action.

.github/workflows/pre-submit.actions.yml
ianlewis: yeah, It's going to be hard to test this properly without `id-token: write`.

.github/workflows/pre-submit.actions.yml
laurentsimon: this was intended

.github/workflows/pre-submit.actions.yml
laurentsimon: we cannot overwrite the `GITHUB_*` variables, by default. The runner does not allow it, unless we patch libc

laurentsimon:
> > Based on the comment, it may be better to invest in unit test in TS code. Wdut?
> 
> I think this is OK -- let's do this. If there are action syntax errors somewhere we'll eventually catch them in post-submits or e2e tests even if there's a lag.

thinking about it more, unit tests also need to be run with `id-token: write`. Or I need to push the run verification to the very end to test all the rest.

.github/workflows/pre-submit.actions.yml
ianlewis: ah right.

ianlewis:
> thinking about it more, unit tests also need to be run with `id-token: write`. Or I need to push the run verification to the very end to test all the rest.

Units should test code without necessarily running on a GH Actions by mocking or otherwise. Agree we need to have some integration/e2e that runs on GHA though.

laurentsimon:
> > thinking about it more, unit tests also need to be run with `id-token: write`. Or I need to push the run verification to the very end to test all the rest.
> 
> Units should test code without necessarily running on a GH Actions by mocking or otherwise. Agree we need to have some integration/e2e that runs on GHA though.

The Action code is simple and contains a single function. It feels weird to break every apart to test each part of it. Maybe we do most of the testing in scheduled runs instead. The Action is unlikely to change, so the delay in catching the problems may be acceptable.
Another solution is to use pull_request_target, but it's a bit dangerous and a pain to setup...

ianlewis:
> > > thinking about it more, unit tests also need to be run with `id-token: write`. Or I need to push the run verification to the very end to test all the rest.
> > 
> > 
> > Units should test code without necessarily running on a GH Actions by mocking or otherwise. Agree we need to have some integration/e2e that runs on GHA though.
> 
> The Action code is simple and contains a single function. It feels weird to break every apart to test each part of it. Maybe we do most of the testing in scheduled runs instead. The Action is unlikely to change, so the delay in catching the problems may be acceptable. Another solution is to use pull_request_target, but it's a bit dangerous and a pain to setup...

Yeah, I agree. We'll at least need to have scheduled e2e tests. Unit-tests are lower priority but would still be nice to be able to catch issues in pre-submits rather than having a lot of lag time.

ianlewis:
> We do have the scheduled e2e tests already in this repo, btw. I'll see what I can do for unit tests. I may close this PR then

Right. I'm not sure these pre-submits are really adding value. For unit tests I had created #1471 already so we can track them there and add them as they make sense.

laurentsimon:
alright. Closing this PR then. Thanks!

PR #1481, opened 2023-01-09
Write subjects to file

### LINKED ISSUE #1191: [feature] Create a docker-based builder binary that executes a builder image to produce a binary
This is a tracking issue for the builder binary that is responsible for executing a docker-based-build SLSA [buildDefinition](https://github.com/MarkLodato/slsa/blob/cec378584e1b24c84fd0a60a0ae5f269e045f4c3/docs/provenance/examples.md#docker-based-rebuilder) and create and measure the output artifact.

The builder binary would take as input:
* The SLSA predicate input: this contains the source, builder base image, and command/output/environment variable. This contains all the needed information to (re)build the artifact

And output:
* The output binary 

Other:
* Note that the builder binary is part of the predicate itself. This is self-invoking. The outer layer should likely add this into the provenance artifacts, potentially in the runDetails: the type of this builder shouldn't affect the rebuild.

**Additional context**
https://github.com/slsa-framework/slsa-github-generator/issues/23
https://github.com/MarkLodato/slsa/blob/cec378584e1b24c84fd0a60a0ae5f269e045f4c3/docs/provenance/examples.md

cc @rbehjati

rbehjati:
Fixes #1191


internal/builders/docker/commands.go
```diff
@@ -70,11 +73,16 @@ func writeBuildDefinitionToFile(bd pkg.BuildDefinition, path string) error {
...
 	var forceCheckout bool
+	var subjectsPath string
 
 	cmd := &cobra.Command{
 		Use:   "build [FLAGS]",
 		Short: "Builds the artifacts using the build config, source repo, and the builder image.",
 		Run: func(cmd *cobra.Command, args []string) {
+			// The BuildArtifacts function changes directory. Here we convert
+			// the output path `subjectsPath` to an absolute path.
+			outPath, err := filepath.Abs(subjectsPath)
```
laurentsimon, line 84: Can you make use of https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/utils/path.go#L40 and https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/utils/path.go#L74 to avoid directory traversal and overwrite of existing files?

/cc @asraa

asraa: Ohh, maybe internally in `writeFile` so it can happen for buildDefinition output too.

internal/builders/docker/commands.go
```diff
@@ -53,14 +56,14 @@ func DryRunCmd(check func(error)) *cobra.Command {
 	return cmd
 }
 
-func writeBuildDefinitionToFile(bd pkg.BuildDefinition, path string) error {
-	bytes, err := json.Marshal(bd)
+func writeToFile[T any](obj T, path string) error {
+	bytes, err := json.Marshal(obj)
 	if err != nil {
-		return fmt.Errorf("couldn't marshal the BuildDefinition: %v", err)
+		return fmt.Errorf("marshaling the object failed: %v", err)
```
ianlewis, line 62:
```suggestion
		return fmt.Errorf("marshaling the object failed: %w", err)
```

internal/builders/docker/commands.go
```diff
@@ -53,14 +56,14 @@ func DryRunCmd(check func(error)) *cobra.Command {
...
 	}
 
 	if err := os.WriteFile(path, bytes, 0o600); err != nil {
-		return fmt.Errorf("couldn't write BuildDefinition to file: %v", err)
+		return fmt.Errorf("writing to file failed: %v", err)
```
ianlewis, line 66:
```suggestion
		return fmt.Errorf("writing to file failed: %w", err)
```

internal/builders/docker/commands.go
rbehjati: Done. Is this the convention everywhere, or only for errors that need to be unwrapped?

internal/builders/docker/commands.go
rbehjati: Done.

internal/builders/docker/commands.go
rbehjati: Thanks. Replaces it with `CreateNewFileUnderCurrentDirectory`. Also changed `writeToFile` to take a writer as input.

rbehjati (commented):
Thanks for the reviews.

asraa (approved):

Issue #1471, opened 2023-01-05
[feature] unit tests for internal actions


ianlewis:
We should have some unit tests for internal actions so that we have a bit more confidence that updating dependencies etc. don't break things.

- [ ] generate-attestations
- [ ] .github/actions/compute-sha256
- [ ] .github/actions/privacy-check
- [ ] .github/actions/generate-attestations
- [ ] .github/actions/sign-attestations
- [ ] ./actions/setup-token

asraa:
I have jest set up for unit testing in generate-attestations!

ianlewis:
> I have jest set up for unit testing in generate-attestations!

I saw that! very nice!

> For discussion: sign-attestations and setup-token require OIDC -- we currently put those in scheduled tests. I can add unit testing in a way that skips the main functionality. Thoughts? Maybe I can look into faking the OIDC client? Not sure

Yeah, I think most actions can have functionality that can be unit tested if we get it out of the main code. Much can be mocked in TS/JS, but it may be a balance between unit and integration(?) tests that make use of GHA infrastructure to test OIDC etc. Many of our actions don't have a lot of code so it may not provide a huge amount of benefit but it helps sometimes to be able to run tests locally.

PR #1482, opened 2023-01-09
fix: fix sign-attestations output path

asraa:

Fixes https://github.com/slsa-framework/slsa-github-generator/issues/1472

Working on adding `main` tests that mock `sigstore` for this one. For https://github.com/slsa-framework/slsa-github-generator/issues/1471


.github/actions/sign-attestations/src/main.ts
```diff
@@ -50,11 +56,10 @@ async function run(): Promise<void> {
         );
         const bundleStr = JSON.stringify(bundle);
         // We detect path traversal for safeOutputFolder, so this should be safe.
-        const outputPath = path.join(
+        const outputPath = `${path.join(
```
laurentsimon, line 59: I dont think you need this. The following should work, no:
```
const outputPath = path.join(
           safeOutputFolder
           path.basename(file) + ".sigstore",
        );
```

laurentsimon (approved):
Thanks!

.github/actions/sign-attestations/src/main.ts
asraa: npm style doesn't like string concatenation and prefers templating D:

/home/asraa/git/slsa-github-generator/.github/actions/sign-attestations/src/main.ts
  61:11  error  Unexpected string concatenation  prefer-template

But it's cleaner to do similar to what you suggested:
```
        const outputPath = path.join(
          safeOutputFolder,
          `${path.basename(fpath)}.sigstore`);
```

.github/actions/sign-attestations/dist/index.js
```diff
@@ -37127,7 +37133,7 @@ function run() {
                     const bundle = yield sigstore.sigstore.signAttestation(buffer, payloadType, signOptions);
                     const bundleStr = JSON.stringify(bundle);
                     // We detect path traversal for safeOutputFolder, so this should be safe.
-                    const outputPath = path.join(safeOutputFolder, path.basename(file), ".sigstore");
+                    const outputPath = `${path.join(safeOutputFolder, path.basename(fpath))}.sigstore`;
```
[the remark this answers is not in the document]
asraa: this is the generated

.github/actions/sign-attestations/src/main.ts
laurentsimon: Got it, thanks!

asraa (commented):
updated, also realized I checked in a spurious outputs/ directory created in local test so deleted here.

PR #1485, opened 2023-01-10
feat(docker): update docker based workflow to use buildDefinition and sign

asraa:

This updates the docker based workflow to use the resolved `buildDefinition` and completes the workflow to sign the attestation. The presubmit is also updated to be functional.

Passing run: https://github.com/asraa/slsa-github-generator/actions/runs/3878872921

The next missing pieces:
* Populate rest of predicate with GH context

Fixes https://github.com/slsa-framework/slsa-github-generator/issues/1499

cc @rbehjati

### DIFF
--- a/.github/workflows/builder_docker-based_slsa3.yml
+++ b/.github/workflows/builder_docker-based_slsa3.yml
@@ -23,6 +23,8 @@ env:
   # Builder.
   BUILDER_BINARY: slsa-builder-docker-linux-amd64 # Name of the binary in the release assets.
   BUILDER_DIR: internal/builders/docker # Source directory if we compile the builder.
+  # Outputs folder.
+  OUTPUT_FOLDER: outputs
 
 defaults:
   run:
@@ -171,7 +173,7 @@ jobs:
           allow-private-repository: ${{ inputs.private-repository }}
 
       - name: Upload builder
-        uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # tag=v3.1.1
+        uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # v3.1.1
         with:
           name: "${{ env.BUILDER_BINARY }}-${{ needs.rng.outputs.value }}"
           path: "${{ env.BUILDER_BINARY }}"
@@ -211,10 +213,10 @@ jobs:
           SOURCE_DIGEST: ${{ inputs.source-digest }}
           GITHUB_SHA: ${{ github.sha }}
         run: |
-          if [ -z "$SOURCE_DIGEST" ]; then 
+          if [ -z "$SOURCE_DIGEST" ]; then
             ref=$(echo ${SOURCE_DIGEST} | cut -d':' -f2)
             echo "ref=$ref" >> $GITHUB_OUTPUT
-          else 
+          else
             echo "ref=$GITHUB_SHA" >> $GITHUB_OUTPUT
           fi
 
@@ -245,18 +247,18 @@ jobs:
 
           # Note: this outputs information about resolved arguments, etc.
           # the values are trusted because the compiler is not invoked.
-          echo "$GITHUB_WORKSPACE/$BUILDER_BINARY" dry-run \
-            --build-config-path ${CONFIG_PATH} \
+          echo "${GITHUB_WORKSPACE}/${BUILDER_BINARY}" dry-run \
+            --build-config-path "${CONFIG_PATH}" \
             --build-definition-path build-definition.json \
-            --builder-image ${BUILDER_IMAGE}@${BUILDER_DIGEST} \
-            --git-commit-hash ${SOURCE_DIGEST} \
-            --source-repo git+https://github.com/${SOURCE_REPOSITORY}
-          "$GITHUB_WORKSPACE/$BUILDER_BINARY" dry-run \
-            --build-config-path ${CONFIG_PATH} \
+            --builder-image "${BUILDER_IMAGE}@${BUILDER_DIGEST}" \
+            --git-commit-hash "${SOURCE_DIGEST}" \
+            --source-repo "git+https://github.com/${SOURCE_REPOSITORY}"
+          "${GITHUB_WORKSPACE}/${BUILDER_BINARY}" dry-run \
+            --build-config-path "${CONFIG_PATH}" \
             --build-definition-path build-definition.json \
-            --builder-image ${BUILDER_IMAGE}@${BUILDER_DIGEST} \
-            --git-commit-hash ${SOURCE_DIGEST} \
-            --source-repo git+https://github.com/${SOURCE_REPOSITORY}
+            --builder-image "${BUILDER_IMAGE}@${BUILDER_DIGEST}" \
+            --git-commit-hash "${SOURCE_DIGEST}" \
+            --source-repo "git+https://github.com/${SOURCE_REPOSITORY}"
 
           echo "build-definition-name=build-definition.json" >> $GITHUB_OUTPUT
 
@@ -332,7 +334,7 @@ jobs:
     needs: [rng, detect-env, build, generate-build-definition, generate-builder]
     permissions:
       id-token: write # Needed to create an OIDC token for keyless signing.
-      contents: read
+      contents: read  # Needed to check out the repository.
       actions: read # Needed to read workflow info.
     outputs:
       provenance: ${{ steps.sign.outputs.output-name }}
@@ -388,10 +390,10 @@ jobs:
 
       - name: Upload unsigned intoto attestations file for pull request
         if: ${{ github.event_name == 'pull_request' }}
-        uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # tag=v3.1.1
+        uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # v3.1.1
         with:
-          name: "attestations-${{ needs.rng.outputs.value }}"
-          path: "attestations"
+          name: "${{ env.OUTPUT_FOLDER }}"
+          path: "attestations-${{ needs.rng.outputs.value }}"
           if-no-files-found: error
 
       ###################################################################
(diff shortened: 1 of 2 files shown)


.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -143,90 +153,118 @@ jobs:
...
+          GITHUB_SHA: ${{ github.sha }}
+        run: |
+          if [ -z "$SOURCE_DIGEST" ]; then 
+            ref=$(echo ${SOURCE_DIGEST} | cut -d':' -f2)
+            echo "ref=$ref" >> $GITHUB_OUTPUT
+          else 
+            echo "ref=$GITHUB_SHA" >> $GITHUB_OUTPUT
+          fi
+
+      # TODO: Use secure-checkout, but this requires configuring the source-repository.
```
laurentsimon, line 221: is this already done via `ref: ${{ steps.resolve.outputs.ref }}`?

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -143,90 +153,118 @@ jobs:
...
+          # Note: this outputs information about resolved arguments, etc.
+          # the values are trusted because the compiler is not invoked.
+          echo "$GITHUB_WORKSPACE/$BUILDER_BINARY" dry-run \
+            --build-config-path ${CONFIG_PATH} \
+            --build-definition-path build-definition.json \
+            --builder-image ${BUILDER_IMAGE}@${BUILDER_DIGEST} \
+            --git-commit-hash ${SOURCE_DIGEST} \
+            --source-repo git+https://github.com/${SOURCE_REPOSITORY}
+          "$GITHUB_WORKSPACE/$BUILDER_BINARY" dry-run \
+            --build-config-path ${CONFIG_PATH} \
```
laurentsimon, line 255: nit: add double quote for all env variables

laurentsimon (approved):

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -143,90 +153,118 @@ jobs:
...
+          ref: "${{ needs.detect-env.outputs.ref }}"
+          go-version: 1.19
+          # Note: This must be the non-randomized binary name, so that it can be downloaded from the release assets.
+          binary: "${{ env.BUILDER_BINARY }}"
+          compile-builder: "${{ inputs.compile-builder }}"
+          directory: "${{ env.BUILDER_DIR }}"
+          allow-private-repository: ${{ inputs.private-repository }}
+
+      - name: Upload builder
+        uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # tag=v3.1.1
```
ianlewis, line 174: Renovate doesn't seem to use `tag=` anymore.
```suggestion
        uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # v3.1.1
```

.github/workflows/builder_docker-based_slsa3.yml
ianlewis: I think the issue is that this workflow takes a `source-repository` input which we don't do for our other workflows in order to make sure we check out the right project repository.

@asraa @rbehjati I'm not sure I understand the need for the `source-repository` input? Why isn't the `github.repository` always used?

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -143,90 +153,118 @@ jobs:
...
+        run: |
+          if [ -z "$SOURCE_DIGEST" ]; then 
+            ref=$(echo ${SOURCE_DIGEST} | cut -d':' -f2)
+            echo "ref=$ref" >> $GITHUB_OUTPUT
+          else 
+            echo "ref=$GITHUB_SHA" >> $GITHUB_OUTPUT
+          fi
+
+      # TODO: Use secure-checkout, but this requires configuring the source-repository.
+      - name: Checkout the source repository
```
asraa, line 222: @rbehjati I believe I actually don't need this here, since the resolution / dry-run step doesn't actually need the source checkout. One question though about this for the actual build step: I believe the builder binary will create the checkout itself when `--force-checkout` is passed. Is it always intended to? Or can the workflow check out the source repo? It didn't seem to skip the checkout step when run from a source repository root.

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -143,90 +153,118 @@ jobs:
...
+            --build-definition-path build-definition.json \
+            --builder-image ${BUILDER_IMAGE}@${BUILDER_DIGEST} \
+            --git-commit-hash ${SOURCE_DIGEST} \
+            --source-repo git+https://github.com/${SOURCE_REPOSITORY}
+          "$GITHUB_WORKSPACE/$BUILDER_BINARY" dry-run \
+            --build-config-path ${CONFIG_PATH} \
+            --build-definition-path build-definition.json \
+            --builder-image ${BUILDER_IMAGE}@${BUILDER_DIGEST} \
+            --git-commit-hash ${SOURCE_DIGEST} \
+            --source-repo git+https://github.com/${SOURCE_REPOSITORY}
```
ianlewis, line 259:
```suggestion
          # Note: this outputs information about resolved arguments, etc.
          # the values are trusted because the compiler is not invoked.
          echo "${GITHUB_WORKSPACE}/${BUILDER_BINARY}" dry-run \
            --build-config-path "${CONFIG_PATH}" \
            --build-definition-path build-definition.json \
            --builder-image "${BUILDER_IMAGE}@${BUILDER_DIGEST}" \
            --git-commit-hash "${SOURCE_DIGEST}" \
            --source-repo "git+https://github.com/${SOURCE_REPOSITORY}"
          "${GITHUB_WORKSPACE}/${BUILDER_BINARY}" dry-run \
            --build-config-path "${CONFIG_PATH}" \
            --build-definition-path build-definition.json \
            --builder-image "${BUILDER_IMAGE}@${BUILDER_DIGEST}" \
            --git-commit-hash "${SOURCE_DIGEST}" \
            --source-repo "git+https://github.com/${SOURCE_REPOSITORY}"
```

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -292,6 +330,10 @@ jobs:
   provenance:
     runs-on: ubuntu-latest
     needs: [rng, detect-env, build, generate-build-definition, generate-builder]
+    permissions:
+      id-token: write # Needed to create an OIDC token for keyless signing.
+      contents: read
```
ianlewis, line 335:
```suggestion
      contents: read # Needed to check out the repository.
```

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -363,40 +383,34 @@ jobs:
...
+          predicate-file: "${{ needs.generate-build-definition.outputs.build-definition-name }}"
           output-folder: "attestations-${{ needs.rng.outputs.value }}"
 
       - name: Upload unsigned intoto attestations file for pull request
         if: ${{ github.event_name == 'pull_request' }}
         uses: actions/upload-artifact@83fd05a356d7e2593de66fc9913b3002723633cb # tag=v3.1.1
         with:
           name: "attestations-${{ needs.rng.outputs.value }}"
-          path: "attestations-${{ needs.rng.outputs.value }}"
+          path: "attestations"
```
ianlewis, line 394: Is this right? I thought this should be the same as `output-folder` on line 387?

.github/workflows/builder_docker-based_slsa3.yml
asraa: Yes I'm actually not sure that it's needed as a workflow level input. I understand it's useful to have the builder binary the source repository/source digest arguments, but inside the workflow, I don't expect this to differ.

I'll discuss online with @rbehjati today. If we remove, I'll update to clean that up.

.github/workflows/builder_docker-based_slsa3.yml
asraa: Ah -- for a pull request event, yes, this is the final output and should be uploaded to a final outputs folder.

I will make the outputs folder an `env` var of the workflow. I'm not so certain that it should be "outputs" hard-coded anyway. Perhaps the workflow outputs should be a string `outputs-folder` that contains either unsigned attestations (for no OIDC token access) or the signed attestations (for real events).

.github/workflows/builder_docker-based_slsa3.yml
rbehjati: Yes. The source repo should be fetched from the GitHub context. But it is needed as an input to the builder command, because it has to be included in the `BuildDefinition`.

.github/workflows/builder_docker-based_slsa3.yml
rbehjati: You are right. Checkout is not needed here. 

For the `build` subcommand, it is more efficient to have the workflow check out the repo, instead of passing in the `--force-checkout` flag.

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -143,90 +155,118 @@ jobs:
...
         # These are the inputs, it may be with: for an action or
         # specified with these env vars.
         env:
           BUILDER_IMAGE: ${{ inputs.builder-image }}
           BUILDER_DIGEST: ${{ inputs.builder-digest }}
-          BUILDER_VERIFIED: ${{ needs.verify-builder-image.outputs.verified }}
           CONFIG_PATH: ${{ inputs.config-path }}
-          OUTPUT: ${{ inputs.builder-output-path }}
           # Either the provided source repository or the caller's GITHUB_REPOSITORY
           SOURCE_REPOSITORY: ${{ inputs.source-repository || github.repository }}
```
rbehjati: I think it is better to drop `inputs.source-repository`. Unless we want to allow building artefacts from arbitrary repositories. In that case we have to in `--force-checkout` to the build subcommand.

.github/workflows/builder_docker-based_slsa3.yml
asraa: Removed from workflow inputs, so we always use the default context.

.github/workflows/builder_docker-based_slsa3.yml
asraa: done! workflow now performs the checkout too and drops `--force-checkout`

.github/workflows/builder_docker-based_slsa3.yml
asraa: done!

.github/workflows/builder_docker-based_slsa3.yml
asraa: Done here as far as placing in the same `output-folder`.

Issue #1501, opened 2023-01-10
[bug] [docker] nilpoint panic on build execution


asraa:
**Describe the bug**
I retrieved the following panic

```
/home/runner/work/slsa-github-generator/slsa-github-generator/slsa-builder-docker-linux-amd64 build --build-config-path internal/builders/docker/testdata/config.toml --builder-image bash@sha256:9e2ba52487d945504d250de186cb4fe2e3ba023ed2921dd6ac8b97ed43e76af9 --git-commit-digest sha1:04516f19474da0123ee96c5c9b5f25821c93701d --source-repo git+https://github.com/slsa-framework/slsa-github-generator --subjects-path subjects.json --force-checkout
2023/01/10 16:11:22 Checking out the repo in "/tmp/release-1122329179".
2023/01/10 16:11:22 Cloning the repo from https://github.com/slsa-framework/slsa-github-generator...
2023/01/10 16:11:24 'git clone' completed. See "/tmp/log-1778144680.txt", and "/tmp/log-3265357455.txt" for logs, and errors.
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x616d8d]

goroutine 1 [running]:
main.BuildCmd.func1(0xc0000ee600?, {0x675c36?, 0xb?, 0xb?})
	/home/runner/work/slsa-github-generator/slsa-github-generator/__BUILDER_CHECKOUT_DIR__/internal/builders/docker/commands.go:82 +0x16d
github.com/spf13/cobra.(*Command).execute(0xc0000ee600, {0xc0001000b0, 0xb, 0xb})
	/home/runner/work/slsa-github-generator/slsa-github-generator/__BUILDER_CHECKOUT_DIR__/vendor/github.com/spf13/cobra/command.go:920 +0x847
github.com/spf13/cobra.(*Command).ExecuteC(0xc0000ee000)
	/home/runner/work/slsa-github-generator/slsa-github-generator/__BUILDER_CHECKOUT_DIR__/vendor/github.com/spf13/cobra/command.go:1044 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
	/home/runner/work/slsa-github-generator/slsa-github-generator/__BUILDER_CHECKOUT_DIR__/vendor/github.com/spf13/cobra/command.go:968
main.main()
[...]

PR #1388, opened 2022-12-15
Add builder functionality 

rbehjati:
Ref #1191 

Builds on top of #1387 and adds builder functionality.

The remaining part is to write the intoto.Subjects for the generated artifacts into a file. I'll do this in a future PR.

PR #1502, opened 2023-01-10
fix: fix defer cleanup before error checking db

asraa:

This was introduced here https://github.com/slsa-framework/slsa-github-generator/pull/1388/commits/e6231cc2161b8924d56ae1abea0da49ce9a246fb but there is a cleanup function that is set before the error on `db` is checked, which gives a nilptr exception. 

Fixes https://github.com/slsa-framework/slsa-github-generator/issues/1501

/review @rbehjati


laurentsimon (approved):

rbehjati (approved):
Ah! Thanks :)

PR #1547, opened 2023-01-17
fix: Unpack config file in BuildDefinition

### LINKED ISSUE #1503: [feature] Expand the content of the config file in the BuildDefinition
We should consider including the `artifact_path` and `command` as additional `externalParameters` in the `BuildDefinition`. This allows easier verification of these values. 

**Describe alternatives you've considered**
Without this feature, performing any checks on `artifact_path` and `command` requires fetching the config file from the source and parsing it. This feels like too much overhead for this kind of check. 

**Additional context**
Ref #1191 

cc @asraa

rbehjati:
Fixes #1503 

To unpack the input config file, even in case of `dry-run` we have to fetch the source, and load the config file. Because of this, some parts of `builder.go` are refactored as part of this PR. The benefit of this is that now `dry-run` and `build` commands now have more in common. In particular, `dry-run` performs more of the checks in `build`.

### DIFF
--- a/.github/workflows/builder_docker-based_slsa3.yml
+++ b/.github/workflows/builder_docker-based_slsa3.yml
@@ -176,6 +176,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [rng, detect-env, verify-builder-image, generate-builder]
     steps:
+      - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
       - name: Checkout builder repository
         uses: slsa-framework/slsa-github-generator/.github/actions/secure-builder-checkout@main
         with:
--- a/internal/builders/docker/pkg/builder.go
+++ b/internal/builders/docker/pkg/builder.go
@@ -106,8 +106,12 @@ func (db *DockerBuild) CreateBuildDefinition() *BuildDefinition {
 	artifacts[SourceKey] = sourceArtifact(db.config)
 	artifacts[BuilderImageKey] = builderImage(db.config)
 
-	// Ignoring the error, as the input is a simple string array.
-	cmdBytes, _ := json.Marshal(db.buildConfig.Command)
+	// The input is a simple string array, no errors occur. But we check and
+	// print the error to make the linters happy.
+	cmdBytes, err := json.Marshal(db.buildConfig.Command)
+	if err != nil {
+		log.Printf("Could not unmarshal command array: %v.", err)
+	}
 	cmd := string(cmdBytes)
 
 	ep := ParameterCollection{


internal/builders/docker/testdata/build-definition.json
```diff
@@ -16,6 +16,8 @@
             }
         },
         "values": {
+            "artifactPath": "config.toml",
+            "command": "cp, internal/builders/docker/testdata/config.toml, config.toml",
```
rbehjati, line 20: I am not sure if this is how we want the command to be included in the BuildDefinition. Ideally, the command should be an array, but the format only allows a string here.

A better encoding would be `"\"cp\", \"internal/builders/docker/testdata/config.toml\", \"config.toml\""` but I am not sure what is the conventional way to do it in Golang. @asraa Any ideas?

rbehjati: Modified it to a JSON encoded array. This should generally work better.

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -208,13 +218,13 @@ jobs:
             --build-config-path "${CONFIG_PATH}" \
             --build-definition-path build-definition.json \
             --builder-image "${BUILDER_IMAGE}@${BUILDER_DIGEST}" \
-            --git-commit-digest "sha1:${GITHUB_SHA}" \
+            --git-commit-digest "sha1:${{needs.commit.outputs.sha}}" \
```
asraa, line 221: Ah, are you having trouble getting `GITHUB_SHA`? Hmmm

rbehjati: Yes. In case of pull requests it is not the actual commit. I am trying to get the latest commit, but now I get the error `fatal: not a git repository (or any of the parent directories): .git`. This is very strange.

internal/builders/docker/commands.go
```diff
@@ -45,13 +45,19 @@ func DryRunCmd(check func(error)) *cobra.Command {
 			config, err := pkg.NewDockerBuildConfig(io)
 			check(err)
 
-			bd := pkg.CreateBuildDefinition(config)
-			check(writeToFile(*bd, w))
+			builder, err := pkg.NewBuilderWithGitFetcher(*config)
+			check(err)
+
+			db, err := builder.SetUpBuildState()
```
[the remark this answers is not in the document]
rbehjati: I don't think so. The command that I have in the README works.

internal/builders/docker/pkg/builder.go
```diff
@@ -98,15 +99,24 @@ func NewBuilderWithGitFetcher(config DockerBuildConfig, forceCheckout bool) (*Bu
...
+
+	// Ignoring the error, as the input is a simple string array.
+	cmdBytes, _ := json.Marshal(db.buildConfig.Command)
+	cmd := string(cmdBytes)
 
 	ep := ParameterCollection{
 		Artifacts: artifacts,
-		Values:    map[string]string{ConfigFileKey: config.BuildConfigPath},
+		Values: map[string]string{
+			ConfigFileKey:   db.config.BuildConfigPath,
```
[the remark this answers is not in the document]
rbehjati: I could add that prefix. Do you mean also in the BuildDefinition? There we could have "inputConfigFile" and "inputCommand". Also I think in the JSON encoded build definition, we want to use camel case, right?

.github/workflows/builder_docker-based_slsa3.yml
rbehjati: I thought `detect-env` would give the correct sha, but it gives empty string.

asraa: Ohhh gotcha, for `pull_request` events it does appear `GITHUB_SHA` is not the triggering commit (https://github.com/orgs/community/discussions/26325). But for that, wouldn't the builder need to checkout the base repository (which may be the fork?)

> I thought detect-env would give the correct sha, but it gives empty string.

detect-env gives the ref of the reusable workflow, not the triggering environment. hmmm. It should give the head sha of that, let me see if I can test that out as well.

internal/builders/docker/pkg/builder.go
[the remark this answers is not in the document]
asraa: Interesting, I found the example here: https://github.com/slsa-framework/slsa/blob/3b728b4c19504d4a5452a7ddfa92acffff2c5c43/docs/github-actions-workflow/v0.1/example.json#L7

where Mark suggests prefixing the workflow inputs (i guess specifically config file) with `inputs_`. I am going to comment about the camelcasing issue.

internal/builders/docker/commands.go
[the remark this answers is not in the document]
rbehjati: Ah! I misunderstood the question... 

With this change, we need to load the config file, for that `git checkout` is needed.

internal/builders/docker/pkg/builder.go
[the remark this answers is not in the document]
rbehjati: Thanks. I am fine either way, but I thought the convention was to have camel case in JSON.

.github/workflows/builder_docker-based_slsa3.yml
rbehjati: >  But for that, wouldn't the builder need to checkout the base repository (which may be the fork?)

Yes, but even in that case the `GITHUB_SHA` we provide to the command is not something that exists on the repo. 

It should be possible to get the actual commit hash. I have one more idea to try :)

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -208,13 +220,13 @@ jobs:
             --build-config-path "${CONFIG_PATH}" \
             --build-definition-path build-definition.json \
             --builder-image "${BUILDER_IMAGE}@${BUILDER_DIGEST}" \
-            --git-commit-digest "sha1:${GITHUB_SHA}" \
+            --git-commit-digest "sha1:${{ github.event.pull_request.head.sha }}" \
```
rbehjati, line 223: With this change we are getting the correct commit sha, but I think we also need to specify the correct branch in `source-repo`. I am not sure if that is possible. I'll give it a try.

ianlewis: IIUC the head SHA will only be available for `pull_request` events? Are there other events you'll need to handle? I assume that the primary event you would need to handle is `tag` or `release` for when release tags are created.

rbehjati: Yes. We have to be able to handle all these events. I am thinking of reverting this back to use `GITHUB_SHA`, but make these steps conditional on events that are not `pull_request`, and add a separate step for `pull_request`.

ianlewis: Yeah, we have a lot of places throughout that act differently for pull requests.

I'm curious how you are planning to support pull requests? we have the problem that `id-token:write` isn't available on pull requests (https://github.com/slsa-framework/slsa-github-generator/issues/131) so we aren't really able to sign provenance for those events currently.

rbehjati: For pull-requests we only run the `dry-run` command. We won't actually generate the provenance or sign, but just to make sure that the inputs are correct, and that we can create a BuildDefinition. 

The idea is to allow the users to verify their workflow parameters before merging.

asraa: > but make these steps conditional on events that are not pull_request, and add a separate step for pull_request.

+1

@ianlewis also so far we can also do this:
https://github.com/slsa-framework/slsa-github-generator/blob/76f03fa7e30209f32ac76ce417ddc43ff98af42a/.github/workflows/builder_docker-based_slsa3.yml#L396

since the signing is isolated in a step. in pull_request we just published the unsigned attestations.

there's a question of whether there are other events besides pull_request that don't have id-token. Perhaps we should actually replace the `if pull_requests` to `if -n $ACTIONS_ID_TOKEN_REQUEST_URL`.

internal/builders/docker/pkg/builder.go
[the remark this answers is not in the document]
asraa: No prob! @MarkLodato  updated it to camelcase 
https://github.com/slsa-framework/slsa/pull/525#discussion_r1080573190

.github/workflows/builder_docker-based_slsa3.yml
ianlewis: I think detect-env has special case for pull requests to enable pre-submit tests. We don't really support it out outside of that context. The action may need to be adjusted.

.github/workflows/builder_docker-based_slsa3.yml
ianlewis: You'll at least need to detect pull request actions so you know what sha to use so it might be best to just use that.

If the code for each is the same outside of deciding what sha to use you could set the correct sha in a prior build step.

```yaml
- name: Detect SHA
  id: get-sha
  env:
    GITHUB_HEAD_SHA: "${{ github.event.pull_request.head.sha }}"
  run: |
    if [ "$GITHUB_EVENT_NAME" == "pull_request" ]; then
      echo "sha=$GITHUB_HEAD_SHA" >> $GITHUB_OUTPUT
    else
      echo "sha=$GITHUB_SHA" >> $GITHUB_OUTPUT
    fi
```

and then use it

`${{ steps.get-sha.outputs.sha }}`

rbehjati: I am not sure if it is even possible to make it work on pull_requests. The issue is not just about the commit sha, but also the repo-url. I am not sure what the repo-url could be in the case of a pull-request.

So I am thinking of a completely different approach:

The first thing that happens during [the setup step](https://github.com/slsa-framework/slsa-github-generator/blob/76f03fa7e30209f32ac76ce417ddc43ff98af42a/internal/builders/docker/pkg/builder.go#L139) is to check if the latest commit matches the given commit (this happens in [`verifyCommit`](https://github.com/slsa-framework/slsa-github-generator/blob/76f03fa7e30209f32ac76ce417ddc43ff98af42a/internal/builders/docker/pkg/builder.go#L303)). If the commits don't match, an error is returned, unless `force-checkout` is provided in which case the repo is cloned. 

I am thinking of allowing a third option. When `verifyCommit` detects a commit that does not match the input commit, we could do either of:
1. Abort. This should be the default option.
2. Force checkout.
3. Use the existing checked out commit as the source. This option only makes sense if the repo is checked out from the same repo that is given in the repo-url. We can check this using `git remote get-url origin` (this command only works if there is a remote named `origin`). 

So I am thinking of replacing the `--force-checkout` option with another flag, e.g., `--unexpected-commit` which could take one of these three values: `abort`, `force-checkout` and `ignore`; with `abort` being the default. 

This way, for `dry-run` we could use `--unexpected-commit=ignore` to allow generating the `BuildDefinition` using the config file on the current working directory even in the case of a `pull-request` without having to clone the repo.

What do you think?

I could give it a try in this PR.

ianlewis: I think that makes sense as long as it's only an option available on the dry run

rbehjati: Perfect. I'll give it a try.

asraa: > and the problem with pull-request events was that the repository was not checked out for the dry-run workflow. 

OH NO!

OK so just to clarify what happens here now: `actions/checkout` by default checks out the github repo at the ref of the merge commit. I think that also references `GITHUB_SHA`, and so the comparison with the checked out repo and the `GITHUB_SHA` of the command line input should always match, pull request or not?

secondly, the new `resolution-strategy` will allow dry-run to create a build definition, ignoring the current checkout and using the command line provided git-commit-digest? if so, i think that makes sense.

what happens to the digest of the config file though? is it checked out at the current checkout or at the git-commit-digest during `--resolution-strategy ignore`? (It seems like `ignore` is like `override` with the current checkout so it uses the current checkout for the config file)

rbehjati: > OK so just to clarify what happens here now: `actions/checkout` by default checks out the github repo at the ref of the merge commit. I think that also references `GITHUB_SHA`, and so the comparison with the checked out repo and the `GITHUB_SHA` of the command line input should always match, pull request or not?

Yes. After all these experiments that is my understanding too. This behaviour makes sense, and makes things simpler :)

> secondly, the new `resolution-strategy` will allow dry-run to create a build definition, ignoring the mismatch and using the current checkout for the source digest? if so, i think that makes sense.

True. But actually this feature is no longer required, at least not for GitHub. So we could either keep it, or drop it in favour of making the code simpler :)

> what happens to the digest of the config file though? is it checked out at the current checkout or at the git-commit-digest during `--resolution-strategy ignore`? (It seems like `ignore` is like `override` with the current checkout so it uses the current checkout for the config file)

It is checked out at the current commit, or what I call the `effective commit`. The generated `BuildDefinition` also contains this effective commit digest, not the one given in the inputs.

asraa: > So we could either keep it, or drop it in favour of making the code simpler :)

Should we keep, and file a follow-up issue to clean it if the code is complete and we find it's not useful for testing?

internal/builders/docker/pkg/options.go
```diff
@@ -37,4 +38,7 @@ func (io *InputOptions) AddFlags(cmd *cobra.Command) {
 
 	cmd.Flags().StringVarP(&io.BuilderImage, "builder-image", "i", "",
 		"Required - URL indicating the Docker builder image, including a URI and image digest.")
+
+	cmd.Flags().BoolVarP(&io.ForceCheckout, "force-checkout", "f", false,
```
asraa, line 42: this can go away, right?

asraa (approved):
one small nit, but otherwise lgtm. thank you for all the documentation!

internal/builders/docker/pkg/options.go
rbehjati: Good catch. Thanks.

.github/workflows/builder_docker-based_slsa3.yml
rbehjati: I can actually drop these changes easily, I just need to revert the last commit.

asraa: easy! thank you!

internal/builders/docker/testdata/build-definition.json
asraa: Issue filed: https://github.com/slsa-framework/slsa-github-generator/issues/1566

.github/workflows/builder_docker-based_slsa3.yml
rbehjati: Done. I think this is now ready to be merged :)

internal/builders/docker/pkg/options.go
asraa: just this one too!

asraa: ah wait, this is reintroduced after removing the 3 val - resolving

asraa (approved):

PR #1553, opened 2023-01-19
feat: [byob] Initial delegated workflow

ianlewis:
WIP delegated workflow for BYOB

TODO: finish implementation

### DIFF
--- a/.github/workflows/builder_delegated_slsa3.yaml
+++ b/.github/workflows/builder_delegated_slsa3.yaml
@@ -1,0 +1,67 @@
+# Copyright 2022 SLSA Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: SLSA Delegated Builder
+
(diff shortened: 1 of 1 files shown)


.github/workflows/builder_delegated_slsa3.yaml
```diff
@@ -0,0 +1,67 @@
...
+          echo "outcome=$([ "$SUCCESS" == "true" ] && echo "success" || echo "failure")" >> $GITHUB_OUTPUT
+
+  verify-slsa-token:
+    - name: Verify SLSA Token
+      uses: slsa-framework/slsa-github-generator/.github/actions/verify-token@main
+      with:
+        # The token must have been created with the "builder_delegated_slsa3.yml" audience.
+        slsa-workflow-recipient: "builder_delegated_slsa3.yml"
+        slsa-unverified-token: ${{ inputs.slsa-token }}
+        # FIXME: what path should this be?
```
asraa, line 66: You should add an RNG job similar to other workflows, and maybe use a `slsa-predicate-${{ rng output }}.json`!

ianlewis: Yeah, thanks. I think I want to make sure it's not going to be stored in the same directory as any checkout directories either but this is still a pretty early version so wasn't sure yet.

asraa: I think in the end this will need to be uploaded anyway. Then after that

build-job:
   run-tool-callback
   upload-slsa-layout

attestation:
   create-attestations (with build output and predicate)
   sign-attestations (if id-token is enabled)
   upload outputs

lmk when you are ready for this one to be merged! happy to just start with this stub.

ianlewis: Yeah, no worries. I just created a draft because I needed to set it aside for a while. Laurent will likely pick this up so will go ahead and close for now.

Issue #1504, opened 2023-01-10
[feature] [byo] Record the tool action inputs into SLSA provenance 


asraa:
**Is your feature request related to a problem? Please describe.**
SLSA v1.0 provenance does not yet supply a canonical place to put derived inputs. In the BYO case, the inputs to the tool may want to be recorded.

These seem like they belong in `runDetails`. It should not be in `externalParameters` as it's not a top-level input, nor should it be in the `systemParameters` as it's not a build platform provided parameter.

PR #1557, opened 2023-01-19
refactor: create common predicate typescript utils

asraa:

Creates a common typescript library for our typescript actions. The first is for `predicate-utils` containing:

* the common SLSAv1Predicate interface definition
* utilities to add GH context to the predicate's parameters


asraa:
`make package` on predicate-utils is not idempotent....

.github/actions/common/predicate-utils/src/index.ts
```diff
@@ -0,0 +1,161 @@
...
+  systemParams["RUNNER_NAME"] = { value: env.RUNNER_NAME || "" };
+  systemParams["RUNNER_OS"] = { value: env.RUNNER_OS || "" };
+
+  if (env.GITHUB_EVENT_NAME === "workflow_dispatch") {
+    if (env.GITHUB_EVENT_PATH) {
+      const ghEvent: WorkflowDispatchEvent = JSON.parse(
+        fs.readFileSync(env.GITHUB_EVENT_PATH).toString()
+      );
+
+      for (const input in ghEvent.inputs) {
```
laurentsimon, line 138: this works for generators, but is that correct for BYOB / builders?
IIUC, for builders, we need to record the TRW's inputs, which are passed to the `setup-token` Action via `slsa-workflow-inputs` field. I think we need to record those.

We temporarily need to read the top-workflow inputs (like in the code) and identify the entries that are not TRW entries, in order to fix https://github.com/actions/runner/issues/2274

Let me know if I mis-understood something

asraa: Here's something I've been thinking about regarding this: 
* Let's keep workflow inputs (the ones extracted here) into systemParameters for ALL workflows. This way this particular function can work solely on modifying systemParmaters to include GH context vars for any time of GH workflow.
* For delegators, we'll add in it's own specific logic to add TRW inputs, since that's a "buildType = delegator" specific thing.

laurentsimon: > Let's keep workflow inputs (the ones extracted here) into systemParameters for ALL workflows

Should this be under `externalParameters` for non-delegator workflows as well? The generic generator *can* report the workflow inputs under `externalParameters` as well to follow v1.0?

asraa: The problem is what is considered "user inputs" for either build type. the workflow inputs or the TRW inputs? If the delegator is attesting to a TRW, then I think the TRW inputs are actually the "top level" and the workflow inputs are the system's parameters.

I think the resolution is this common library can expose `GetInputs` instead, and leave this for the user to decide.

laurentsimon: that makes sense. So the generic generator could use `GetInputs` to set their own `externalParameters`, whereas the delegator would not use the `GetInputs`.

asraa: This is the TODO as well: https://github.com/slsa-framework/slsa-github-generator/issues/1504

I'll table this for an in person discussion, would be good to get the story clear for both types of generators.

.github/actions/common/predicate-utils/Makefile
```diff
@@ -0,0 +1,41 @@
...
+				} \
+			} \
+		}'
+
+node_modules/.installed: package.json package-lock.json
+	npm ci
+	touch node_modules/.installed
+
+.PHONY: action
+action: node_modules/.installed ## Builds the action.
```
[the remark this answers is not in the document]
asraa: Addressed in https://github.com/slsa-framework/slsa-github-generator/pull/1612

PR #1568, opened 2023-01-25
feat: byob reusable workflow

laurentsimon:

A few AIs for me to update this PR:
- [X] Update the reference pre-submits for release (`@main` vs `@vx.y.z`) - this was already taken care of
- [X] Update the pre-submit that checks which internal Actions are allowed to be called - this was already taken care of

I will add e2e tests in a follow-up PR, including PR support with unsigned attestations https://github.com/slsa-framework/slsa-github-generator/issues/1569


.github/actions/secure-upload-artifact/action.yml
```diff
@@ -21,7 +21,7 @@ runs:
     # `.github/actions/secure-builder-checkout`.
     - name: Compute binary hash
       id: compute-digest
-      uses: ./__BUILDER_CHECKOUT_DIR__/.github/actions/compute-sha256
+      uses: slsa-framework/slsa-github-generator/.github/actions/compute-sha256@main
```
asraa, line 24: is this a temporary change? i thought it was intended for the top-level workflows to check out this repo before calling this action

.github/workflows/delegator_generic_slsa3.yml
```diff
@@ -0,0 +1,257 @@
+# Copyright 2022 SLSA Authors
```
asraa, line 1: nit: 2023

.github/workflows/delegator_generic_slsa3.yml
```diff
@@ -0,0 +1,257 @@
...
+          Optional secret.
+
+          This argument is passed, unchanged, to the builder's Action's `build-artifacts`.
+        required: false
+
+    # Inputs provided as arguments are passed as inputs to the callback Action,
+    # formatted as a map.
+    inputs:
+      slsa-token:
+        description: "The SLSA token identifying the request"
```
asraa, line 50:
```suggestion
        description: "The signed SLSA token identifying the request"
```

we should remember to document carefully the format of the signed SLSA token input: that it's the `<slsa-token>.<signature-bundle>` (or reversed, i forgot)

.github/workflows/delegator_generic_slsa3.yml
```diff
@@ -0,0 +1,257 @@
...
+    # Outputs are provided to the caller TRW.
+    outputs:
+      # This contains all the ouputs of the callback Action, formatted as a map.
+      build-artifacts-outputs:
+        description: "The outputs from the build-artitacts Action, unchanged."
+        value: ${{ jobs.build-artifacts-ubuntu.outputs.outputs }}
+      
+      # This is an output from the framework.
+      attestations-download-name:
+        description: "Name of the artifact to download all the attestations."
```
asraa, line 63: Either in the comment or you can defer to some docs that this folder will contain `.intoto` attestations when run in pre-submit without token access, or signed attestations `intoto.sigstore`.

.github/workflows/delegator_generic_slsa3.yml
```diff
@@ -0,0 +1,257 @@
...
+    steps:
+      - name: Verify token with test action
+        id: verify
+        uses: slsa-framework/slsa-github-generator/.github/actions/verify-token@main
+        with:
+          slsa-workflow-recipient: "delegator_generic_slsa3.yml"
+          slsa-unverified-token: ${{ inputs.slsa-token }}
+          output-predicate: ${{ env.SLSA_PREDICATE_FILE }}
+
+      - name: Debug output predicate
```
asraa, line 96: since we upload the predicate i think we can remove the debug step

.github/workflows/delegator_generic_slsa3.yml
```diff
@@ -0,0 +1,257 @@
...
+        run: |
+          cat "$PREDICATE_FILE" |  jq -r
+          echo "TOKEN: $TOKEN"
+          echo "RUNNER: $RUNNER"
+
+      - name: Upload predicate
+        id: upload
+        uses: slsa-framework/slsa-github-generator/.github/actions/secure-upload-artifact@main
+        with:
+          name: "${{ needs.rng.outputs.value }}-${{ env.SLSA_PREDICATE_FILE }}"
```
asraa, line 110: in the other workflows we suffix with RNG, are you deviating from this on purpose? (I like this better, because it preserves the file extension)

.github/workflows/delegator_generic_slsa3.yml
```diff
@@ -0,0 +1,257 @@
...
+          path: ${{ env.SLSA_PREDICATE_FILE }}
+
+  # privacy-check verifies that the user has agreed for their repository name to be made public, via the rekor log.
```
asraa, line 113: do we not need detect-env to be performing that? or is that part of the TODO to update refs?

.github/workflows/delegator_generic_slsa3.yml
```diff
@@ -0,0 +1,257 @@
...
+
+      - name: Sign attestations
+        id: sign
+        uses: slsa-framework/slsa-github-generator/.github/actions/sign-attestations@main
+        with:
+          attestations: attestations
+          output-folder: slsa-attestations-${{ needs.rng.outputs.value }}
+
+      # NOTE: we don't need to perform a secure upload, since the attestations are signed.
+      - name: Upload attestations
```
asraa, line 248: here you can do the TODO where if you don't have OIDC action you could instead add another job that uploads the output of `attestations` to the folder

asraa (commented):
pretty much lgtm!

.github/actions/secure-upload-artifact/action.yml
laurentsimon: I ran into problems due to all the directory moves we do: the post-clean up the runner was doing kept failing, so I made this change. Will be a long-term change. I will update the pre-submits accordingly.

Wdut?

.github/workflows/delegator_generic_slsa3.yml
laurentsimon: is this relevant for public use? the TRW writers only need to know they should provide the slsa token generated with slsa-setup Action, nothing more, I think.

+1 on documenting. How about a README.md under the `setup-token` Action? Maybe we can also rename it to `delegator/init`

.github/workflows/delegator_generic_slsa3.yml
laurentsimon: I'll add this in the description. Relevant for TRW writers. Thanks

.github/workflows/delegator_generic_slsa3.yml
laurentsimon: yes I did it to preserve the extension, you read my mind. Keep like this is fine?

.github/workflows/delegator_generic_slsa3.yml
laurentsimon: If we refer to the Actions by ref, we actually don't need the `detect-env` Action. We just need to be careful to update the ref pre-submits.

.github/workflows/delegator_generic_slsa3.yml
laurentsimon: done. I've kept a few other debug to make sure we can quickly fix problems as we start building the builders.

.github/workflows/delegator_generic_slsa3.yml
laurentsimon: added a TODO.

.github/workflows/delegator_generic_slsa3.yml
laurentsimon: I added a README.md under the setup-token.

asraa: > the TRW writers only need to know they should provide the slsa token generated with slsa-setup Action, nothing more, I think.

README.md under setup-token is good, I think it may be relevant for them if they're curious to inspect the inputs and outputs they're passing in the TRW (like Fredrik had asked at some point).

.github/workflows/delegator_generic_slsa3.yml
asraa: yes! just wanted to be on the same page.

.github/workflows/delegator_generic_slsa3.yml
asraa: Hmm I think I see what you mean. if user calls `builder@tag`, then we have already hard-coded that reusable workflow at that ref to call actions at that tag in the builder repo.

And it was difficult I think to do this before because we would get stale pinned actions and so on during development.

laurentsimon: one problem before was also that we were calling scripts, and scripts cannot be referred by 'tag'. We could probably change the behavior in other workflows now

.github/actions/secure-upload-artifact/action.yml
asraa: Makes sense to me.

asraa (approved):

Issue #1423, opened 2022-12-28
[feature] Add more malicious SLSA subject layout tests to generate-attestations


asraa:
**Is your feature request related to a problem? Please describe.**
Add strict validation to the SLSA subject layout for generate-attestations. 

Validate that malformed subject layouts cause errors.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

laurentsimon:
@asraa can you elaborate on "strict validation"?

asraa:
Yes! I'm fairly certain it doesn't even have to be a valid JSON of the type we have - i don't think javascript enforced that. Meaning, we could produce some arbitrary JSON and have that incorporated as the subjects part which would then make it not a valid intoto statement

laurentsimon:
so iiuc, this line https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/generate-attestations/src/attestation.ts#L70-L72 is what needs further validation, correct?

laurentsimon:
Given that subject is defined in https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/generate-attestations/src/intoto.ts#L17-L20, I think validation means:
- verify the hash algorithm 
- verify the length of its value

The name is arbitrary.

Is the above correct?

asraa:
I think it is this entire piece 
https://github.com/slsa-framework/slsa-github-generator/blob/7f9306a166326f33f0aeb3c1a217cedf3fb981c5/.github/actions/generate-attestations/src/attestation.ts#L42

I'm not sure how the type casting to types.Layout works with errors, but basically the idea is that if I add in test cases with completely bogus JSON, will it fail gracefully?

asraa:
Yes - it was more for robustness for the TRW author to see if they are getting an error.

PR #1581, opened 2023-01-27
feat: check for empty attestation in generate-attestation Action

laurentsimon:


asraa (approved):
thanks! there are definitely more tests to do here :|
Specifically to just call it out, I'm not sure that if the layout has malformed intoto subject that it will complain... there's an open issue for malicious cases here

laurentsimon:
do we have an issue for it?

asraa:
> do we have an issue for it?

we have adding tests to generate-attestation here:
https://github.com/slsa-framework/slsa-github-generator/issues/1423

PR #1582, opened 2023-01-27
fix: workflowPath in provenance is fully qualified

asraa:

The run fails because the `workflowPath` should be fully qualified, and `e2e_this_file` only returns the basename.

cc @laurentsimon


laurentsimon (approved):
Thanks.

PR #1592, opened 2023-01-30
fix: generate-attestations

laurentsimon:


asraa (approved):
+1 on layout as the name.

@ianlewis thinks so too?

PR #1595, opened 2023-01-31
fix: fix verify-token predicate values check

asraa:

One more fix: https://github.com/slsa-framework/slsa-github-generator/actions/runs/4055979822/jobs/6979852459

systemParamters, at least right now, have "values".

(this may change later with v1.0)

### DIFF
--- a/.github/workflows/scripts/e2e-utils.sh
+++ b/.github/workflows/scripts/e2e-utils.sh
@@ -93,7 +93,7 @@ e2e_verify_predicate_v1_buildDefinition_buildType() {
 }
 
 e2e_verify_predicate_v1_buildDefinition_systemParameters() {
-    _e2e_verify_query "$1" "$3" '.buildDefinition.systemParamters."$2".value'
+    _e2e_verify_query "$1" "$3" ".buildDefinition.systemParameters.\"$2\".value"
 }
 
 e2e_verify_predicate_v1_runDetails_builder_id() {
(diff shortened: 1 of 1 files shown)


laurentsimon (approved):

laurentsimon:
@ianlewis `pre-submit e2e docker-based default / build / provenance (pull_request)` is failing

asraa:
Yes -- also don't merge yet, I was testing this and the jq var sub needs those jq variables declared

.github/workflows/scripts/e2e-utils.sh
```diff
@@ -93,7 +93,7 @@ e2e_verify_predicate_v1_buildDefinition_buildType() {
 }
 
 e2e_verify_predicate_v1_buildDefinition_systemParameters() {
-    _e2e_verify_query "$1" "$3" '.buildDefinition.systemParamters.'"$2"
+    _e2e_verify_query "$1" "$3" '.buildDefinition.systemParamters."$2".value'
```
ianlewis, line 96: > Expressions don't expand in single quotes, use double quotes for that.

Yeah, this needs to be fixed. Also parameters was misspelled.
```suggestion
    _e2e_verify_query "$1" "$3" ".buildDefinition.systemParameters.\"$2\".value"
```

ianlewis (approved):

.github/workflows/scripts/e2e-utils.sh
asraa: thank you!

PR #1600, opened 2023-01-31
fix: fix snake case env var

asraa:

These two particular GH env vars are snake case.......


.github/workflows/scripts/e2e-verify.common.sh
```diff
@@ -85,8 +85,8 @@ e2e_verify_common_buildDefinition_v1() {
     e2e_verify_predicate_v1_buildDefinition_systemParameters "$1" "GITHUB_REPOSITORY_OWNER_ID" "$GITHUB_REPOSITORY_OWNER_ID"
     e2e_verify_predicate_v1_buildDefinition_systemParameters "$1" "GITHUB_WORKFLOW_REF" "$GITHUB_WORKFLOW_REF"
     e2e_verify_predicate_v1_buildDefinition_systemParameters "$1" "GITHUB_WORKFLOW_SHA" "$GITHUB_WORKFLOW_SHA"
-    e2e_verify_predicate_v1_buildDefinition_systemParameters "$1" "IMAGE_OS" "$IMAGE_OS"
-    e2e_verify_predicate_v1_buildDefinition_systemParameters "$1" "IMAGE_VERSION" "$IMAGE_VERSION"
+    e2e_verify_predicate_v1_buildDefinition_systemParameters "$1" "IMAGE_OS" "$ImageOs"
```
ianlewis, line 88: This seems to be `ImageOS`?
```suggestion
    e2e_verify_predicate_v1_buildDefinition_systemParameters "$1" "IMAGE_OS" "$ImageOS"
```

ianlewis (approved):
You can shellcheck comments also to ignore the errors/warnings about the variables being set.

*shellcheck assumes non-capitalized variables should be local.

laurentsimon:
Interesting. Shall we respect the original variable case-sensitivity? Wha's GitHub team's view?
/cc @kommendorkapten

asraa:
> Interesting. Shall we respect the original variable case-sensitivity? What's GitHub team's view?

Seems like they also include the var names in the provenance as `IMAGE_OS` and `IMAGE_VERSION`. But yeah, I don't know why those in particular were snake cased

kommendorkapten:
As @asraa said, npm CLI extracts them like this:

```
          IMAGE_OS: env.ImageOS,
          IMAGE_VERSION: env.ImageVersion,
```

I think SLSA generator should do the same, I have also asked if we can get environment variables with a consistent name too so we don't need to transform them. 

Only one concern, would we pose a risk from a reproducible build scenario by transforming them? I think the risk is negligible, as JavaScript/npm is virtually not possible to build reproducible. If it's more for capturing the provenance, I would favour consistency in the variable names.

ianlewis:
> Only one concern, would we pose a risk from a reproducible build scenario by transforming them? I think the risk is negligible, as JavaScript/npm is virtually not possible to build reproducible. If it's more for capturing the provenance, I would favour consistency in the variable names.

`systemParameters` don't have to be 1:1 environment variable names so I think we're fine here as long as there is an ability to understand what parameters like `IMAGE_OS` and `IMAGE_VERSION` meant.

PR #1604, opened 2023-02-02
Update README.md

lakshya8066:
Removed hyperlink since the heading no longer exists.


asraa (approved):
Thanks!

reviewer comment: blog post is still present, but just not under the heading

Issue #1505, opened 2023-01-10
[feature] [byo] Record the GitHub event payload in SLSA provenance


asraa:
**Describe the bug**
Currently the SLSA provenance only allows for string-string ParameterValues. The GitHub event payload is a JSON object. 

While we may be able to flatten it, it's inconvenient. 

Add if the SLSA v1.0 updates to allow objects or more complex JSON types.

asraa:
Also relevant here is the context.

We currently record these fields https://github.com/slsa-framework/slsa-github-generator/blob/76f03fa7e30209f32ac76ce417ddc43ff98af42a/.github/actions/verify-token/src/predicate.ts#L158-L184 in SLSA v1.0 provenance (which will be reused for BYOB as well generically).

Currently, the human readable `ACTOR` and other fields that may have PII are not noted. If we put the full context, we would have to ensure or allow an opt-out model for the included paramters.

My suggestion here is to add, from github context and event context, all "allowlisted" safe and relevant fields (e.g. ensure that `base_ref` is added from `event_payload`). I would rather this than an opt-out model. For e.g.:
   * Remove `owner` subcontext, (we do have `REPOSITORY_OWNER_ID already)
   * Include branch data -- e.g. default branch of repository, ref information (name, type, head, base, any other data in the event context about branch).
 
The context also includes whether the ref was protected. IMO that is important but I remember we have had discussions on whether this information is "public".

cc @laurentsimon @kommendorkapten

laurentsimon:
Good point. There's a lot of value in recording those fields (actor_id, owner_id) as we explained in https://slsa.dev/blog/2022/06/slsa-github-workflows: it allows monitoring for changes for account / repo re-creation.

What if we could record `opii = H(Nonce, _pii_)` where Nonce is a 128-bit nonce / secret that only the builder knows? It would provide privacy but allow for linkability between two attestations, which would allow monitoring for changes.

We could further scope the obfuscated version by repo, with `opii = H(Nonce, repo_name, _pii_)`

laurentsimon:
Sync'ed with @asraa today. Looks like recording a non-human readable version `field_id` is acceptable, so let's ignore my proposal!

asraa:
https://github.com/slsa-framework/slsa-github-generator/issues/1575#issuecomment-1409547881

For the rest of the work of masking sensitive info, see here

Issue #1575, opened 2023-01-26
[feature][byob] Allow TRW to omit / mask certain fields from attestations


laurentsimon:
Fields like usernames should be omitted from the provenance. We can't have them be secrets because it would cause the GitHub masking engine to detect secrets when there is a match with a sha we use for uploading / download artifacts.

laurentsimon:
I propose to have an a new input to the SRW that takes a "list", maybe `mask-inputs: name1, name2, name3`?

Wdut?

ianlewis:
> I propose to have an a new input to the SRW that takes a "list", maybe `mask-inputs: name1, name2, name3`?
> 
> Wdut?
> 
> /cc @asraa @ianlewis

We'll have to instruct TRW authors to use it but yeah, that's probably fine.

laurentsimon:
I'll start adding this feature then.

asraa:
Are these fields in the GitHub context? I worry that we won't have uniform conformence of the produced provenance, which is fine because it's TRW controlled.

But are you thinking our implementation of the npm one will `mask-inputs`, rather than us masking usernames by default?

asraa:
It's a cool solution, just wondering.

Because we can publish the github context and say these fields exist, mask anything you don't like.

laurentsimon:
So there are 2 aspects:
1. sensitive data in GH context that will be masked by default, as agreed on with GH team. This issues does not change that
2. sensitive data that are builder-specific, e.g. a container builder requires a username to log in a registry. Users won't want to expose this. The TRW authors can mark these fields as sensitive so they will be ignore in the provenance - either we omit them entirely, or we mask them a-laGitHub `***`.

Sorry, I was not clear in the original description

laurentsimon:
See comment https://github.com/slsa-framework/slsa-github-generator/issues/1410#issuecomment-1431615223

We need a way to omit the plaintext "masked" inputs from the token, so that they are not recorded by Rekor. I think there's a way to do that

PR #1611, opened 2023-02-06
feat: update slsa v1 provenance and add github event payload

asraa:
* Fixes predicate format to use any for systemParameters and externalParameters
* Add the GH event payload. Related https://github.com/slsa-framework/slsa-github-generator/issues/1505 


.github/actions/verify-token/src/predicate.ts
```diff
@@ -194,20 +185,19 @@ export function createPredicate(
     },
   };
 
+  // Put GitHub event payload into systemParameters.
+  // TODO(1505): Redact sensitive information.
```
ianlewis, line 189: Redacting fields can be covered by #1575 

```suggestion
  // TODO(github.com/slsa-framework/slsa-github-generator/issues/1575): Redact sensitive information.
```

ianlewis (approved):
LGTM

.github/actions/verify-token/src/predicate.ts
asraa: Ah ok - that issue does cover both aspects of default hiding email / PII and allowing extra config.

I'll mark this one as closing 1505 then.

Issue #1566, opened 2023-01-24
[feature] Update the provenance format in the docker-based generator


rbehjati:
**Describe the solution you'd like**
The current provenance format does not quite comply with SLSA provenance v1.0. The following changes are needed.
1. Move config parameters to the `input` mapValue (see https://github.com/slsa-framework/slsa-github-generator/pull/1547#discussion_r1073705487)
2. The `command` is encoded as a JSON string (see https://github.com/slsa-framework/slsa-github-generator/pull/1547#discussion_r1081594741)
3. Restructure external parameters  according to the new format as in this [example](https://github.com/slsa-framework/slsa/blob/3626d484a68075a5e2af8e18c9112273a700bf59/docs/github-actions-workflow/v0.1/example.json#L6).

**Additional context**
See also 
- https://github.com/in-toto/in-toto-golang/issues/197
- https://github.com/slsa-framework/slsa-github-generator/pull/1547
- https://github.com/slsa-framework/slsa-github-generator/issues/1191

asraa:
Thanks for this issue! I was just thinking about this. OK if I make this update?

I'm also thinking when in-toto is ready to make a PR we can probably move the definitions here upstream.

Only still confused about the `oneof` representation. 

For generated proto, it uses interfaces -- so something like that can work, but I'd like to not introduce proto runtimes into the deps

rbehjati:
> Thanks for this issue! I was just thinking about this. OK if I make this update?

Yes. Sure. Thanks :)

> I'm also thinking when in-toto is ready to make a PR we can probably move the definitions here upstream.
> 
> Only still confused about the `oneof` representation.
> 
> For generated proto, it uses interfaces -- so something like that can work, but I'd like to not introduce proto runtimes into the deps

I don't have any experience with using protobufs in Go myself, but had a chat with @tiziano88, and his suggestion was that if interoperability with other languages and services is needed, then code generation from protobuf is preferable. 

But I agree with you about runtime dependencies. Is it possible to start with a hand-written definition and replace it with auto-generated code in the future if there is need for it?

PR #1615, opened 2023-02-08
fix: update docker-based predicate type to use updated format

asraa:

Partial fix https://github.com/slsa-framework/slsa-github-generator/issues/1566

Updates the docker-based build definition's external parameter structure. This is intended to hold the inputs to a docker-based build (which includes the unpacked command and artifactPath).

The issue also mentions added workflow inputs: these will be added by the workflow, and capture the inputs to the workflow run at the top level. Because they are read from GH context, I think they should be in systemParameters as trusted values alongside the rest of the GH context and workflow path.

cc @rbehjati


laurentsimon (approved):

asraa:
Going to submit, but leaving issue open for review of completion.

rbehjati:
Nice. Thanks :)

Issue #1470, opened 2023-01-05
[feature] Sharing typescript code between GitHub Actions


asraa:
**Is your feature request related to a problem? Please describe.**

Some GitHub Actions require sharing typescript code between the packages. For e.g. predicate structs, shared path utility code.

**Describe the solution you'd like**
It may be possible to require and add packages from local directories outside the package.

If anyone has ideas please let me know!

asraa:
I was reading up on this and saw something that might lead me in the right direction: https://stackoverflow.com/questions/55753163/package-json-is-not-under-rootdir

It seems like we can create a `tsconfig.json` at the top level of our repository, and then have each of our sub-projects have their own `tsconfig.json`. This will allow sharing code between other folders so long as they're under the top-level rootDir.

WDYT?

asraa:
This means we can also separate our common `lib/` code from all the actions.

E.g. structure like `.github/actions/lib/package.json`

Issue #1513, opened 2023-01-11
[feature] harden file writes


laurentsimon:
The builders take untrusted path from projects and use them to create files. We do have code in place to detect directory traversal, but it's easy to forget to use the functions. Directory traversals don't remove all risks, because a project could attempt to overwrite an existing file that belongs to the builder.
I would like to propose to harden our implementation the following way:
1. create a file, say `safe-file.[go|ts]`. 

Have a 
```
func safeFileWrite() error{
  verifyDirTraversal()
  writeIfNotExits()
}
```

2. All file writes should use this function in the code
3. Add pre-submit to verify that only the safe-file directly calls the language-specific `fs.Write()`.

NOTE: the above is similar to what we do in the workflow: all download must use `secure-artifact-download` and we have pre-submit in place to verify it.

laurentsimon:
We should also capture `read` and `mkdir`.
I'll wait for https://github.com/slsa-framework/slsa-github-generator/pull/1557/ to land first

ianlewis:
> We should also capture `read` and `mkdir`. I'll wait for #1557 to land first

We could just forbid other modules from importing `fs` and have the hardened module proxy any file system interactions. That's probably the easiest thing to check for anyway.

laurentsimon:
Sounds like a good idea to catch all fs calls, so long as we don't need other functions like "isFilePresent", etc.
I'll wait to be sure https://github.com/slsa-framework/slsa-github-generator/pull/1612/files works properly, since it's the same pattern we need here to share a module. @asraa I think you ran into problems when trying to share a module: remember what the problems were?

ianlewis:
> Sounds like a good idea to catch all fs calls, so long as we don't need other functions like "isFilePresent", etc.

If we do, we can always just create a proxy function to `fs.isFilePresent` in the "approved" module.

> @asraa I think you ran into problems when trying to share a module: remember what the problems were?

I think she mentioned something about her local setup. I'll take a look at it tomorrow and see if I can't get it working since she'll be off for the next week.

PR #1616, opened 2023-02-08
feat: add GH info into docker-based predicate

asraa:

This adds the following to the docker based build predicate:
* GH context and workflow into the systemParameters
* adds the builder ID from `detect-env`
*  adds the builder binary reference as a `resolvedDependency` 
* adds the invocation ID

The relevant funcs /types in `github.ts` and `predicate.ts` are copies from https://github.com/slsa-framework/slsa-github-generator/pull/1612, so at least I have a POC of how to replace and clean-up this code to share between the actions, but until I can correctly generate typescript libraries, progress is stalled on code cleanup. See https://github.com/slsa-framework/slsa-github-generator/issues/1470


.github/actions/detect-workflow/action.yml
```diff
@@ -21,6 +21,9 @@ outputs:
   ref:
     description: The current workflow reference
     value: ${{ steps.detect.outputs.ref }}
+  workflow:
```
asraa, line 24: note for review: i've added this, because i'd like to consolidate the number of places that we have to fetch an OIDC token in order to detect the reusable workflow. this will help if GitHub ever exposes this in a workflow context to remove the restrictions on `pull_request`.

(In the delegator, we use the TRW as the builder.id, which we get from the SLSA token.)

laurentsimon: in the delegator, we don't even call detect-env, correct?

.github/actions/detect-workflow/action.yml
```diff
@@ -21,6 +21,9 @@ outputs:
...
+    description: The full current job workflow
```
laurentsimon, line 25: nit: maybe explain why is meant by `full`, with an example

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -350,17 +350,15 @@ jobs:
...
-      #     binary-name: "internal/builders/docker-based/cmd/build"
-      #     binary-repository: "${{ needs.detect-env.outputs.repository }}"
-      #     binary-ref: "${{ needs.detect-env.outputs.ref }}"
+      - name: Create predicate
+        id: predicate
+        uses: ./__BUILDER_CHECKOUT_DIR__/internal/builders/docker/actions/create-predicate
+        with:
+          build-definition: "${{ needs.generate-build-definition.outputs.build-definition-name }}"
+          binary-sha256: "${{ needs.generate-builder.outputs.builder-binary-sha256 }}"
+          binary-uri: "git+https://github.com/${{ needs.detect-env.outputs.repository }}@refs/heads/${{ needs.detect-env.outputs.ref }}"
```
laurentsimon, line 359: `binary-uri` is the binary of the builder, correct?

laurentsimon: ignore my question, it's answered inn the action.yml :-)

internal/builders/docker/actions/create-predicate/src/main.ts
```diff
@@ -0,0 +1,94 @@
+import * as core from "@actions/core";
```
laurentsimon, line 1: Do you think we should move this action under the `.github/actions`? I think they are internal too. If we add tests for all actions (e.g., https://github.com/slsa-framework/slsa-github-generator/issues/1513), it would avoid missing this action.

laurentsimon (approved):

asraa:
Sorry about auto-merge, I will address in a second PR!

laurentsimon:
created https://github.com/slsa-framework/slsa-github-generator/issues/1622 for tracking

.github/actions/detect-workflow/action.yml
asraa: HMM! You're right - we populate the builder ID via the tool's slsa-token, so we never need to introspect. Perfect.

PR #1619, opened 2023-02-09
fix: fix delegator predicate's source parameters

asraa:

The source parameter should be an `ArtifactReference`: so ensure that (and remove the extraneous "artifact" wrapping, this was probably vestigial).


.github/actions/verify-token/dist/index.js
```diff
@@ -320,6 +320,12 @@ function createPredicate(rawTokenObj, toolURI, currentRun) {
     var _a;
     const { env } = process;
     const callerRepo = createURI(env.GITHUB_REPOSITORY || "", env.GITHUB_REF || "");
+    const sourceRef = {
+        uri: callerRepo,
```
laurentsimon, line 324: I think the current v1.0 draft a dedicated `ref` field as well https://github.com/slsa-framework/slsa/blob/main/docs/github-actions-workflow/examples/v0.1/example.json#L16

laurentsimon: ignore this comment, I confused it with the `workflow` field.

laurentsimon (approved):

.github/actions/verify-token/dist/index.js
asraa: Speaking of, I do plan on doing a thorough review of the provenance formats when we finalize them. I'm still not totally sure we have the best representation.

PR #1621, opened 2023-02-10
fix: Add verify subcommand

### LINKED ISSUE #1618: [feature] New verify command in the Docker-based builder
Currently to use the Docker-based builder only has `build` and `dry-run` commands that take as input the source, the docker imager, etc. We'd like to have a new command (e.g., `verify`) that takes as input in generated BuildDefinition or provenance, builds the binary the same way the `build` command does, and verifies that the generated binary has the same digest specified in the provenance.

**Describe alternatives you've considered**
Alternatively we could just pass in the BuildDefinition, but that does not contain the binary digest and cannot be used to do a meaningful verification.

**Additional context**
This could be used locally to verify that the provenance generated on GitHub is valid.

rbehjati:
Fixes #1618 

Adds a `verify` subcommand as described in #1618. Fixes and updates the README and testdata used in the README and examples.

### DIFF
--- a/internal/builders/docker/testdata/build-definition.json
+++ b/internal/builders/docker/testdata/build-definition.json
@@ -2,9 +2,9 @@
     "buildType": "https://slsa.dev/container-based-build/v0.1?draft",
     "externalParameters": {
         "source": {
-            "uri": "git+https://github.com/project-oak/transparent-release",
+            "uri": "git+https://github.com/slsa-framework/slsa-github-generator",
             "digest": {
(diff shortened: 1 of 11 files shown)


laurentsimon (approved):

internal/builders/docker/testdata/build-definition.json
```diff
@@ -15,7 +15,11 @@
         },
         "buildConfig": {
```
laurentsimon, line 16: note: I think buildConfig is from slsa specs v0.2, not v1.0?

PR #1624, opened 2023-02-10
refactor: move create-predicate to internal actions

asraa:

Fixes https://github.com/slsa-framework/slsa-github-generator/issues/1622

Also updates in-toto library (for a fixup in the predicate generated in the workflow too)


.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -352,7 +352,7 @@ jobs:
 
       - name: Create predicate
         id: predicate
-        uses: ./__BUILDER_CHECKOUT_DIR__/internal/builders/docker/actions/create-predicate
+        uses: ./__BUILDER_CHECKOUT_DIR__/.github/actions/create-predicate
```
laurentsimon, line 355: will this Action be used by other builders? if not, maybe name it `create-predicate-docker-based` of something of the sort?

laurentsimon (approved):

.github/workflows/builder_docker-based_slsa3.yml
asraa: AH! true. yes. 

eventually we can refactor to one where a BYO builder provides the build definition and inputs some resolved deps under the hood, but yes, will update name.

PR #1633, opened 2023-02-13
chore: Renovate: Only upgrade Go patch versions

ianlewis:
Update Renovate config to only update Go patch versions as minor versions need more care for upgrades. Docker image updates also happen even without a version upgrade.

Reformatted JSON back to use output of `prettier`.


asraa (commented):
Sorry if I don't totally understand - will this remove an alert that a minor version update occurred?

ianlewis:
> Sorry if I don't totally understand - will this remove an alert that a minor version update occurred?

I think so. The problem I'm trying to solve is that renovate sends PRs to upgrade the minor version if one exists rather than patch versions. However, minor versions of Go can have significantly different behavior so I don't really want to upgrade them casually (I'd prefer to upgrade go.mod, Dockerfiles etc. at the same time).

Maybe just configuring it to group the Dockerfile and go.mod Go updates into one PR is good enough?

ianlewis:
Updated to allow minor upgrades but group the Dockerfile and Go updates in go.mod.

asraa (approved):
Ohhh got it - i think it's on the Go version itself, not on Go packages. Makes sense - I agree handling the go base images and go version together are best and minor ones are difficult to handle

PR #1644, opened 2023-02-15
feat: record the BYOB inputs as externalParameters

laurentsimon:
@asraa  OK with this change? Is it fine to start with BYOB support?


.github/actions/verify-token/src/predicate.ts
```diff
@@ -138,12 +137,25 @@ export function createPredicate(
...
+        inputs: rawTokenObj.tool.inputs,
+        // Variables are always empty for BYOB / builders.
+        // TODO(#1555): add support for generators.
+        vars: {},
         // NOTE: This is equivalent to the v0.2 entryPoint.
-        workflowPath: currentRun.path,
+        workflow: {
+          ref: env.GITHUB_REF,
+          repository: env.GITHUB_REPOSITORY,
+          // TODO(#1643): exrtact path from GITHUB_WORKFLOW_REF.
```
ianlewis, line 156:
```suggestion
          // TODO(#1643): extract path from GITHUB_WORKFLOW_REF.
```

.github/actions/verify-token/src/predicate.ts
```diff
@@ -138,12 +137,25 @@ export function createPredicate(
...
+        // TODO(#1575): mask inputs.
+        inputs: rawTokenObj.tool.inputs,
+        // Variables are always empty for BYOB / builders.
+        // TODO(#1555): add support for generators.
+        vars: {},
         // NOTE: This is equivalent to the v0.2 entryPoint.
-        workflowPath: currentRun.path,
+        workflow: {
+          ref: env.GITHUB_REF,
+          repository: env.GITHUB_REPOSITORY,
```
ianlewis, line 155: I guess this should be the same but should this info not be taken from the token? Maybe from `rawTokenObj.github.env.XXX`

ianlewis (approved):
LGTM but let's wait for @asraa to comment.

.github/actions/verify-token/src/predicate.ts
laurentsimon: Correct, these values should be the same, since they are part of the same run.

laurentsimon: NOTE: we use these values in https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/verify-token/src/predicate.ts#L128-L131 to populate the source as well.

ianlewis: Yeah,
https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/verify-token/src/main.ts#L256-L266

laurentsimon: yes, it should, see function `validateGitHubFields` in main.ts

laurentsimon: Shall we add a comment in the code?

ianlewis: I think what we should be really doing is validating that the data on the token object is correct prior to this, and using data on the token rather than data that may or may not be validated. It *should* be the same run but I think it could be written in a safer way.

Not a major issue right now though so I'll send a follow up PR myself.

.github/actions/verify-token/src/predicate.ts
```diff
@@ -138,12 +137,25 @@ export function createPredicate(
   };
 
   const payload = github.context;
+  // NOTE: see example at https://github.com/slsa-framework/slsa/blob/main/docs/github-actions-workflow/examples/v0.1/example.json.
   const predicate: SLSAv1Predicate = {
     buildDefinition: {
       buildType: DELEGATOR_BUILD_TYPE,
       externalParameters: {
+        // Inputs to the TRW, which define the interface of the builder for the
```
asraa, line 145: +1 I'm in agreement that because this is provenance for a tool action invocation, that the "top-level inputs" we consider here are the TRW inputs. It took me a while to reconcile that (because technically they are "derived", but I think we should take the stance that the BUILD is for the TRW invocation, not for a higher level workflow invocation)

.github/actions/verify-token/src/predicate.ts
```diff
@@ -194,12 +206,6 @@ export function createPredicate(
     predicate.buildDefinition.systemParameters.GITHUB_EVENT_PAYLOAD = ghEvent;
   }
 
-  // Add workflow inputs to top-level externalParameters.
```
asraa, line 197: These may belong somewhere else in `systemParameters` (as trusted env parameters from GH context) but it's not critical so let's defer.

asraa (approved):

PR #1648, opened 2023-02-15
docs: Add changelog entry for registry-username secret

ianlewis:
Signed-off-by: Ian Lewis ianlewis@google.com


asraa (approved):

CHANGELOG.md
```diff
@@ -86,6 +86,7 @@
 #### New Features
 
 - A new boolean [`continue-on-error`](https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/container/README.md#workflow-inputs) input was added which, when set to `true`, prevents the workflow from failing when a step fails. If set to true, the result of the reusable workflow will be return in the [`outcome`](https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/container/README.md#workflow-outputs) output.
+- A new [`repository-username`](https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/container/README.md#workflow-inputs) secret input was added to allow users to pass their repository username that is stored in a [Github Actions encrypted secret](https://docs.github.com/en/actions/security-guides/encrypted-secrets).
```
laurentsimon, line 89: this will make the workflow fail if the username has low entropy, because scorecard scanning will find it in other strings like our hashes. Can we be more prescriptive about which option to use? For example, can we say that the secret option should only be used with high-entropy secrets and give an example of cloud providers where it's suitable? I think someone had mentioned the AWS username have high entropy.

In the case where username should not be a secret but users still want to avoid commiting it to the source code of the repo, we should tell users to use https://github.blog/2023-01-10-introducing-required-workflows-and-configuration-variables-to-github-actions/

Wdut?

ianlewis: Thanks. That's a good call. I'll add a note about it here and more details in the docs on workflow inputs that it links to.

I assume the entropy issue is not that GitHub checks the entropy but just that something with low entropy will match hashes etc. more often?

laurentsimon: Correct, hat's my understanding as well.

PR #1654, opened 2023-02-15
feat: upload generated artifacts from docker-based build

asraa:
This PR uploads the generated artifacts by the docker run command to GH artifacts.

It adds an output folder to the build command, since the build command can work in a temporary checkout and the artifacts would be lost. We write to a given path, a temp dir in the workflow, that is later uploaded to GH artifacts. 
Ensures that the path does not already exist, and does not traverse outside the `output-folder` specified.

In a follow-up, I'll upload to a GH release based on whether options `upload-assets` and `upload-tag-name` are set.

cc @rbehjati


.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -318,6 +323,14 @@ jobs:
           name: "${{ steps.build.outputs.slsa-outputs-name }}-${{ needs.rng.outputs.value }}"
           path: "${{ steps.build.outputs.slsa-outputs-name }}"
 
+      - name: Upload output artifacts
+        id: upload-artifacts
+        uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
+        with:
+          name: ${{ env.BUILD_OUTPUTS_FOLDER }}
+          path: /tmp/build-outputs-$${{ needs.rng.outputs.value }}
```
laurentsimon, line 331: nit: there are 2 `$` in the name

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -318,6 +323,14 @@ jobs:
           name: "${{ steps.build.outputs.slsa-outputs-name }}-${{ needs.rng.outputs.value }}"
           path: "${{ steps.build.outputs.slsa-outputs-name }}"
 
+      - name: Upload output artifacts
```
laurentsimon, line 326: no need for a secure upload since the predicate file is securely uploaded, hence verification would fail. Correct?

internal/builders/docker/commands.go
```diff
@@ -87,20 +92,24 @@ func BuildCmd(check func(error)) *cobra.Command {
 			// Remove any temporary files that were generated during the setup.
 			defer db.RepoInfo.Cleanup()
 
-			artifacts, err := db.BuildArtifacts()
+			// Build artifacts and write them to the output folder.
+			artifacts, err := db.BuildArtifacts(absoluteOutputFolder)
```
laurentsimon, line 96: not sure where we verify the `absoluteOutputFolder`. This input is always trusted and provided by the reusable workflow e control? Maybe add a comment. Also you could verify that it starts with `/tmp/`.

laurentsimon (approved):

internal/builders/docker/commands.go
asraa: Yeah, it is always provided by the reusable workflow to internally pass the built artifacts. 
The only verification that needs to happen is that it doesn't overwrite any existing files I think, because it may not be in the current working directory. That check happens when we write a path.

> Maybe add a comment. Also you could verify that it starts with /tmp/.

Yeah, I think this might be the better case. It was really hard to think of exactly what verification was needed here. But really I need a tmp out.

.github/workflows/builder_docker-based_slsa3.yml
asraa: yeah, the idea here was that the slsa layout was uploaded. but thinking about it more, I think I will secure upload it. I'll need to download them and upload it to a release in a later job.

asraa: I'm filing a follow-up, secure-upload doesn't support folders to compute sha256

PR #1660, opened 2023-02-16
refactor: Refactor verify-token validation

### LINKED ISSUE #1410: [feature] Re-evaluate if we need to record the GH context in setup-token Action

### LINKED ISSUE #1411: [feature] Record repo and owner id in setup-token Action
repository_id and repository_owner_id are not available in env variables. If they become, we can record them.
For now, we'll have access to them only thru the OIDC / Fulcio cert

### LINKED ISSUE #1643: [feature] remove call to getWorkflowRun API
The GH env variable contains a `GITHUB_WORKFLOW_REF` `octocat/hello-world/.github/workflows/my-workflow.yml@refs/heads/my_branch`.

I think we can parse this variable to set the workflow https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/verify-token/src/predicate.ts#L146

/cc @asraa

ianlewis:
- Fixes #1410
- Fixes #1411
- Fixes #1643 

Various refactors to verify-token

- Removes the call to getWorkflowRun API
- Refactors and cleans up validation code in `verify-token`.
- Improves the GitHub values included in the token in `setup-token` and validation of those values in `verify-token`.
- Adds some unit tests to the `verify-token` action.

Signed-off-by: Ian Lewis ianlewis@google.com


.github/actions/verify-token/__tests__/validate.test.ts
```diff
@@ -0,0 +1,247 @@
...
+  it("does not validate githubObj with empty fields", () => {
+    process.env.GITHUB_ACTOR = "asraa";
+    process.env.GITHUB_ACTOR_ID = "123";
+    process.env.GITHUB_EVENT_NAME = "workflow_dispatch";
+    process.env.GITHUB_EVENT_PATH = "/path/to/event.json";
+    process.env.GITHUB_JOB = "234";
+    process.env.GITHUB_REF = "refs/heads/mybranch";
+    process.env.GITHUB_REF_TYPE = "branch";
+    process.env.GITHUB_REPOSITORY = "ianlewis/byob-test";
+    process.env.GITHUB_REPOSITORY_ID = "456";
```
ianlewis, line 166: This should also be empty so the values are equal
```suggestion
    process.env.GITHUB_REPOSITORY_ID = "";
```

.github/actions/verify-token/src/predicate.ts
```diff
@@ -10,133 +10,51 @@ WIHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-    inputs: Map<string, string | number | boolean>;
-  };
-}
+// getWorkflowPath returns the workflow's path from the workflow_ref.
+export function getWorkflowPath(obj: githubObj): string {
+  // GITHUB_WORKFLOW_REF contains the repository name in the path. We will trim
+  // it out.
+  // e.g. 'octocat/hello-world/.github/workflows/my-workflow.yml@refs/heads/my_branch'
+  const repoPrefix = `${obj.repository}/`;
+  if (!obj.workflow_ref.startsWith(repoPrefix)) {
```
ianlewis, line 32: Previous comment said this should be validated rather than checked here. Move logic to validateGitHubFields

.github/actions/verify-token/src/types.ts
```diff
@@ -0,0 +1,123 @@
...
+  version: number;
+  context: string;
+  builder: {
+    private_repository: boolean;
+    runner_label: string;
+    audience: string;
+  };
+  github: githubObj;
+  runner: runnerObj;
+  image: imageObj;
```
ianlewis, line 111: Should this be under `builder`?

.github/actions/verify-token/src/validate.ts
```diff
@@ -0,0 +1,173 @@
...
+WIHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { githubObj } from "../src/types";
+
+export function validateGitHubFields(gho: githubObj): void {
+  // actor
+  validateFieldNonEmpty("github.actor", gho.actor);
```
laurentsimon, line 18: maybe you could have the 2 functions `validateFieldNonEmpty` and `validateField` in a single function, like `validateFieldEqualNonEmpty`

laurentsimon (approved):

actions/delegator/setup-token/dist/index.js
```diff
@@ -96,22 +96,31 @@ function run() {
                     audience: workflowRecipient,
                 },
                 github: {
-                    // TODO(#1410): Re-evaluate if we need the context.
-                    context: github,
+                    actor: process.env.GITHUB_ACTOR,
```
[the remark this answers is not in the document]
ianlewis: I'd like to keep things in the token that we are adding to provenance (because we can verify them). If we are removing the plaintext actor and repository_owner from the provenance then we could do that.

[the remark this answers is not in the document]
ianlewis: BTW, how does removing the plaintext `actor` help with privacy if the `actor_id` is included?

.github/actions/verify-token/src/validate.ts
ianlewis: `validateField` and `validateFieldNonEmpty` were used elsewhere and I didn't want to change that logic. Though I think that `validateField` always wants to check non-empty as well so maybe adding that check is ok.

.github/actions/verify-token/src/types.ts
ianlewis: Will leave these as is for now.

PR #1667, opened 2023-02-21
feat: Update nodejs builder to use BYOB

ianlewis:
- Depends on #1670 
- Depends on #1660 
- Depends on #1671 

Updates the nodejs builder to use BYOB.

Publishing to npmjs.com will be done on a separate PR.

TODO:
- [x] Create an issue for publishing to npmjs.com
- [x] Remove nodejs builder Go code.
- [x] Merge with of #1660 
- [x] Break out workaround for #1668 into a separate PR
- [x] Break out yamlint config change into separate PR


internal/nodejs-action/action.yml
```diff
@@ -0,0 +1,77 @@
...
+    required: false
+
+runs:
+  using: "composite"
+  steps:
+    # NOTE: the repository is already cloned by the caller, so there's no need to
+    # checkout ourselves.
+
+    - name: Setup Node
+      uses: actions/setup-node@v3
```
laurentsimon, line 32: nit: pin setup-node?

laurentsimon (approved):

internal/nodejs-action/action.yml
ianlewis: Done.

PR #1697, opened 2023-02-27
feat: Add triggering_actor_id to SLSA v1.0 provenance

### LINKED ISSUE #1695: [feature][byob] Record triggering_actor_id
The [SLSA v1.0 RC](https://slsa.dev/github-actions-workflow/v1-rc1#system-parameters) says that the `triggering_actor_id` SHOULD be included in the provenance (if different from the `actor_id`).

I think @MarkLodato generally says fields like this should be left off if they are the same as their "sister" fields (in this case `actor_id`) but is there a reason why we can't just always include it?

Currently the [`github` context](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context) doesn't include the `triggering_actor_id`, only the `triggering_actor` which could be reclaimed by another user. I think the only way we can retrieve it is via the [Workflow Run API](https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2022-11-28#get-a-workflow-run).

ianlewis:
Fixes #1695 

Adds a `GITHUB_TRIGGERING_ACTOR_ID` to the `systemParameters` in SLSA v1.0 provenance.


actions/delegator/setup-token/src/main.ts
```diff
@@ -81,7 +73,6 @@ async function run(): Promise<void> {
         run_id: process.env.GITHUB_RUN_ID,
         run_number: process.env.GITHUB_RUN_NUMBER,
         sha: process.env.GITHUB_SHA,
-        triggering_actor_id: current_run.triggering_actor.id,
```
laurentsimon: is it the same as actor_id?

laurentsimon (approved):

actions/delegator/setup-token/src/main.ts
ianlewis: Most often it's the same, but it could be different. For example, the original actor_id could be you, but if I trigger a re-run of the job you originally triggered, then the triggering_actor_id could be my id for the re-run while actor_id is still you.

PR #1701, opened 2023-02-28
ci: Add merge_group trigger

ianlewis:
Support the new `merge_group` event type and allow PRs to be merged via a merge queue.

See: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue
See: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/merging-a-pull-request-with-a-merge-queue


.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -213,7 +212,7 @@ jobs:
           set -euo pipefail
 
           REF="@${GITHUB_REF}"
-          if [[ "${GITHUB_EVENT_NAME}" == "pull_request" ]]; then
+          if [[ "${GITHUB_EVENT_NAME}" == "pull_request" ]] || [[ "${GITHUB_EVENT_NAME}" == "merge_group" ]]; then
```
ianlewis, line 215: I actually don't think I need this since AFAICT `merge_group` has appropriate permissions.

laurentsimon (approved):
iiuc, this will re-run the workflow when the PR is LGTM'ed but rebased.

ianlewis:
Yeah, 

IIUC 

1. pre-submits in the PR are run
2. If the PR passes pre-submits and is approved it can be added to the merge queue
3. on the merge queue it rebases and runs the pre-submits again
4. If those pass it actually merges

It's worth noting that I believe the `merge_group` event runs with the normal permissions (e.g. it can do `id-token: write` etc. just like a `push` event) since the code has presumably been approved.

laurentsimon:
> It's worth noting that I believe the merge_group event runs with the normal permissions

that's interesting. So a malicious PR author could push some additional commits after review. I think the same attack is possible today, if maintainers just push "rebase" manually.

ianlewis:
> > It's worth noting that I believe the merge_group event runs with the normal permissions
> 
> that's interesting. So a malicious PR author could push some additional commits after review. I think the same attack is possible today, if maintainers just push "rebase" manually.

Nah, I was actually wrong about this. `merge_group` doesn't allow creating tokens.

At least for new commits, PR authors that don't have write access would have their PRs removed from the queue if they push new commits (same as how auto-merge works now).

laurentsimon:
Thanks for the info.

Issue #1703, opened 2023-02-28
[bug] Token validation error


AdamKorcz:
**Describe the bug**
I have the following workflow:
```yaml
jobs:
  demo:
    runs-on: ubuntu-latest
    steps:
    - name: Print something
      run: echo "hello"
  slsa-setup:
    permissions:
      id-token: write # For token creation.
    outputs:
      slsa-token: ${{ steps.generate.outputs.slsa-token }}
    runs-on: ubuntu-latest
    steps:
      - name: Generate the token
        id: generate
        uses: slsa-framework/slsa-github-generator/actions/delegator/setup-token@main
        with:
          slsa-workflow-recipient: "delegator_generic_slsa3.yml"
          slsa-rekor-log-public: ${{ inputs.rekor-log-public }}
          slsa-runner-label: "ubuntu-latest"
          slsa-build-action-path: "./builder"
          slsa-workflow-inputs: ${{ toJson(inputs) }}
      - name: run echo after slsa-setup
        run: echo "echo after slsa-setup"
  slsa-run:
    needs: [slsa-setup]
    permissions:
      id-token: write # For signing.
      contents: write # For asset uploads.
      actions: read   # For the entrypoint.
      packages: write
    uses: slsa-framework/slsa-github-generator/.github/workflows/delegator_generic_slsa3.yml@main
    with:
      slsa-token: ${{ needs.slsa-setup.outputs.slsa-token }}
```

When I run it, I get the following error in the `verify-token` step of the generaic delegator: 
`mismatch github.job: got 'slsa-setup', expected 'verify-token'.`

Which looks like it fails here: https://github.com/slsa-framework/slsa-github-generator/blob/4ed6a517a4eda072ff1fc2395f806a82741400a6/.github/actions/verify-token/src/validate.ts#L35

cc @ianlewis since you recently refactored the `verify-token` validation.

asraa:
Yes, the environment vars for the github job would differ here between the setup where it was created and the job that verifies it - I'll put out a fix, I thikn this check shouldn't be there

AdamKorcz:
Thank you, @asraa. I can confirm that `verify-token` now passes with the above workflow.

PR #1704, opened 2023-02-28
fix: remove job match validation for setup-token

asraa:
Fixes https://github.com/slsa-framework/slsa-github-generator/issues/1703

cc @ianlewis 

verify-token validates that the GITHUB_JOB in the token (produced in a separate job) matches the current env var. This will always fail.

I still include the GITHUB_JOB inside the provenance, but I'm actually thinking it should be the current GITHUB_JOB, not from the token. Let me know for a follow-up.


laurentsimon (approved):

ianlewis:
Thanks for fixing this! apologies for introducing the bug.

Issue #1685, opened 2023-02-24
[bug]


rbehjati:
**Describe the bug**
I am trying to use [builder_docker-based_slsa3.yml]() in Oak (see [this PR](https://github.com/project-oak/oak/pull/3669)), and there is problem in the `generate-builder` step. 

Here is the error ([the Action](https://github.com/project-oak/oak/actions/runs/4263246828/jobs/7419761814)):
```
Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/oak/oak/__BUILDER_CHECKOUT_DIR__/.github/actions/privacy-check'. Did you forget to run actions/checkout before running your local action?
```

I believe the point of this step is to generate the Docker-based builder binary, but the following parameters are passed to the action, which does not seem right:
```
with:
    repository: rbehjati/oak
    ref: bbe9bf9844c845d0b1b2fb2e50c86e2a123c6324
    go-version: 1.19
    binary: slsa-builder-docker-linux-amd64
    compile-builder: true
    directory: internal/builders/docker
    allow-private-repository: false
    token: ***
```

**To Reproduce**
See the PR linked above.

**Expected behavior**
No error.

**Screenshots**
-- 

**Additional context**
--

rbehjati:
cc @asraa 

I wonder if this is related to the issue I had observed before in https://github.com/slsa-framework/slsa-github-generator/pull/1547#discussion_r1085353821.

In any case, we need to separate between two repositories here:
1. The slsa-github-generator repository, and
2. The user repository (in this case Oak)


and make sure both repositories are checked out separately. Is this currently the case?

asraa:
I think this is because on pull_request detect-env doesn't correctly detect the slsa-github-generator repository, and instead detects the oak repository:
```
repository:rbehjati/oak
ref:bbe[9](https://github.com/project-oak/oak/actions/runs/4263246828/jobs/7419741188#step:3:10)bf9844c845d0b1b2fb2e50c86e2a123c6324
```

So I think pull_request only works in our repo, but not in others repos correctly: https://github.com/slsa-framework/slsa-github-generator/issues/124

Let me see, I've noticed a new appearance in GitHub context called `referenced_workflows` that seem to show the reusbale workflows called!
```
  "run_attempt": 1,
  "referenced_workflows": [
    {
      "path": "slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.5.0",
[...]

Issue #37, opened 2022-04-26
Feature: remove dockerfile 


laurentsimon:
See https://github.com/slsa-framework/slsa-github-generator/pull/35/files#r858908103

Something to try.

tiziano88:
Isn't it actually better to have all the hashes explicitly baked in the Docker image though? It seems that relying on github actions would make it harder to track down exactly which versions are being used. GH actions are also basically impossible to run locally, while it's trivial to run a docker container locally that would execute exactly the same steps that CI would also run.

laurentsimon:
From that angle, I think you're right. Actions can be pinned by hash, so I guess we'd need to check how setup-go works under the hood and make sure it does pin what it downloads :-) That may be a good exercise in itself to see if it's do-able in general. 

But I suspect Docker is easier :-)

ianlewis:
I think we should re-write `detect-workflow` to TypeScript since it's essentially building a Docker image each run just so we can make a single HTTP request. We can speed it up a lot by re-writing it to a [JavaScript action](https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action).

laurentsimon:
I'm onboard with removing it and replacing with TS

laurentsimon:
Note that once we have released the generic container workflow, we can improve speed by using a container-based Action, like https://github.com/ossf/scorecard-action/blob/main/action.yaml#L54-L56.

For transparency, we can keep the SLSA attestation in the repository and add pre-submit that the hash of the attestation is the same as the one used by the action's container. We could later use the slsa-verifier when it's ready.

Wdut?

ianlewis:
> Note that once we have released the generic container workflow, we can improve speed by using a container-based Action, like https://github.com/ossf/scorecard-action/blob/main/action.yaml#L54-L56.

Yeah, that's one way to do it. It's a bit more work to set up a place to store the images and the CI for it, but we could probably just use GHA/ghcr.io for that.

laurentsimon:
I was thinking of using ghcr.io for it.

Issue #124, opened 2022-05-30
Get repository, ref via Github API


ianlewis:
Currently we have to have special case code to run e2e tests in pull requests due to #131. I'd like to get rid of that code so that pull requests run normally.

I want to see if I can't get the repository and ref via the Github API rather than by creating an OIDC token. Creating an OIDC token requires `id-token` scope which is not normally given to workflows triggered by the `pull_request` event.

This wouldn't allow us to sign using sigstore's Github provider, but it would at least solve the problem of getting the reusable workflow repo and ref.

laurentsimon:
This was done and can be closed?

ianlewis:
No, this is just an idea I had that I wanted to get the repo and ref via the GitHub API instead of by using `job_workflow_ref` an OIDC token. I'm not sure it's really possible. It's more of a refactor than a feature and it's not really high priority.

laurentsimon:
ho right, my bad.

asraa:
Update here: detect-workflow-js works this way so the remaining item here is to (after ensuring some stability in that new action) to deprecate `detect-workflow` and replace with `detect-workflow-js`

ianlewis:
The old `detect-workflow` action was removed in #1988 and `detect-workflow-js` does this already. Closing.

https://github.com/slsa-framework/slsa-github-generator/blob/f8e470688f4d6523b8afa65618cf5e8a1183fcf4/.github/actions/detect-workflow-js/src/main.ts#L42-L60

ianlewis:
So I think the issue is that this is fixed for `detect-workflow-js` but we only use that for our pre-BYOB builders/generators that were written in Go in order to detect which repo/ref to checkout in order to build the builder binary for pre-submits/e2e tests.

We still need to update the pre-BYOB builder code itself to clean up how we deal with pre-submits and e2e tests.

PR #1706, opened 2023-02-28
feat: add detect-workflow-js to supersede detect-workflow

asraa:
Fixes https://github.com/slsa-framework/slsa-github-generator/issues/37
Fixes https://github.com/slsa-framework/slsa-github-generator/issues/1685

Related: https://github.com/slsa-framework/slsa-github-generator/issues/124

This re-writes the detect-workflow action in Typescript. It currently follows the same initial pattern of looking at the OIDC token, but as a fallback when it is not available, it looks to the `referenced_workflows` in the current workflow run. When there is exactly one of these, we know it must be the current reusbale workflow. So this works OK for our current presubmit tests, as well as for some initial support for general pull_request events in other repositories (that may want to run the build of the workflow but not sign as the docker-based workflow supports this @rbehjati )

Once `github.job_workflow_sha` is available, as it says in the documentation https://docs.github.com/en/actions/learn-github-actions/contexts#github-context, we can ALWAYS use `referenced_workflows`, and match on its SHA. This will allow for deprecation of requiring the ID token for detection.


.github/actions/detect-workflow-js/src/main.ts
```diff
@@ -0,0 +1,118 @@
...
+    core.setFailed("No repository detected.");
+    return;
+  }
+
+  // Set outputs.
+  let repository = "";
+  let ref = "";
+  let workflow = "";
+
+  try {
```
ianlewis, line 42: nit: I would perhaps split these out into functions rather than have a large if/else-like try/catch block.

We could more easily unit test each function that way as well.

ianlewis (approved):
@asraa Did you not want to just replace the `detect-workflow` action?

Could we add a simple scheduled e2e test as well? Kind of like `e2e.verify-token.schedule.yml`?

ianlewis:
@asraa Also, thanks for doing this! I really wanted to do this but just never got around to it.

asraa:
Unit tests with mock OIDC and context complete, and added a scheduled test to test the actual OIDC.

> Did you not want to just replace the detect-workflow action?

yes! I'd like to but I wanted to ensure that this one works with testing in the docker-based workflow.

Issue #1707, opened 2023-03-01
[chore] Use a GitHub merge queue


ianlewis:
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue

A merge queue should help with merging a lot of PRs and making sure the tests are all run against HEAD.

ianlewis:
Initial support for `merge_group` was implemented on #1701

ianlewis:
Looks like CodeQL won't support the merge queue. Uploading results [fails](https://github.com/slsa-framework/slsa-github-generator/actions/runs/4340331986/jobs/7578762989) with

```
Error: ref 'refs/heads/gh-readonly-queue/main/pr-1740-9a7ef41ec68d53579328852d39a790cfa1f258db' not found in this repository
```

ianlewis:
Created an issue on the CodeQL action for the issue
https://github.com/github/codeql-action/issues/1572

PR #1708, opened 2023-03-01
fix: Don't run setup-token on pull_request

ianlewis:
Updates #1707 

This removes the code to set a placeholder signature on pull-requests and removes the pre-submit test. We have [e2e.verify-token.schedule.yml](https://github.com/slsa-framework/slsa-github-generator/blob/0b7c296a7e0f815a893dd2d9f2ce005919d5a6ff/.github/workflows/e2e.verify-token.schedule.yml) that tests the full flow for `setup-token` and `verify-token` will create issues if it has problems.


actions/delegator/setup-token/src/main.ts
```diff
@@ -103,17 +103,11 @@ async function run(): Promise<void> {
     core.info(`unsignedB64Token: ${unsignedB64Token}`);
 
     // Sign and prepare the base64 bundle.
-    const eventName = process.env.GITHUB_EVENT_NAME || "";
```
laurentsimon, line 106: so one problem could be that some users (TRW authors) may want to have some tests on PRs, even if the attestation is not signed. Can we handle this use case?

ianlewis: I think `verify-token` would fail in that case wouldn't it?
https://github.com/slsa-framework/slsa-github-generator/blob/970b0e13262f8912dfe7522b0dbf47b7ea8a7d96/.github/workflows/delegator_generic_slsa3.yml#L92-L98

IIUC, I think this was only there to allow a presubmit on `setup-token` without doing verification.

.github/workflows/e2e.verify-token.schedule.yml
```diff
@@ -28,18 +28,21 @@ jobs:
           slsa-runner-label: "ubuntu-latest"
           slsa-build-action-path: "./actions/build-artifacts-composite"
           slsa-workflow-inputs: '{"name1":"value1","name2":"value2","private-repository":true}'
+      - env:
+          SLSA_TOKEN: ${{ steps.setup.outputs.slsa-token }}
+        run: ./.github/workflows/scripts/schedule.actions/setup-token.sh }}
```
ianlewis, line 33: If I'm going to do this I should merge the logic with verify-token.sh

laurentsimon (approved):

Issue #1702, opened 2023-02-28
[bug] internal error Error: Process completed with exit code 1.


konstruktoid:
**Describe the bug**

https://github.com/konstruktoid/ansible-role-docker-rootless/actions/runs/4291543901/jobs/7476983701#step:6:1

```console
Run set -euo pipefail
  set -euo pipefail
  id=""
  if [[ -n "$RELEASE_ID_NEW_TAGS" ]]; then
    id="$RELEASE_ID_NEW_TAGS"
  elif [[ -n "$RELEASE_ID_TAG_NAME" ]]; then
    id="$RELEASE_ID_TAG_NAME"
  else
    echo "internal error"
    exit 1
  fi
  echo "id=$id" >> "$GITHUB_OUTPUT"
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    BUILDER_BINARY: slsa-generator-generic-linux-amd6[4](https://github.com/konstruktoid/ansible-role-docker-rootless/actions/runs/4291543901/jobs/7476983701#step:6:4)
    BUILDER_DIR: internal/builders/generic
    RELEASE_ID_NEW_TAGS: 
    RELEASE_ID_TAG_NAME: 
internal error
Error: Process completed with exit code 1.
```

**Additional context**

https://github.com/konstruktoid/ansible-role-docker-rootless/actions/workflows/slsa.yml
https://github.com/konstruktoid/ansible-role-docker-rootless/blob/main/.github/workflows/slsa.yml

```sh
$ git diff fcac96092050cfed84150ff4a8a71cc89472f759 .github/workflows/slsa.yml 
diff --git a/.github/workflows/slsa.yml b/.github/workflows/slsa.yml
index 6979597..9f85e95 100644
--- a/.github/workflows/slsa.yml
+++ b/.github/workflows/slsa.yml
@@ -20,7 +20,7 @@ jobs:
         shell: bash
 
       - name: Checkout repository
-        uses: actions/checkout@master
+        uses: actions/checkout@61b9e3751b92087fd0b06925ba6dd6314e06f089 # master
 
       - name: Build artifacts
         run: |
@@ -46,7 +46,7 @@ jobs:
       actions: read
       id-token: write
       contents: write
-    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.4.0
[...]

PR #1712, opened 2023-03-01
fix: Don't try to upload assets without a tag name

ianlewis:
Updates #1702 


asraa (approved):

.github/workflows/builder_go_slsa3.yml
```diff
@@ -329,7 +329,7 @@ jobs:
       contents: write # Needed to write artifacts to a release.
     runs-on: ubuntu-latest
     needs: [build-dry, build, provenance, detect-env]
-    if: inputs.upload-assets == true
+    if: startsWith(github.ref, 'refs/tags/') && inputs.upload-assets
```
laurentsimon, line 332: is this correct? I think we want users to be able to upload assets on non-tag triggers, like workflow_dispatch, so long as they can a tag-name to indicate which tag they us to upload to?
This allows some teams to create a release draft out-of-band and give us the tag.

asraa: Ah, nice catch - agreed with the above, either tag or non-empty tag name. then the individual steps below switch on either case

ianlewis: Yes, you're right I missed this. Updated.

ianlewis:
I'll go ahead and merge this, write e2e tests and point folks to them on the bug.

PR #1722, opened 2023-03-01
fix: fix verify-token test for workflow

asraa:
Attempts to fix https://github.com/slsa-framework/slsa-github-generator/issues/1700

There were a couple of mismatches - `workflow` was not present in the token, we didn't test for `workflow_(ref/sha)` and we also had a few comparisons on actor/repository without the id.


laurentsimon (approved):

actions/delegator/setup-token/src/main.ts
```diff
@@ -76,6 +76,7 @@ async function run(): Promise<void> {
         sha: process.env.GITHUB_SHA,
         workflow_ref: process.env.GITHUB_WORKFLOW_REF,
         workflow_sha: process.env.GITHUB_WORKFLOW_SHA,
+        workflow: process.env.GITHUB_WORKFLOW,
```
ianlewis, line 79: note: I believe I removed this intentionally since it's not a very exact field and we can uniquely identify the workflow from the `workflow_ref` and `workflow_sha`.

asraa: Posting a follow-up, thanks for the feedback. I agree let's remove this - it's usually just he workflow name  anyway

PR #1726, opened 2023-03-02
fix: fix one more verify-token script check

asraa:
I'm starting to think it would be good to run this during pre-submit at some point...

I can't think of the best way beyond a bunch of fakes that are just as hard to maintain

### DIFF
--- a/.github/workflows/scripts/e2e-verify.common.sh
+++ b/.github/workflows/scripts/e2e-verify.common.sh
@@ -68,7 +68,7 @@ e2e_verify_common_all_v1() {
 # $1: the predicate content
 e2e_verify_common_buildDefinition_v1() {
     # This does not include buildType since it is not common to all.
-    e2e_verify_predicate_v1_buildDefinition_externalParameters_workflowPath "$1" "$(e2e_this_file_full_path)"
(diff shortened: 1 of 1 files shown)


laurentsimon (approved):

laurentsimon:
> I'm starting to think it would be good to run this during pre-submit at some point...

Can we have this as pull_request_target and run only if it's one of the maintainers (us)? @ianlewis is also thinking of removing the token generation from pre-submit https://github.com/slsa-framework/slsa-github-generator/pull/1708, and we could run into similar issues

asraa:
> Can we have this as pull_request_target and run only if it's one of the maintainers (us)? Need to think a bit more on the implication of this for the token-id permission. 

I'm thinking it might be worth it as we make predicate changes, at least until v1.0 is finalized and this workflow is also stable.

ianlewis:
> > Can we have this as pull_request_target and run only if it's one of the maintainers (us)? Need to think a bit more on the implication of this for the token-id permission.
> 
> I'm thinking it might be worth it as we make predicate changes, at least until v1.0 is finalized and this workflow is also stable.

Can we create an issue to track this?

Issue #1623, opened 2023-02-10
[cleanup] Cleanup unnecessary artifacts from delegator


asraa:
**Describe the bug**
The slsa layout and predicate file are not needed by the end user, so clean these up since they are only uploaded to pass between jobs.

An API exists here: https://docs.github.com/en/rest/actions/artifacts?apiVersion=2022-11-28#delete-an-artifact

There are some existing actions:
https://github.com/GeekyEggo/delete-artifact

But no official one:
https://github.com/actions/upload-artifact/issues/5

WDYT? Should we use the existing action or re-write out own? It would end up looking very similar.

asraa:
This also applies to the layout and build definition of the docker-based workflows.

PR #1729, opened 2023-03-02
feat: delete internal artifacts in BYO and docker

asraa:
Fixes https://github.com/slsa-framework/slsa-github-generator/issues/1623

The slsa layout and predicate file are not needed by the end user, so clean these up since they are only uploaded to pass between jobs.

### DIFF
--- a/.github/workflows/delegator_generic_slsa3.yml
+++ b/.github/workflows/delegator_generic_slsa3.yml
@@ -244,3 +244,15 @@ jobs:
         with:
           name: "${{ needs.rng.outputs.value }}-slsa-attestations"
           path: "${{ needs.rng.outputs.value }}-slsa-attestations"
+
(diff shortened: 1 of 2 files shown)


laurentsimon (approved):

.github/workflows/delegator_generic_slsa3.yml
```diff
@@ -244,3 +244,15 @@ jobs:
         with:
           name: "${{ needs.rng.outputs.value }}-slsa-attestations"
           path: "${{ needs.rng.outputs.value }}-slsa-attestations"
+
+  # cleanup deletes internal artifacts used by the delegator workflow
+  cleanup:
```
laurentsimon, line 249: no permissions for this? nice

asraa: seems like no! i think it may rely on an ambient token.

Issue #1638, opened 2023-02-13
[feature] [docker] Provenance output name


asraa:
**Is your feature request related to a problem? Please describe.**
Currently the provenance output is uploaded to a folder called `outputs` and then the filename is hardcoded as `attestation1.intoto.sigstore`.  

See this workflow as an example https://github.com/slsa-framework/example-package/actions/runs/4159801908

**Describe the solution you'd like**
I'd like to 
1. Rename the output folder to `slsa-outputs`
2. Cleanup the extraneous files 
3. Rename the filename of the provenance. 

For the final one, we currently use a workflow option `provenance-name` to configure the provenance filename. We may include that here too. Using the `artifact_path` as a default filename is not possible because it may include wildcards.

@rbehjati From a user perspective, what do you think about a default filename? Should we fix it at `attestation.intoto.sigstore`, and allow the user to configure it with a workflow input?

rbehjati:
I think we need to allow the user to specify the name of the provenance. A `provenance-name` workflow option sounds good to me. 

In Oak we generate multiple provenances, and we want to be able to get a separate provenance file for each.

PR #1730, opened 2023-03-02
feat: add a provenance-name input option to the docker builder

asraa:
Fixes https://github.com/slsa-framework/slsa-github-generator/issues/1638

* Adds a `provenance-name` input option to the docker builder to configure the name of the attestation that gets uploaded
* Updates docs
* Renames `provenance-name` output to `attestations-download-name` for consistency
* Fixes a bug when using a `--force-checkout` locally, the build output needs to be absolute to generate the relative path.

cc @rbehjati this will fix an issue I noticed in https://github.com/project-oak/oak/actions/runs/4312859723 where the attestation was using the same name. Also checkout the workflow output documentation for the location of the build/provenance output.

### DIFF
--- a/.github/workflows/builder_docker-based_slsa3.yml
+++ b/.github/workflows/builder_docker-based_slsa3.yml
@@ -68,19 +68,19 @@ on:
         required: false
         type: boolean
         default: false
-        # TODO: Options for requesting and verifying provenance of builder image
-        # This will use slsa-verifier and slsa-verifier option.
-        #      * request-builder-verification
-
+      provenance-name:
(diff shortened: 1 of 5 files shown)


laurentsimon (approved):

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -79,8 +79,12 @@ on:
       build-outputs-name:
```
rbehjati, line 79: Thanks @asraa. Do you want to rename `build-outputs-name` to `build-artifacts-name` too?

Both work for me.

asraa: Follow-up in https://github.com/slsa-framework/slsa-github-generator/issues/1735 !

Issue #1668, opened 2023-02-21
[bug] verify-token: Cannot find module '../../store/public-good-instance-root.json'


ianlewis:
When javascript is compiled using `ncc`, the `public-good-instance-root.json` is not included in the built output and can't be read by `sigstore-js`.

Related https://github.com/sigstore/sigstore-js/issues/301

ianlewis:
I have a working workaround in #1667. I'll split it into a separate PR so we can work around the issue until it's fixed in `sigstore-js`.

ianlewis:
sigstore-js 1.1.1 was published so we should be able to upgrade and then remove the workaround we have put in place.

https://github.com/sigstore/sigstore-js/issues/301#issuecomment-1468817183s

ianlewis:
Fixed by #1838

PR #1738, opened 2023-03-03
feat: Support masked TWR inputs

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/1575


.github/actions/verify-token/dist/dist/index.js
```diff
@@ -64,7 +64,7 @@ function run() {
             /* Test locally. Requires a GitHub token:
                 $ env INPUT_SLSA-WORKFLOW-RECIPIENT="delegator_generic_slsa3.yml" \
                 INPUT_SLSA-UNVERIFIED-TOKEN="$(cat testdata/slsa-token)" \
-                INPUT_TOKEN="$(gh auth token)" \
+                INPUT_TOKEN="$(echo $GH_TOKEN)" \
```
asraa, line 67: any reason to switch this? i found it easiest to rely on the `gh` CLI

.github/actions/verify-token/dist/dist/index.js
```diff
@@ -82,7 +82,7 @@ function run() {
                 GITHUB_REF_TYPE="branch" \
                 GITHUB_ACTOR="laurentsimon" \
                 GITHUB_WORKSPACE="$(pwd)" \
-                nodejs ./dist/index.js
+                nodejs ./dist/dist/index.js
```
asraa, line 85: why is this sub-nested now?

.github/actions/verify-token/src/predicate.ts
```diff
@@ -69,8 +69,7 @@ export async function createPredicate(
       buildType: DELEGATOR_BUILD_TYPE,
       externalParameters: {
         // Inputs to the TRW, which define the interface of the builder for the
-        // BYOB framework.
-        // TODO(#1575): mask inputs.
+        // BYOB framework. Some of these values may be masked by the caller.
```
asraa, line 72: nit: instead of caller, let's use TRW author - technically caller shouldn't be able to manipulate the provenance, right?

.github/actions/verify-token/src/validate.ts
```diff
@@ -98,6 +98,26 @@ export function validateGitHubFields(gho: githubObj): void {
   );
 }
 
+export function validateAndMaskInputs(
```
asraa, line 101: could you add a unit test for this? I think it should be fairly easy

.github/actions/verify-token/dist/dist/index.js
laurentsimon: let me undo it. For some reason it was no long working locally.. could be because I was not authenticated.

.github/actions/verify-token/dist/dist/index.js
laurentsimon: This is a temporary fix for https://github.com/slsa-framework/slsa-github-generator/issues/1668 that @ianlewis sent.

.github/actions/verify-token/src/predicate.ts
laurentsimon: +1, I'll use TRW instead. The TRW can manipulate the provenance if they want today, since they provide the inputs to record. (could change if GH context had the information https://github.com/actions/runner/issues/2274#issuecomment-1456475392)

.github/actions/verify-token/src/validate.ts
laurentsimon: +1, will do.

.github/actions/verify-token/dist/dist/index.js
asraa: ah, thank you!

.github/actions/verify-token/src/validate.ts
```diff
@@ -98,6 +98,34 @@ export function validateGitHubFields(gho: githubObj): void {
...
+    }
+    // Ignore empty keys.
+    if (key === undefined || key.trim().length === 0) {
+      continue;
+    }
+    const value = maskedMapInputs.get(key);
+    // boolean and numbers have default values
+    // so we only check for empty string values.
+    const s = value as string;
+    if (s === undefined || s.length === 0) {
```
asraa, line 119: just curious, but why not mask the ones that are empty? in case the user explicitly passed an empty string, then it should be masked (and we should not know it was empty)

.github/actions/verify-token/src/predicate.ts
laurentsimon: done.

.github/actions/verify-token/src/validate.ts
laurentsimon: I think the motivation was that it's it's empty, there's no sensitive data to mask. The masked fields are for values that are sensitive but low-entropy value, like a username or other PII. I was not trying to avoid side-channel information leak like whether the data was empty or not. Let me know what you think. If we think it's better to keep the field masked all the time, I can do that

/cc @ianlewis

asraa: Yeah, I can see - the usage practically would likely be for hiding an input that we don't want on the public ledger, not to prevent someone inspecting it from, say, knowing whether an input like a username was needed or not.

To me it's simplifies it to always mask, so there's no confusion on why the input appears in the empty case. 

But I don't think it is critical, so I can defer to suggesting we document the empty input case.

laurentsimon: let's wait for @ianlewis feedback and I can make the changes

.github/workflows/e2e.verify-token.schedule.yml
```diff
@@ -57,8 +60,24 @@ jobs:
           slsa-unverified-token: aGVsbG8K
           slsa-workflow-recipient: "delegator_generic_slsa3.yml"
           output-predicate: mismatch-token-predicate.json
+      - id: verify
```
ianlewis, line 63: Give this a unique id?

actions/delegator/setup-token/action.yml
```diff
@@ -42,6 +42,18 @@ inputs:
     type: string
     required: true
 
+  slsa-workflow-inputs-mask:
```
ianlewis, line 45: Why not name it `slsa-workflow-masked-inputs` to be consistent?

ianlewis (approved):

.github/actions/verify-token/src/validate.ts
```diff
@@ -98,6 +98,34 @@ export function validateGitHubFields(gho: githubObj): void {
...
+  const maskedMapInputs = new Map(Object.entries(token.tool.inputs));
+
+  for (const key of token.tool.masked_inputs) {
+    if (!maskedMapInputs.has(key)) {
+      throw new Error(`input ${key} does not exist in the input map`);
+    }
+    // Ignore empty keys.
+    if (key === undefined || key.trim().length === 0) {
+      continue;
+    }
```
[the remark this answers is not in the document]
laurentsimon: updated to throw an error.

actions/delegator/setup-token/action.yml
laurentsimon: works for me.

.github/workflows/e2e.verify-token.schedule.yml
```diff
@@ -57,8 +60,24 @@ jobs:
...
+        with:
+          slsa-workflow-recipient: "delegator_generic_slsa3.yml"
+          slsa-rekor-log-public: true
+          slsa-runner-label: "ubuntu-latest"
+          slsa-build-action-path: "./actions/build-artifacts-composite"
+          slsa-workflow-inputs: '{"name1":"value1","name2":"value2","name3":"value3","name4":"","name5":"value5","name6":"value6","private-repository":true}'
+          slsa-workflow-inputs-mask: name2, name4,name7 # name7 does not exist so it should trigger an error.
       - env:
-          SUCCESS: ${{ steps.verify-mismatch-recipient.outcome == 'failure' && steps.verify-mismatch-token.outcome == 'failure' }}
+          SUCCESS: ${{ steps.verify-mismatch-recipient.outcome == 'failure' && steps.verify-mismatch-token.outcome == 'failure' && steps.setup-invalid-mask.outcome == 'failure' }}
```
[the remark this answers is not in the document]
laurentsimon: it does have a `continue-on-error`

.github/workflows/e2e.verify-token.schedule.yml
laurentsimon: I removed this `uses`, not sure why it was present.

.github/actions/verify-token/src/validate.ts
laurentsimon: made the changes. Thanks!

PR #1762, opened 2023-03-07
fix: Update cosign flag

hiyosi:
This PR fixes workflow failure for generic container.

* `--force` flag is removed by https://github.com/sigstore/cosign/pull/2399
* Use `--yes` instead of `--force`

### DIFF
--- a/.github/workflows/generator_container_slsa3.yml
+++ b/.github/workflows/generator_container_slsa3.yml
@@ -207,7 +207,7 @@ jobs:
 
           COSIGN_EXPERIMENTAL=1 cosign attest --predicate="$predicate_name" \
             --type slsaprovenance \
-            --force \
+            --yes \
             "${UNTRUSTED_IMAGE}@${UNTRUSTED_DIGEST}"
 
       - name: Final outcome


ianlewis (approved):

laurentsimon:
I suggest we "pin" the version of the installer to install the same version via the Action argument to avoid major version updates under the hood https://github.com/sigstore/cosign-installer/blob/main/action.yml#L10

ianlewis:
> This means we need to push a release?
> 
> @ianlewis I just realized, I _think_ `cosign-installer` defaults to the latest cosign version when a version isn't specified. The bump to v2 was very incompatible, and so now anyone using generic_contianer will break. Maybe we should ping cosign-installer moving forward?

It doesn't determine the release to install on every run. It just has a [default `cosign-release`](https://github.com/sigstore/cosign-installer/blob/bd2d1189b064bcddc3903176a807dcdba72d7fd0/action.yml#L13) that they update when there is a new release. The issue was caused by upgrading the action itself on https://github.com/slsa-framework/slsa-github-generator/pull/1741.

asraa:
AH! Thank you for that @ianlewis - looks like we're fine...

ianlewis:
> I suggest we "pin" the version of the installer to install the same version via the Action argument to avoid major version updates under the hood https://github.com/sigstore/cosign-installer/blob/main/action.yml#L10

I did think about this when writing the workflow originally but opted to leave it off so that, effectively, the `cosign-release` would be updated automatically by renovate. We should at least have a comment there that updating the installer could bump the cosign release.

laurentsimon:
added to https://github.com/slsa-framework/slsa-github-generator/issues/1765

PR #1764, opened 2023-03-07
fix: pin cosign installation version

laurentsimon:
Avoid https://github.com/slsa-framework/slsa-github-generator/pull/1762 in the future


laurentsimon:
> LGTM. Definitely agree for this not to pull in latest version.
> 
> One more comment - let's see if we can ask cosign-installer to support a minor patch version like 2.0, in case there are any patch updates? I'll post an issue

SG, I had the same thought. Definitely more complex on their end, though

PR #1768, opened 2023-03-07
TEST feat: add new predicate-utils

asraa:


ianlewis:
Maybe these predicate types can just go in the tscommon library that Laurent added?

PR #1773, opened 2023-03-08
feat: Add support for SLSA v0.2 in verify-token

### LINKED ISSUE #1521: [feature] SLSA v0.2 support for npm builder
Due to the timeline of when the npm SLSA RFC will be implemented it's likely we'll need support for SLSA v0.2 in the npm builder. Is that the case?

Currently we are supporting only v1.0 in the [BYOB Framework](https://github.com/slsa-framework/slsa-github-generator/milestone/11) but we may need to support v0.2 as well?

/cc @asraa @laurentsimon

ianlewis:
Updates #1521

Adds a new `slsa-version` input to `verify-token` which allows TRWs to output a SLSA v0.2 predicate.


.github/actions/verify-token/action.yml
```diff
@@ -5,24 +5,25 @@ description: "Verify a token"
...
 
   output-predicate:
     description: "The filename to place the output SLSA predicate based on the verified token"
-    type: string
     required: true
 
+  slsa-version:
+    description: "The version of SLSA provenance format to output in the predicate."
+    required: false
+    default: "1.0-rc1"
```
asraa, line 25: thank you for the RC change - just FYI we will need to update https://github.com/slsa-framework/slsa-github-generator/blob/80da39e3f22bc13d965dfbccc408098ca2f362fe/.github/workflows/delegator_generic_slsa3.yml#L229

.github/actions/verify-token/action.yml
```diff
@@ -5,24 +5,25 @@ description: "Verify a token"
...
 
   slsa-unverified-token:
     description: "The token to verify"
-    type: string
     required: true
 
   token:
     description: "The GitHub Actions token."
     required: false
-    default: ${{ github.token }}
```
asraa, line 19: qq: why is the default here removed? the code doesn't look for GITHUB_TOKEN as a fallback (but maybe there's some magic i don't know about)

.github/actions/verify-token/src/predicate1.ts
```diff
@@ -14,32 +14,18 @@ limitations under the License.
...
-  // GITHUB_WORKFLOW_REF contains the repository name in the path. We will trim
-  // it out.
-  // e.g. 'octocat/hello-world/.github/workflows/my-workflow.yml@refs/heads/my_branch'
-  // Strip off the repo name and git ref from the workflow path.
-  return obj.workflow_ref
-    .substring(`${obj.repository}/`.length)
-    .split("@", 1)[0];
-}
+const DELEGATOR_BUILD_TYPE_V1 =
+  "https://github.com/slsa-framework/slsa-github-generator/delegator-generic@v1";
```
asraa, line 22: the other file versions the delegator build type at v0, why differ?

asraa (commented):
thank you for doing this!

.github/actions/verify-token/action.yml
```diff
@@ -5,24 +5,25 @@ description: "Verify a token"
 inputs:
   slsa-workflow-recipient:
     description: "The audience of the token"
-    type: string
```
laurentsimon, line 8: is the default type `string`? Can we keep them for readability?

laurentsimon (commented):
qq: shall we expose the SLSA version as an env variable, so that it's not exposed to TRWs but only our workflows internally? Since this seems like a temporary solution for npm, maybe we should not expose this functionality?
Starting with an env variable lets us make it public later if we need. Not exposing lets us deprecate the v0.2 more easily too in the future.

Wdut?

laurentsimon:
> > so that it's not exposed to TRWs but only our workflows internally?
> 
> It will still be exposed only internally to our workflows - the callers of these actions are our internal workflows and we set the inputs, TRW users won't be able to change them

ho, my bad. I read `setup-token` instead of `verify-token`

.github/actions/verify-token/src/predicate1.ts
ianlewis: I thought a bit about this and went this way because it was different SLSA versions, but I guess it doesn't really matter much. I'll just make them both v0.

.github/actions/verify-token/src/slsatypes1.ts
```diff
@@ -0,0 +1,69 @@
...
+export interface RunDetails {
+  builder: Builder;
+
+  metadata: Metadata;
+
+  byproducts?: ArtifactReference[];
+}
+
+// SLSAPredicate is a SLSA v1.0-rc1 provenance predicate.
+export interface SLSAPredicate {
```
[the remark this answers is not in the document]
ianlewis: I'm not really sure that that gets us much since the actual formats are pretty different.

.github/actions/verify-token/action.yml
ianlewis: I'll go ahead and make the change on this PR.

ianlewis:
> > > so that it's not exposed to TRWs but only our workflows internally?
> > 
> > 
> > It will still be exposed only internally to our workflows - the callers of these actions are our internal workflows and we set the inputs, TRW users won't be able to change them
> 
> ho, my bad. I read `setup-token` instead of `verify-token`

We will have to plumb it through the delegator at some point though. This PR leaves that TBD.

.github/actions/verify-token/action.yml
ianlewis: Nope. This was a mistake. That's what we have reviews for I guess 😞

ianlewis:
> qq: shall we expose the SLSA version as an env variable, so that it's not exposed to TRWs but only our workflows internally?

I think we will still want to support an input that allows users to generate different versions of SLSA provenance in order to better support downstream consumers. For example, even if 1.1 is released and we generate that by default users may want to generate 1.0 provenance because the downstream tooling hasn't caught up to 1.1 yet.

We can discuss that separately though. I'll submit a separate PR where we can discuss this.

asraa (approved):

PR #1776, opened 2023-03-08
fix: use RNG for cleanup

asraa:
fixes the cleanup action by directly specifying the RNG value in the cleanup files.

when running in parallel (through multiple jobs or matrix strategy) the cleanup of one can affect the other with the wildcard specifier.

cc @rbehjati


laurentsimon (approved):

.github/workflows/delegator_generic_slsa3.yml
```diff
@@ -248,11 +248,13 @@ jobs:
...
   cleanup:
     runs-on: ubuntu-latest
-    needs: [generate-provenance]
+    needs: [rng, generate-provenance]
+    env:
+      RNG: ${{ needs.rng.outputs.value }}
     steps:
       - uses: geekyeggo/delete-artifact@54ab544f12cdb7b71613a16a2b5a37a9ade990af # v2.0.0
         with:
           name: "*-${{ env.SLSA_PREDICATE_FILE }}"
```
ianlewis, line 257: Do you still want to delete all predicate files?
```suggestion
          name: "${{ env.RNG }}-${{ env.SLSA_PREDICATE_FILE }}"
```

ianlewis (approved):

.github/workflows/delegator_generic_slsa3.yml
asraa: oops good catch - i missed that one.

PR #1777, opened 2023-03-08
fix: Make sure each docker-based job only cleans up its own files

rbehjati:
Ref https://github.com/actions/runner/issues/2475

### DIFF
--- a/.github/workflows/builder_docker-based_slsa3.yml
+++ b/.github/workflows/builder_docker-based_slsa3.yml
@@ -461,18 +461,18 @@ jobs:
   cleanup:
     # Deletes internal artifacts
     runs-on: ubuntu-latest
-    needs: [generate-build-definition, build, provenance]
+    needs: [rng, generate-build-definition, build, provenance]
     env:
       BUILD_DEFINITION_NAME: ${{ needs.generate-build-definition.outputs.build-definition-name }}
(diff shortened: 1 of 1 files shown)


asraa (approved):

.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -461,18 +461,18 @@ jobs:
...
-    needs: [generate-build-definition, build, provenance]
+    needs: [rng, generate-build-definition, build, provenance]
     env:
       BUILD_DEFINITION_NAME: ${{ needs.generate-build-definition.outputs.build-definition-name }}
       SLSA_OUTPUTS_NAME: ${{ needs.build.outputs.slsa-outputs-name }}
     steps:
       - uses: geekyeggo/delete-artifact@54ab544f12cdb7b71613a16a2b5a37a9ade990af # v2.0.0
         with:
-          name: "${{ env.BUILD_DEFINITION_NAME }}-*"
+          name: "${{ env.BUILD_DEFINITION_NAME }}-${{ needs.rng.outputs.value }}"
```
asraa, line 471: I found that this action does not support expanding inputs - so you have to add an env var `RNG: ${{ needs.rng.outputs.value }}` and replace this with `${{ env.RNG }}`

rbehjati: Thanks. Done.

asraa (approved):

Issue #1681, opened 2023-02-24
[feature] Update references checks for BYOB


ianlewis:
Update references checks to check for references to the delegator workflow.

- [pre-release/references.sh](https://github.com/slsa-framework/slsa-github-generator/blob/1ab4b3fedd147dbaa1530ff9c3b59f48ace2bae2/.github/workflows/scripts/pre-release/references.sh)
- [pre-submit.actions/references.sh](https://github.com/slsa-framework/slsa-github-generator/blob/1ab4b3fedd147dbaa1530ff9c3b59f48ace2bae2/.github/workflows/scripts/pre-submit.actions/references.sh)
- [RELEASE.md](https://github.com/slsa-framework/slsa-github-generator/blob/1ab4b3fedd147dbaa1530ff9c3b59f48ace2bae2/RELEASE.md)

laurentsimon:
I think this is already working properly. The scripts search under `.github/workflows/` for the string `slsa-framework/slsa-github-generator/.github/actions/...`, so the delegator workflows will be taken into account.

ianlewis:
Right now I think the scripts miss uses of [delegator_generic_slsa3.yml](https://github.com/slsa-framework/slsa-github-generator/blob/cd68d888a824244d863cc4ba49b17eaa824215b0/.github/workflows/delegator_generic_slsa3.yml).

This will be an issue for any workflows (like the nodejs one) that use BYOB in our repo.
https://github.com/slsa-framework/slsa-github-generator/blob/71cb86986c57cd8173e44c4746c09c3cce7c1c5f/.github/workflows/builder_nodejs_slsa3.yml#L78

laurentsimon:
Got it. I understand what you mean. Agreed.

PR #1778, opened 2023-03-08
feat: check ref to workflows

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/1681

### DIFF
--- a/.github/workflows/scripts/pre-release/references.sh
+++ b/.github/workflows/scripts/pre-release/references.sh
@@ -26,10 +26,11 @@ if [ "$RELEASE_TAG" == "" ]; then
     exit 1
 fi
 
-# Verify internal Actions are referenced by the release tag.
 cd __THIS_REPO__
+
+# Verify internal Actions are referenced by the release tag.
 results=$(
-    find .github/workflows/ -maxdepth 1 -type f -print0 -name '*.yaml' -o -name '*.yml' |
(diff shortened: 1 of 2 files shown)


.github/workflows/scripts/pre-release/references.sh
```diff
@@ -40,6 +40,19 @@ if [[ "$results" != "" ]]; then
     exit 1
 fi
 
+# Verify the internal calls to workflows are referenced by the release tag.
+results=$(
+    find .github/workflows/ -maxdepth 1 -type f -print0 -name '*.yaml' -o -name '*.yml' |
```
ianlewis, line 45: Do we want this to run on all workflows including pre-submits and such?

Maybe we can exclude workflows that start with `pre-submit.` or `e2e.`?

That would still match the `release.yml` and `scorecards.yml` but maybe we can come up with an update naming scheme for them? Something like `ci.tag.release.yml` and `ci.schedule.scorecards.yml` and then ignore the `ci.` prefix?

ianlewis (approved):

.github/workflows/scripts/pre-release/references.sh
laurentsimon: updated the code to only look for `builder_*` files, as they are the only ones that will use the delegator workflows. Wdut?

ianlewis: SG

Issue #1669, opened 2023-02-21
[feature] [nodejs] Publish npm packages to npmjs.com


ianlewis:
We need to publish packages to npmjs.com with the SLSA attestations created by BYOB. `npm publish` currently only allows for generating the SLSA provenance itself.

This means 
1. we need logic to upload to npmjs.com ourselves, or
2. we need to add support for attaching separately generated provenance (sigstore bundle) to the uploaded package to `npm publish`. Maybe something like `npm publish --provenance-bundle bundle.sigstore`.

ianlewis:
Publishing is now partially supported via a [fork of the npm cli](https://github.com/ianlewis/cli/tree/oob-provenance). We will eventually get this merged upstream.

ianlewis:
Created #1897 to track merging support for uploading provenance in the upstream npm cli and using that for publishing.

Issue #1574, opened 2023-01-26
[docs][byob] docs for tool writers


asraa:
**Is your feature request related to a problem? Please describe.**
Start docs aimed towards tool writers to create their own SLSA 3 builder.

Some notable items and example we should provide:
* An example of a composite tool action of an existing action hosted in their repo (this should be referenced as a local action in order to ensure that the action is called at the correct reference)
* Passing write permissions through a `github.token` for various permissions: packages, contents
* Passing secret(s)
* An example tool that produces multiple attestations
* An example TRW that publishes outside of the tool action
* An Example PW which calls in parallel multiple TRW
* The output builder ID referencing their tool reusable workflow
* Verification with slsa-verifier
* Performing authentication (e.g. docker registry, GCP) inside a tool action
* TRW Action outputs of different types: array, json, string
* Private project repositories

All of these will need corresponding e2e tests, maybe under `example-byob` or some other name?

cc @laurentsimon

laurentsimon:
- secrets don't support defaults in reusable workflows. Need to show how to use use defaults `github.token`
- inputs need a type

laurentsimon:
note: I'm doing a Syft PoC at https://github.com/laurentsimon/sbom-action/blob/feat/slsa-trw/.github/workflows/slsa3.yml

ianlewis:
Related to #1552 

I started a template example repo at https://github.com/ianlewis/slsa-byob-template

@asraa @laurentsimon We can maybe move the repo under the SLSA org and start iterating on docs there. wdut?

ianlewis:
With all the docs we might have to write, maybe we should start thinking about setting up a proper website with rendered HTML docs...

laurentsimon:
Do you think GitHub pages would be a good start? 
Do you think we could host the doc in this repo under a different branch with no branch protection enabled? Or ding it on another repo is simpler?

Shall we brainstorm the high-level sections we have in mind before?
0. Intro
 - Terminology
 - Supported triggers
2. Getting started
- I want to wrap an existing Action
- I want to create a compliant builder from scratch
3. Maintenance
 - Use X
 - Use Y
2. Improve my security
 - Generate provenance for my binaries
 - Verify them in Action
 - Follow Scorecard best practice
4. Let your users know
 - Update your README
 - Opt in the compliance program
5. FAQs / Common pitfalls
 - Error msg "bla"
 - 
6. Examples
 - Passing secrets
 - Using GH token

[...]

PR #1787, opened 2023-03-10
feat: Publish to npm registry

ianlewis:
Updates #1669 

Adds support for publishing to the npm registry using a [fork of `npm/cli`](https://github.com/ianlewis/cli) that supports publishing provenance that was generated separately from `npm` itself.

Note that this PR also modifies the BYOB `buildInvocationId` for SLSA v0.2 to `<GITHUB_RUN_ID>-<GITHUB_RUN_ATTEMPT>` because npm seems to validate it.

Example successful published package: https://www.npmjs.com/package/@ianlewis/actions-test/v/0.1.72
Example successful GHA run: https://github.com/ianlewis/actions-test/actions/runs/4486596545/attempts/1


internal/nodejs-action/generate-layout.sh
```diff
@@ -2,21 +2,35 @@
...
+# package name. So the subject will take the form of:
+#
+# With scope:
+#   pkg:npm/<scope>/<name>@<version>
+#
+# Without scope:
+#   pkg:npm/<name>@<version>
+raw_package_scope=$(echo "${PACK_JSON}" | jq -r '.[0].name' | cut -d'/' -f1)
+raw_package_name=$(echo "${PACK_JSON}" | jq -r '.[0].name' | cut -d'/' -f2)
+if [ "${raw_package_name}" == "" ]; then
```
laurentsimon, line 18: Given the examples above, it seems `raw_package_name` would never be empty?
`pkg:npm/<scope>/<name>@<version>` -> `<scope>`
`pkg:npm/<name>@<version>` -> `<name>@<version>`

.github/workflows/builder_nodejs_slsa3.yml
```diff
@@ -91,6 +107,42 @@ jobs:
...
+      # TODO(#1669): Use mainline version of npm
+      # TODO(#1669): Generalize directory path checking. The same logic is needed in internal action.
+      - name: Publish the package
+        id: publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.node-auth-token }}
+          ATTESTATION_DIR: ${{ steps.attestation-download.outputs.download-path }}
+          ATTESTATION_NAME: ${{ fromJSON(needs.slsa-run.outputs.build-artifacts-outputs).attestation-name }}
+          PACKAGE_DIR: ${{ steps.package-download.outputs.download-path }}
+          PACKAGE_FILE_NAME: ${{ fromJSON(needs.slsa-run.outputs.build-artifacts-outputs).package-filename }}
```
laurentsimon, line 138: I think all these env variables are untrusted (either provided by the TRW or the PW), so pre-pending `UNTRUSTED_` to all of us would be preferable?

internal/nodejs-action/generate-layout.sh
```diff
@@ -2,7 +2,35 @@
 
 set -euo pipefail
```
laurentsimon, line 4: thoughts on using TS for this code? Maybe as a follow-up?

internal/nodejs-action/action.yml
```diff
@@ -54,7 +63,8 @@ runs:
     - uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
       with:
         name: package.tgz
-        path: ${{ steps.build.outputs.filename }}
+        # TODO(#1669): Should this have an rng value?
```
asraa, line 66: In case someone runs with a matrix strategy, then I believe multiple artifacts would be uploaded to the name `package.tgz`. The `file-path` is likely package-specific and run-specific, so I don't think you'd have to worry about adding an RNG in the path. 

But, using a hard-coded fixed name for the aritfact may be overwritten by matrix or concurrent runs by the same caller

laurentsimon: I think I agree with Asra. The name is important and needs to be unique across concurrent runs.

internal/nodejs-action/generate-layout.sh
ianlewis: The name from the output of `npm pack` only contains the `<scope>/<package name>`. It's not a purl. The comment above was intended as a general comment about why I'm parsing out the parts of the name, and url encoding it etc.

Maybe I can add more comments to make that more clear.

internal/nodejs-action/generate-layout.sh
```diff
@@ -2,21 +2,35 @@
...
+#
+# The npm package's scope is considered a purl "namespace" and not part of the
+# package name. So the subject will take the form of:
+#
+# With scope:
+#   pkg:npm/<scope>/<name>@<version>
+#
+# Without scope:
+#   pkg:npm/<name>@<version>
+raw_package_scope=$(echo "${PACK_JSON}" | jq -r '.[0].name' | cut -d'/' -f1)
```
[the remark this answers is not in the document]
ianlewis: The version in a purl could be any kind of string. I'm not sure what you're getting at here? Maybe you could elaborate?

internal/nodejs-action/generate-layout.sh
ianlewis: Yeah, I'd like to use TS as much as possible.

It's just easier to use Bash since to use TS I'd need to write yet another GitHub action with all the boilerplate (`action.yml`, `package.json`, `Makefile` etc.)

I could use [`github-script`](https://github.com/actions/github-script) but that suffers from a lot of the same problems as using bash (hard to test, lint etc.)

I added #1849 to track.

internal/nodejs-action/generate-layout.sh
```diff
@@ -20,17 +48,18 @@ alg=$(echo "${integrity_digest}" | cut -d'-' -f1 | tr '[:upper:]' '[:lower:]')
 digest=$(echo "${integrity_digest}" | cut -d'-' -f2- | base64 -d | od -A n -v -t x1 | tr -d ' \n')
 
 # NOTE: the name of the attestation should be configurable.
+attestation_name="attestation.intoto"
```
asraa, line 51: what if it's just the packages `<name>.intoto`?

internal/nodejs-action/generate-layout.sh
```diff
@@ -2,7 +2,35 @@
...
+if [ "${raw_package_name}" == "" ]; then
+    raw_package_name="${raw_package_scope}"
+    raw_package_scope=""
+fi
+# package scope (namespace) is URL(percent) encoded.
+package_scope=$(echo "\"${raw_package_scope}\"" | jq -r '. | @uri')
+# package name is URL(percent) encoded.
+package_name=$(echo "\"${raw_package_name}\"" | jq -r '. | @uri')
+# version is URL(percent) encoded.
+package_version=$(echo "${PACK_JSON}" | jq -r '.[0].version | @uri')
```
laurentsimon, line 27: maybe add a comment that version may be a commit, or a string provided by user

laurentsimon (approved):

.github/workflows/builder_nodejs_slsa3.yml
```diff
@@ -91,6 +107,42 @@ jobs:
...
+      - name: Clone npm fork
+        id: clone-fork
+        run: |
+          set -euo pipefail
+
+          # Clone npm fork.
+          cd "$(dirname "$(which node)")"
+          mkdir -p node_modules
+          cd node_modules
+          git clone https://github.com/ianlewis/cli.git npm
```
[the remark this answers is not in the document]
laurentsimon: unresolving since I don't see the changes and this is an important one

internal/nodejs-action/action.yml
```diff
@@ -22,6 +22,15 @@ inputs:
     type: string
     required: false
 
+outputs:
+  attestation-name:
```
laurentsimon, line 26: should this be called provenance-name or attestation-name?

internal/nodejs-action/generate-layout.sh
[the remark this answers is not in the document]
laurentsimon: the publish attestation has a version, the package.json has one too; so I wanted to have some clarification about the assumption we're making on these fields. Could go in a .md explaining the design and assumptions end-to-end.

.github/workflows/builder_nodejs_slsa3.yml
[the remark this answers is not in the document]
ianlewis: Sorry. I'm in the process of pushing the changes and writing the responses and GitHub doesn't allow you to resolve comments as part of a review.

I pinned the digest and updated it so that it fetches the digest to checkout with --depth 1

internal/nodejs-action/generate-layout.sh
ianlewis: I updated this to use the package `filename` included in the output of `npm pack --json`. I can't just use the package name since it could be something like `@ianlewis/mypackage` if it's a scoped package.

Right now though, this is only used internally in the workflow as the provenance is uploaded as json metadata by `npm publish` so this file isn't actually stored anywhere.

internal/nodejs-action/action.yml
ianlewis: I think `attestation-name` is more accurate since it's the attestation generated by `generate-attestations` and signed by `sign-attestations`.

This is only really used internally by the workflow so it shouldn't be exposed to users anyway.

internal/nodejs-action/generate-layout.sh
ianlewis: Done.

internal/nodejs-action/action.yml
ianlewis: I added an rng value to the uploaded package artifact name.

internal/nodejs-action/generate-layout.sh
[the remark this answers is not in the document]
ianlewis: Ok, AFAIK package versions aren't really validated by npm in any way. They could be any kind of string. I think that the slsa-verifier expects versions to be in semver format and that's perhaps fine, though I don't think our builder should make any assumptions about the version.

ianlewis:
@laurentsimon @asraa I made some more changes to address the comments and tested it on my test repo. I think it should be good to go but wanted to give you a chance to take one more pass before merging.

.github/workflows/builder_nodejs_slsa3.yml
```diff
@@ -71,15 +76,30 @@ jobs:
...
+        id: attestation-download
         uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
         with:
           name: ${{ needs.slsa-run.outputs.attestations-download-name }}
 
-      - name: Download artifact
-        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+      # NOTE: secure-download-artifact ensures that the downloaded file doesn't overwrite an existing file.
+      - name: Download package
+        id: package-download
```
laurentsimon, line 95: so this would mean we'll have to expose the action `secure-download-artifact` to TRWs in the future. Can we instead download insecurely and use the slsa-verifier to verify the artifact is the correct one?

What we'll need is:
1. BYOB support inn slsa-verifier
2. Support for `--source-digest` in slsa-verifier, to prevent someone from replacing the genuine artifact with an older artifact

Wdut?

internal/nodejs-action/action.yml
```diff
@@ -51,10 +68,21 @@ runs:
...
+    - id: sha
+      uses: slsa-framework/slsa-github-generator/.github/actions/compute-sha256@main
+      with:
+        path: ${{ steps.build.outputs.file-path }}
+
+    # rng generates a random number to avoid name collision in artifacts
+    # when multiple workflows run concurrently.
+    - name: Generate random 16-byte value (32-char hex encoded)
+      id: rng
+      uses: slsa-framework/slsa-github-generator/.github/actions/rng@main
```
laurentsimon, line 80: so we'll need to expose this function to TRWs, it seems.
It's an interesting and very important use case, I added it to the BYOB doc list https://github.com/slsa-framework/slsa-github-generator/issues/1574#issuecomment-1405878671

.github/workflows/builder_nodejs_slsa3.yml
asraa: > I think provenance should also be securely downloaded.

This finally only made sense to me since the TRW process should not be influenced by the maintainer, and we need to avoid the potential provenance confusion Laurent mentioned above.

I expect in this case we are only creating 1 attestation file, but the delegator handles a folder of attestations inc ase the layout specified more. The upload:
https://github.com/slsa-framework/slsa-github-generator/blob/ab4b569619e185570320838296f57451d79d3547/.github/workflows/delegator_generic_slsa3.yml#L257

I think it should be fairly easy to support folders in secure upload/download, if we tar the folders before and after download to checksum

laurentsimon: In fact, I think the provenance is the only file that needs to be securely downloaded. The artifact verification can then be done via slsa-verifier. It would fail if the user changed the artifact. This would probably also simplify the Action. So it would look like this:

1. TRW MUST securely download the attestation we return
2. TRW MUST slsa-verifier with `--source-uri=$GITHUB_REPOSITORY --build-id <their-builder-id>` before uploading artifact

laurentsimon: > I think it should be fairly easy to support folders in secure upload/download, if we tar the folders before and after download to checksum

That sounds like a good idea. Do we have an issue for it already?

laurentsimon: Created https://github.com/slsa-framework/slsa-github-generator/issues/1863.

internal/nodejs-action/action.yml
ianlewis: I was thinking a bit about this. I think we should think more about this. I'm not sure we need to provide a lot of "utilities" to TRW authors or we should provide better ways to pass out info about the artifacts & attestations that are uploaded by the internal action.

.github/workflows/builder_nodejs_slsa3.yml
ianlewis: Thanks. I think this is a good approach. I think we can address this on a follow up PR.

internal/nodejs-action/action.yml
ianlewis: I think #1863 covers this pretty well and we won't really need this if we implement that idea.

PR #1793, opened 2023-03-10
feat: Add slsa-version input to setup-token

ianlewis:
Adds a `slsa-version` input to `setup-token`. Once we agree on this approach we can add tests. The initial version will support v0.2 provenance but we can potentially remove before the final release of BYOB. 

In the future, I think we will still want to support an input that allows users to generate different versions of SLSA provenance in order to better support downstream consumers. For example, even if 1.1 is released and we generate that by default users may want to generate 1.0 provenance because the downstream tooling hasn't caught up to 1.1 yet. We can do this via the `slsa-version` input.


asraa (approved):

laurentsimon (approved):
Let's create an issue to discuss / track the removal before byob release.

ianlewis:
> Let's create an issue to discuss / track the removal before byob release.

#1808

ianlewis (commented):
Lost changes for review comments on #1773. I should fix them here.

https://github.com/slsa-framework/slsa-github-generator/pull/1773#discussion_r1129479763
> thank you for the RC change - just FYI we will need to update
> 
> https://github.com/slsa-framework/slsa-github-generator/blob/80da39e3f22bc13d965dfbccc408098ca2f362fe/.github/workflows/delegator_generic_slsa3.yml#L229

https://github.com/slsa-framework/slsa-github-generator/pull/1773#discussion_r1129484096
> the other file versions the delegator build type at v0, why differ?

Issue #1795, opened 2023-03-10
[feature] Enable Go Builder to create GitHub Release as a prerelease


suzuki-shunsuke:
**Is your feature request related to a problem? Please describe.**

Currently, Go Builder (builder_go_slsa3.yml) doesn't support creating GitHub Release as a prerelease.
When we publish a prerelease, we have to create GitHub Release once and change it as a prerelease.
It's undesirable.

**Describe the solution you'd like**

Add an input `prerelease` to Reusable workflow to support creating GitHub Release as a prerelase.
If this input is true, GitHub Release is created as a prerelease.
This input is false by default, so the default behaviour isn't changed.

**Describe alternatives you've considered**

I have no idea.

BTW, [GoReleaser supports `auto`](https://goreleaser.com/customization/release/#github).

```yaml
  # If set to auto, will mark the release as not ready for production
  # in case there is an indicator for this in the tag e.g. v1.0.0-rc1
  # If set to true, will mark the release as not ready for production.
  # Default is false.
  prerelease: auto
```

This feature is useful. But Go Builder uses [softprops/action-gh-release](https://github.com/softprops/action-gh-release) and this action supports only boolean. So it is difficult to support `auto`.
If we consider supporting `auto` in future, the type of `prerelease` should be `string` rather than `boolean` to keep the compatibility.

**Additional context**

asraa:
Completed in the PR above, thanks!

asraa:
Reopen to track 

- [ ] add 1 e2e test
- [ ] Update the documentation
- [ ] Add a changelog entry

asraa:
Test successful at https://github.com/slsa-framework/example-package/releases/tag/v34.0.1 !

PR #1796, opened 2023-03-10
feat: enable Go Builder to create GitHub Release as a prerelease

suzuki-shunsuke:
- https://github.com/slsa-framework/slsa-github-generator/issues/1795

Add an input `prerelease` to Reusable Workflow `builder_go_slsa3.yml`. If this input is true, GitHub Release is created as a prerelease.


.github/workflows/builder_go_slsa3.yml
```diff
@@ -365,6 +371,7 @@ jobs:
         uses: softprops/action-gh-release@de2c0eb89ae2a093876385947365aca7b0e5f844 # v0.1.15
         if: inputs.upload-tag-name != ''
         with:
+          prerelease: ${{ inputs.prerelease }}
```
laurentsimon, line 374: is this one necessary? If the user provides the tag-name, does it mean they already created the release so we need not do anything?

I'm not familiar enough with `softprops/action-gh-release`, I think.

laurentsimon (commented):
Thanks

@ianlewis @asraa we're ok with the name? `prerelease` or `pre-release`?

We'll need to add e2e tests and update the documentation, created https://github.com/slsa-framework/slsa-github-generator/issues/1801 for tracking

We may want a similar feature for generators, but let's wait until someone asks?

haydentherapper:
Yes, though we worked around it: https://github.com/sigstore/timestamp-authority/blob/main/.github/workflows/release.yaml#L59-L87

suzuki-shunsuke:
> we're ok with the name? prerelease or pre-release?

I think `prerelease` is better because `prerelease` is used in several places.

- GitHub API: https://docs.github.com/ja/rest/releases/releases?apiVersion=2022-11-28#create-a-release
- `softprops/action-gh-release`: https://github.com/softprops/action-gh-release#inputs
- GoReleaser: https://goreleaser.com/customization/release/#github

.github/workflows/builder_go_slsa3.yml
suzuki-shunsuke: Fixed. 142ad30dc70bfe8dea2628686c55c081ff7b7817

I don't have a strong opinion about this because I usually don't set `upload-tag-name`.

ianlewis:
> > we're ok with the name? prerelease or pre-release?
> 
> I think `prerelease` is better because `prerelease` is used in several places.
> 
> * GitHub API: https://docs.github.com/ja/rest/releases/releases?apiVersion=2022-11-28#create-a-release
> * `softprops/action-gh-release`: https://github.com/softprops/action-gh-release#inputs
> * GoReleaser: https://goreleaser.com/customization/release/#github

It's "Pre-release" in the UI but `prerelease` is fine if we're being consistent with other actions.

ianlewis (approved):

asraa (approved):

PR #1798, opened 2023-03-10
fix: add error message when using a sha for a workflow

asraa:
cc @rbehjati 

This adds a log message that alerts the user that the workflow is failing due to using a workflow pinned by SHA

@ianlewis @laurentsimon if we ever do want to support referencing by SHA the error would obviously need to be removed.


laurentsimon (approved):

laurentsimon:
looks like we're both hitting the same errors on different PRs, regarding the dist-matrix pre-submit. Not sure what's happening

laurentsimon:
I updated ncc and it seems to fix the pre-submit problem. Rebase and you should get the updated version.

Issue #1809, opened 2023-03-13
[e2e]: docker-based workflow_dispatch main default slsa3 


ianlewis:
Repo: https://github.com/slsa-framework/example-package/tree/main
Run: https://github.com/slsa-framework/example-package/actions/runs/4401076291
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.docker-based.workflow_dispatch.main.default.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.docker-based.workflow_dispatch.main.default.slsa3.yml
Trigger: workflow_dispatch
Branch: main
Date: Mon Mar 13 03:23:32 UTC 2023

ianlewis:
Is the number in the `RUNNER_NAME` some kind of version number? I suppose this test would then fail whenever the version is updated?

```
 GitHub Actions 9 == GitHub Actions 4 :: .predicate.buildDefinition.systemParameters.RUNNER_NAME should be GitHub Actions 4
```

ianlewis:
The check looks like it should be ok. It seems like perhaps a workflow could run jobs on GitHub Actions runners that have different versions and that `RUNNER_NAME` could be different from job to job? That brings into question some of our assumptions about values being the same between the `setup-token` job and the job that runs the delegator workflow.

asraa:
> It seems like perhaps a workflow could run jobs on GitHub Actions runners that have different versions and that RUNNER_NAME could be different from job to job?

Yes, I keep hitting this in my workflows: see https://github.com/slsa-framework/slsa-github-generator/pull/1785

I havent yet seen it in any of the delegator workflows, but maybe we don't have that testing set up yet -- our verify-token (which does check RUNNER_NAME) runs in the same job where it was created: https://github.com/slsa-framework/slsa-github-generator/blob/8762ebb0d785f04513ff2900d2a5c7e771cf81b3/.github/workflows/e2e.verify-token.schedule.yml#L21-L48

Realistically, we should separate the jobs out, to mimic how they would be called in a TRW, IMO cc @laurentsimon

asraa:
RUNNER_NAME is verified in the common build definition scripts: https://github.com/slsa-framework/slsa-github-generator/blob/4df91f822551fe7916cc7ebc513c8f429fd26181/.github/workflows/scripts/e2e-verify.common.sh#L90

(but yes, wouldn't have seen it in actual delegator runs!)

[...]

PR #1815, opened 2023-03-13
fix: fix runner verification of predicate

asraa:
May fix https://github.com/slsa-framework/slsa-github-generator/issues/1809

* RUNNER_NAME and RUNNER_VERSION may differ between the jobs that create the predicate and the one that verifies, so only test for presence of the field.

### DIFF
--- a/.github/workflows/scripts/e2e-utils.sh
+++ b/.github/workflows/scripts/e2e-utils.sh
@@ -84,8 +84,8 @@ _e2e_verify_query() {
 _e2e_verify_presence() {
     local attestation="$1"
     local query="$2"
-    name=$(echo -n "${attestation}" | jq -c -r "${query}")
-    assert_true "${name}" "${query} should be ${expected}"
+    name=$(echo -n "${attestation}" | jq -c -r "${query} | select(type != \"null\")")
+    assert_not_empty "${name}" "${query} should not be empty"
 }
 
 e2e_verify_predicate_v1_buildDefinition_externalParameters_workflow_path() {
--- a/.github/workflows/scripts/e2e-utils.sh
+++ b/.github/workflows/scripts/e2e-utils.sh
@@ -80,6 +80,14 @@ _e2e_verify_query() {
     e2e_assert_eq "${name}" "${expected}" "${query} should be ${expected}"
 }
 
+# _e2e_verify_presence verifies that the result of the given jq query is present.
+_e2e_verify_presence() {
+    local attestation="$1"
+    local query="$2"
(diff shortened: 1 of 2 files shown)


.github/workflows/scripts/e2e-utils.sh
```diff
@@ -80,6 +80,14 @@ _e2e_verify_query() {
     e2e_assert_eq "${name}" "${expected}" "${query} should be ${expected}"
 }
 
+# _e2e_verify_presence verifies that the result of the given jq query is present.
+_e2e_verify_presence() {
+    local attestation="$1"
+    local query="$2"
+    name=$(echo -n "${attestation}" | jq -c -r "${query}")
+    assert_true "${name}" "${query} should be ${expected}"
```
ianlewis, line 88: IIUC `assert_true` is going to check for the JS boolean value `true` (or string value `"true"`). You probably want to check that the value is empty.

Also `jq` prints "null" when the value isn't present so you can add another check for the return value's type != null.

```suggestion
    name=$(echo -n "${attestation}" | jq -c -r "${query}  | select(type != \"null\")")
    assert_not_empty "${name}" "${query} should be ${expected}"
```

.github/workflows/scripts/e2e-verify.common.sh
```diff
@@ -84,11 +84,13 @@ e2e_verify_common_buildDefinition_v1() {
...
-    e2e_verify_predicate_v1_buildDefinition_systemParameters "$1" "IMAGE_VERSION" "$ImageVersion"
     e2e_verify_predicate_v1_buildDefinition_systemParameters "$1" "RUNNER_ARCH" "$RUNNER_ARCH"
-    e2e_verify_predicate_v1_buildDefinition_systemParameters "$1" "RUNNER_NAME" "$RUNNER_NAME"
     e2e_verify_predicate_v1_buildDefinition_systemParameters "$1" "RUNNER_OS" "$RUNNER_OS"
+    # These may differ between the jobs that created the predicate and the job that
+    # verifies.
+    e2e_present_predicate_v1_buildDefinition_systemParameters "$1" "RUNNER_NAME"
+    e2e_present_predicate_v1_buildDefinition_systemParameters "$1" "IMAGE_VERSION"
+
 }
```
ianlewis, line 94: nit: remove extra line.
```suggestion
    e2e_present_predicate_v1_buildDefinition_systemParameters "$1" "RUNNER_NAME"
    e2e_present_predicate_v1_buildDefinition_systemParameters "$1" "IMAGE_VERSION"
}
```

ianlewis (approved):

Issue #1727, opened 2023-03-02
[feature] [docker] Implement registry authentication for docker-based workflows


asraa:
**Is your feature request related to a problem? **

Add registry username/password secrets to authenticate to a registry in order to use the base image.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #1823, opened 2023-03-14
feat: add registry authentication to docker-based workflow

asraa:
Fixes https://github.com/slsa-framework/slsa-github-generator/issues/1727


.github/workflows/builder_docker-based_slsa3.yml
```diff
@@ -37,6 +37,11 @@ defaults:
 ###################################################################
 on:
   workflow_call:
+    secrets:
+      registry-username:
```
laurentsimon, line 41: How about adding a note in the description that it should only be used for high-entropy value, and ask to read the doc (with a link).
We could also add it for the generic container.

laurentsimon (approved):

.github/workflows/builder_docker-based_slsa3.yml
ianlewis: We have a note about this in the docs for the container generator.

See `registry-username` under 'Secrets:'
https://github.com/slsa-framework/slsa-github-generator/blob/6a2cc1cb559a81ffbbcd4248026c6ea89bdab2b6/internal/builders/container/README.md#workflow-inputs

Issue #1814, opened 2023-03-13
[feature] separate setup-token and verify-token


laurentsimon:
See this discussion https://github.com/slsa-framework/slsa-github-generator/issues/1809
Jobs may be scheduled on different runners, so we should not assume they are the same during generation and verification

ianlewis:
We don't validate the runner is the same in `verify-token`. This should only be for tests.

PR #1824, opened 2023-03-14
feat: separate setup-token and verify-token into 2 jobs

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/1814

successful run at https://github.com/laurentsimon/slsa-github-generator/actions/runs/4419476426


.github/workflows/scripts/e2e-verify.common.sh
```diff
@@ -115,7 +115,7 @@ e2e_verify_decoded_token() {
     # GitHub's information.
     _e2e_verify_query "$decoded_token" "$GITHUB_ACTOR_ID" '.github.actor_id'
     _e2e_verify_query "$decoded_token" "$GITHUB_EVENT_NAME" '.github.event_name'
-    _e2e_verify_query "$decoded_token" "$GITHUB_JOB" '.github.job'
+    _e2e_verify_query "$decoded_token" "setup-token" '.github.job'
```
ianlewis, line 118: I wonder if we should even record `$GITHUB_JOB` if it's going to be different than the build job. Maybe make an issue for this so we can follow up?

ianlewis (approved):

.github/workflows/scripts/e2e-verify.common.sh
laurentsimon: I removed the job from the token. Successful run https://github.com/laurentsimon/slsa-github-generator/actions/runs/4429655425

ianlewis: Looks like the unit tests for `verify-token` need to be updated to remove the job id.

laurentsimon: They are removed in the latest commit.

laurentsimon: my bad, let me fix this!

PR #1825, opened 2023-03-14
chore: update docs for verifier version

asraa:
See script https://github.com/slsa-framework/slsa-verifier/pull/530 for SHA verification


laurentsimon (approved):

asraa:
Why is it complaining about `--branch` when that's changed in this PR?

```
verifier hash verification has passed
Usage:
  slsa-verifier [flags]
  slsa-verifier [command]

Available Commands:
  completion         Generate the autocompletion script for the specified shell
  help               Help about any command
  verify-artifact    Verifies SLSA provenance on artifact blobs given as arguments (assuming same provenance)
  verify-image       Verifies SLSA provenance on a container image
  verify-npm-package Verifies SLSA provenance for an npm package tarball [experimental]
  version            Prints the version

Flags:
  -h, --help   help for slsa-verifier

Use "slsa-verifier [command] --help" for more information about a command.

unknown flag: --branch
```

ianlewis:
> Why is it complaining about `--branch` when that's changed in this PR?

It looks like `generate-builder` is checking out the source code from `refs/tags/v1.2.2`.

https://github.com/slsa-framework/slsa-github-generator/blob/13bd59d67f5aeff4317cd943f7cb2bbc8e6d2049/.github/workflows/pre-submit.actions.yml#L305

And we run various actions and scripts from that checkout.

https://github.com/slsa-framework/slsa-github-generator/blob/13bd59d67f5aeff4317cd943f7cb2bbc8e6d2049/.github/actions/generate-builder/action.yml#L78

I'm actually not sure what the right thing to do is here but it feels wrong to me that we are running actions and scripts from the checkout. We should just be using the source code from the checkout to build the builder (I think?).

asraa:
> It looks like generate-builder is checking out the source code from refs/tags/v1.2.2.

Hmm I assumed that this would checkout the source code from the action reference like here https://github.com/slsa-framework/slsa-github-generator/blob/13bd59d67f5aeff4317cd943f7cb2bbc8e6d2049/.github/workflows/builder_go_slsa3.yml#L122

and then pull the builder from v1.2.2 -- this is the one that downloads from the release and uses slsa verifier.

asraa:
Updated to include an optional `builder-ref`. Otherwise, this test was not really acting as a presubmit. It was always checking out v1.2.2's source.

asraa:
Blocked by https://github.com/slsa-framework/slsa-github-generator/pull/1836

.github/actions/generate-builder/action.yml
```diff
@@ -67,11 +70,11 @@ runs:
...
         VERIFIER_REPOSITORY: slsa-framework/slsa-verifier # The repository to download the pre-built verifier binary from.
         VERIFIER_RELEASE_BINARY: slsa-verifier-linux-amd64 # The name of the verifier binary in the release assets.
-        VERIFIER_RELEASE_BINARY_SHA256: b1d6c9bbce6274e253f0be33158cacd7fb894c5ebd643f14a911bfe55574f4c0 # The expected hash of the verifier binary.
-        VERIFIER_RELEASE: v1.3.2 # The version of the verifier to download.
+        VERIFIER_RELEASE_BINARY_SHA256: 538e5346784ddab14189dd66fb995e49c69247985a364558ad3b44e1fbff04cf # The expected hash of the verifier binary.
+        VERIFIER_RELEASE: v2.1.0 # The version of the verifier to download.
 
         COMPILE_BUILDER: "${{ inputs.compile-builder }}"
-        BUILDER_REF: "${{ inputs.ref }}"
+        BUILDER_REF: "${{ inputs.builder-ref || inputs.ref }}"
```
laurentsimon, line 77: Can you add a comment why we need the `||`. By default we take the `builder-ref` but if empty we take the `ref`, why?

.github/actions/generate-builder/action.yml
```diff
@@ -67,11 +70,15 @@ runs:
...
         VERIFIER_RELEASE_BINARY: slsa-verifier-linux-amd64 # The name of the verifier binary in the release assets.
-        VERIFIER_RELEASE_BINARY_SHA256: b1d6c9bbce6274e253f0be33158cacd7fb894c5ebd643f14a911bfe55574f4c0 # The expected hash of the verifier binary.
-        VERIFIER_RELEASE: v1.3.2 # The version of the verifier to download.
+        VERIFIER_RELEASE_BINARY_SHA256: 538e5346784ddab14189dd66fb995e49c69247985a364558ad3b44e1fbff04cf # The expected hash of the verifier binary.
+        VERIFIER_RELEASE: v2.1.0 # The version of the verifier to download.
 
         COMPILE_BUILDER: "${{ inputs.compile-builder }}"
-        BUILDER_REF: "${{ inputs.ref }}"
+        # NOTE: If a builder reference is specified, then we will download this version of the builder.
+        # Otherwise, we sue the same tag reference as the source.
```
laurentsimon, line 78: bit: sue -> use

.github/workflows/pre-submit.actions.yml
```diff
@@ -299,10 +299,14 @@ jobs:
...
+      - name: Detect the builder ref
+        id: detect
+        uses: ./.github/actions/detect-workflow
       - uses: ./.github/actions/generate-builder
         with:
-          repository: "slsa-framework/slsa-github-generator"
-          ref: "refs/tags/v1.2.2"
+          repository: ${{ steps.detect.output.repository }}
+          ref: ${{ steps.detect.outputs.ref }}
+          builder-ref: "refs/tags/v1.2.2"
```
laurentsimon, line 309: should the value of `builder-ref` be retrieved dynamically? Example, the latest version or is does not really matter?
Can you add a comment about it?

asraa: > should the value of builder-ref be retrieved dynamically?

For users outside of this, we should not be retrieving dynamically. That will cause problems like the cosign-installer problems.

It should always be the detected workflow reference.

asraa: I added a small comment about avoiding fetching dynamically and that it should always match the reusable workflow reference for callers outside of presubmit.

laurentsimon: Thanks.

asraa: This was always hard-coded - we don't need to update it. We are testing that any builder version can be fetched from the releases and verified.

There is no canonical builder version to bump here.

asraa: We could invoke via a matrix if you wanted to test all? but one was the only path we were testing.

laurentsimon: sorry, I got confused. I checked the code and it makes sense now. The only reason we're setting`ref: ${{ steps.detect.outputs.ref }}` is because it's a required field, but it's not actually used

Issue #1801, opened 2023-03-10
[feature] prerelease check list


laurentsimon:
Before releasing this feature e2e tests for prerelease support

- [x] add 1 e2e test
- [ ] Update the documentation
- [ ] Add a changelog entry

ianlewis:
Could we just add this is tasks to complete #1795?

asraa:
e2e test created: https://github.com/slsa-framework/example-package/blob/main/.github/workflows/e2e.go.tag.main.config-ldflags-assets-prerelease-tag.slsa3.yml

laurentsimon:
> e2e test created: https://github.com/slsa-framework/example-package/blob/main/.github/workflows/e2e.go.tag.main.config-ldflags-assets-prerelease-tag.slsa3.yml

do we verify the release is marked pre-release in the verification part?

asraa:
> do we verify the release is marked pre-release in the verification part?

can add that - i just looked at manually but that shouldn't be hard

asraa:
> do we verify the release is marked pre-release in the verification part?

done - https://github.com/slsa-framework/example-package/commit/f19de18c243902859071c2367d06724a292e3081

PR #1830, opened 2023-03-16
docs: prerelease doc and CHANGELOG updates for Go builder

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/1795
closes https://github.com/slsa-framework/slsa-github-generator/issues/1801


.github/workflows/builder_go_slsa3.yml
```diff
@@ -372,6 +372,7 @@ jobs:
         if: inputs.upload-tag-name != ''
         with:
           tag_name: "${{ inputs.upload-tag-name }}"
+          prerelease: ${{ inputs.prerelease }}
```
laurentsimon, line 375: NOTE: updated because the Action code updates the prerelease tag in all cases https://github.com/softprops/action-gh-release/blob/master/src/github.ts#L259

ianlewis (approved):

.github/workflows/builder_go_slsa3.yml
asraa: Thanks! I had assumed in doing the review that it would preserve this. Great check.

laurentsimon: just in case...

PR #1836, opened 2023-03-17
fix: detect-workflow-js ref on slsa-github-generator PRs

asraa:
This fixes the behavior of detect-workflow-js for PRs against slsa-github-generator, where the ref should not be the refs/pulls/ but rather the head_sha of the PR.

Adds a test.

This wasn't detected before, and then in PR https://github.com/slsa-framework/slsa-github-generator/pull/1825 I had to add an explicit `git checkout` when compiling the builder from source at the particular builder-ref, which comes from the detect-workflow action. See this run:  https://github.com/slsa-framework/slsa-github-generator/actions/runs/4449301907/jobs/7813310086

Before this, secure-builder-checkout was OK checkout out at a PR ref - I think it was checkout out the merge PR.


laurentsimon (approved):

.github/actions/detect-workflow-js/src/detect.ts
```diff
@@ -60,42 +60,55 @@ export async function detectWorkflowFromContext(
...
+    // GITHUB_JOB_WORKFLOW_SHA becomes fully functional, the OIDC token
+    // detection can be removed and we can identify the current reusable workflow
+    // through the sha of a referenced workflow, fully supporting all triggers
+    // without the repository filter.
+    for (const reusableWorkflow of workflowData.referenced_workflows) {
+      const workflowPath = reusableWorkflow.path.split("@", 1);
+      const [workflowOwner, workflowRepo, ...workflowArray] =
+        workflowPath[0].split("/");
+      const tmpRepository = [workflowOwner, workflowRepo].join("/");
+      if (!reusableWorkflow.ref) {
```
ianlewis, line 87: This seems like it would fail if any reusable workflows in the chain are referenced by digest when I think we only care about the slsa-github-generator ones? Perhaps it should be this?

```suggestion
      if (workflowRepo === "slsa-github-generator" && !reusableWorkflow.ref) {
```

Issue #1864, opened 2023-03-22
[feature] Support folders in secure-upload and secure-download


asraa:
Blocking 
https://github.com/slsa-framework/slsa-github-generator/issues/1655
https://github.com/slsa-framework/slsa-github-generator/issues/1863

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

PR #1866, opened 2023-03-23
feat: secure folder upload Action

laurentsimon:
Part of https://github.com/slsa-framework/slsa-github-generator/issues/1864


.github/actions/secure-upload-folder/action.yml
```diff
@@ -0,0 +1,32 @@
...
+    value: "${{ steps.upload.outputs.sha256 }}"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Create tarball
+      env:
+        PATH: ${{ inputs.path }}
+      run: |
+        tar -cvf "$PATH" "$PATH.tar"
```
ianlewis, line 23: Do you not want to compress it? Also, the arguments are reversed.

```suggestion
        tar -czvf "$PATH.tgz" "$PATH"
```

.github/actions/secure-upload-folder/action.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    - name: Upload the artifact
+      id: upload
+      uses: slsa-framework/slsa-github-generator/.github/actions/secure-upload-artifact@main
+      with:
+        name: "${{ inputs.name }}"
+        path: "${{ inputs.path }}.tar"
+        if-no-files-found: error
+        retention-days: 5
```
ianlewis, line 32: More extensive suggestions.

- Compress the tar file
- `tar` will overwrite the file silently. Write it to a temp directory.
- Don't overwrite `$PATH`
```suggestion
      id: create
      env:
        DIRPATH: ${{ inputs.path }}
      run: |
        # check if a directory
        if [[ ! -d "${DIRPATH}" ]]; then
            >&2 echo "Error: ${DIRPATH} is not a directory."
            exit 1
        fi
        tempdir=$(mktemp -d)
        foldername=$(basename "${DIRPATH}")
        tarball_path="${tempdir}/${foldername}.tgz"
        tar -czvf "${tarball_path}" "${DIRPATH}"
        echo "tarball-path=${tarball_path}" >> "$GITHUB_OUTPUT"

    - name: Upload the artifact
      id: upload
      uses: slsa-framework/slsa-github-generator/.github/actions/secure-upload-artifact@main
      with:
        name: "${{ inputs.name }}"
        path: "${{ steps.create.outputs.tarball-path }}"
        if-no-files-found: error
        retention-days: 5
```

asraa (commented):
I understand this for purposes of uploading folders between trusted VMs. What if these are also outputs? Would the outputs be a tar file, instead of a collection of files when download the artifact?

laurentsimon:
> I understand this for purposes of uploading folders between trusted VMs. What if these are also outputs? Would the outputs be a tar file, instead of a collection of files when download the artifact?

I'll create a secure download folder which will untar and extract the files. Is that OK?

I may be mis-understanding the question

laurentsimon:
> > I may be mis-understanding the question
> 
> If the attestation folder is uploaded using this in the SRW, so that the TRW can securely download it, should the TRW then upload the folder with the normal `upload-artifact` back to the caller? Otherwise the caller will get the attestations uploaded in a tar file.
> 
> I guess I'm curious whether the tar needs to be uploaded, rather than just recompute the tar from the downloaded folder and verify its sum in secure-download.

That's a good point. I was thinking that we'd make a public secure-download-attestations Action that untars. But it's not ideal you're right.

Re-tar / zip may not work unless we make sure timestamps are not recorded and other compression methods don't get in the way. So maybe we need a merkle tree of folder content.. this would still require a public Action for folks to verify but at least the caller can download using the official download-artifact.

Google tells me `find /path/to/folder -type f -print0 | sort -z | xargs -0 sha1sum | sha1sum` may work. I'll need to try it out. TRWs will need to use this so we'd document it.

Wdut?

asraa:
> TRWs will need to use this so we'd document it or create an Action for it. 

I believe we can still package that line into a `secure-download-folder`, so TRWs may not need to know the internals (or even public users where it's less important to do this verification but they may)

Hmmm. 

> Google tells me find /path/to/folder -type f -print0 | sort -z | xargs -0 sha1sum | sha1sum may work. I'll need to try it out.

I think the filenames may matter here (say we download to a temp folder, then we must execute that command INSIDE the temp folder so that the prefix change wouldn't matter). That's not too bad.

I don't know @ianlewis WDYT? in my view it's nice for a user to not have to deal with the tar'd attestations, but I can't tell if introducing this complexity is too much. 

The  folder of build artifacts will also need to be either securely downloaded via the checksums or verified with provenance, right? IMO secure download is easier, since we'll need to match on the many provenance files and artifacts to find the correspondence. In this case as well, we need to either tell TRW users the folder uploaded will be a tar and provide the action, or  do something like the above and hope it stays deterministic

laurentsimon:
> The folder of build artifacts will also need to be either securely downloaded via the checksums or verified with provenance, right? IMO secure download is easier, since we'll need to match on the many provenance files and artifacts to find the correspondence. 

BTW, the official download / upload Action *cannot* download / upload from / to outside the GITHUB_WORKSPACE. At least for download I tried and it cannot. To do that we'd need to call the API ourselves, maybe

> In this case as well, we need to either tell TRW users the folder uploaded will be a tar and provide the action, or do something like the above and hope it stays deterministic

I'd like to avoid relying on determinism, that's likely to break sometimes. Especially since there's no guarantees that the VM use the same runner, iiuc.

ianlewis:
> BTW, the official download / upload Action _cannot_ download / upload from / to outside the GITHUB_WORKSPACE. At least for download I tried and it cannot. To do that we'd need to call the API ourselves, maybe

Yeah, it can. It doesn't check the workspace directory explicitly. Download does fail if the file exists and it won't overwrite it. See this run which successfully uploads `/usr/bin/bash` but fails to overwrite it when it downloads.
https://github.com/ianlewis/actions-test/actions/runs/4506456118/jobs/7933255194

It seems that the download action also lacks permissions to write to most directories outside the workspace. I still think it's valuable for us to limit the path to the workspace dir and not rely on protections inherent to the runner's environment.

ianlewis:
> I believe we can still package that line into a secure-download-folder, so TRWs may not need to know the internals (or even public users where it's less important to do this verification but they may)

I think that if there is a `secure-upload-folder` and `secure-download-folder` action then users shouldn't need to care about how the folder was uploaded as long as the actions work together, will they? Unless they need to use `secure-download-folder` with a tarball they uploaded themselves for some reason?

laurentsimon:
> > BTW, the official download / upload Action _cannot_ download / upload from / to outside the GITHUB_WORKSPACE. At least for download I tried and it cannot. To do that we'd need to call the API ourselves, maybe
> 
> Yeah, it can. It doesn't check the workspace directory explicitly. Download does fail if the file exists and it won't overwrite it. See this run which successfully uploads `/usr/bin/bash` but fails to overwrite it when it downloads. https://github.com/ianlewis/actions-test/actions/runs/4506456118/jobs/7933255194
> 
> It seems that the download action also lacks permissions to write to most directories outside the workspace. I still think it's valuable for us to limit the path to the workspace dir and not rely on protections inherent to the runner's environment.

Agreed. I can add a check to verify the destination does not exist.

laurentsimon:
> > I believe we can still package that line into a secure-download-folder, so TRWs may not need to know the internals (or even public users where it's less important to do this verification but they may)
> 
> I think that if there is a `secure-upload-folder` and `secure-download-folder` action then users shouldn't need to care about how the folder was uploaded as long as the actions work together, will they? Unless they need to use `secure-download-folder` with a tarball they uploaded themselves for some reason?

That was my reasoning too. 

So the summary seems that:
1. Create internal secure upload / download folder by tar-ing the folder first
2. Check in both upload and download that it does not overwrite existing files
3. create a public secure-download-attestation Action that will just call the secure download folder Action

Wdut?

laurentsimon:
> > Create a public secure-download-attestation Action that will just call the secure download folder Action
> 
> Yes! This plan sounds good. I too would like to avoid relying on the determinism hope.

SG. I'll make these changes then. Thanks for all the feedback!

.github/workflows/e2e.upload-folder.schedule.yml
```diff
@@ -0,0 +1,155 @@
+name: schedule secure-upload-folder
+
+on:
+  # Daily run.
```
laurentsimon, line 4: Added this daily run because some of the new Action call other actions `@main`, so it we re to change something in these, we would not catch it during pre-submits.

laurentsimon:
PATL at the changes.

.github/actions/secure-download-folder/action.yml
```diff
@@ -0,0 +1,91 @@
+name: "Secure folder download"
+description: "Download a folder and verify its SHA256"
```
asraa, line 2:
```suggestion
description: "Download a folder uploaded by secure-upload-folder and verify its SHA256"
```

because it relies on the internals of secure-upload-folder (the hardcoded naming of folder.tgz)

asraa (approved):
Nice presubmit tests!

.github/actions/secure-upload-folder/action.yml
```diff
@@ -0,0 +1,50 @@
+name: "Secure folder upload"
+description: "Upload a folder as a tarball and outputs its SHA256"
+inputs:
+  name:
+    description: "Artifact name."
+    default: "artifact"
```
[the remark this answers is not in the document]
laurentsimon: done. Also made artifact name `required`

.github/actions/secure-download-folder/action.yml
laurentsimon: done, thanks

Issue #1885, opened 2023-03-27
[bug][pre-submits] secure-upload-folder.sh: No such file or directory


ianlewis:
Pre-submits for `secure-upload-folder` are failing.

```
/home/runner/work/_temp/e5433f62-ba9c-4718-8a4a-b56fd743f47f.sh: line 2: ./.github/workflows/scripts/schedule.actions/secure-upload-folder.sh: No such file or directory
```
Failing run: https://github.com/slsa-framework/slsa-github-generator/actions/runs/4537510993/jobs/7995441763?pr=1875

laurentsimon:
Let me fix now.

PR #1886, opened 2023-03-27
fix: Update path for secure-upload-folder pre-submit test

ianlewis:
Fixes #1885

PR #1887, opened 2023-03-27
fix: Fix pre-submit for upload-secure-folder

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/1885


ianlewis:
I have #1886 in the works. Will send it in a few minutes.

PR #1888, opened 2023-03-28
feat: harden file access for TS Actions

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/1513


laurentsimon:
@ianlewis @asraa are the pre-submit errors the same you were encountering when you tried the common library?

laurentsimon:
Looks like npm was unable to install when a dependency is a folder and there's no `integrity` field. So I added the tarball of the library `tscommon-0.0.0.tgz` and added it as a dependency instead. I also added a pre-submit checking that the tarball's content is the same as the source code.

.github/actions/tscommon/Makefile
```diff
@@ -0,0 +1,48 @@
...
+
+.PHONY: unit-test
+unit-test: node_modules/.installed ## Runs all unit tests.
+	# NOTE: Make sure the package builds.
+	npm run build
+	npm run test
+
+.PHONY: lint
+lint: node_modules/.installed ## Runs eslint.
+	npm run lint
```
ianlewis, line 48: nit: add `format-check` to make sure the code is formatted.
```suggestion
        npm run format-check
	npm run lint
```

.github/actions/tscommon/README.md
```diff
@@ -0,0 +1,24 @@
...
+npm ERR! 
+npm ERR! Clean install a project
+```
+
+Use:
+```shell
+$ cd tscommon
+$ npm pack
+$ cd ../another-action
+$ npm install --save ../tscommon/tscommon-0.0.0.tgz
```
ianlewis, line 23: Does this mean that we need to run `npm pack` in the `tscommon` directory once to create the `tscommon-0.0.0.tgz` in order to build other actions locally?

.github/actions/tscommon/tscommon-0.0.0.tgz
ianlewis, line 1: Do we need to include the tgz in the repo? I think ncc would include tscommon in each dist for each action so we shouldn't need to keep this in the repo IIUC.

It might make it a bit cumbersome to test locally but it might help to keep binaries out of the repository.

wdut?

.github/workflows/scripts/pre-submit.apis/verify-safefs.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -euo pipefail
+
+# Verify that only safe file functions are used.
+results=$(
+    find .github/actions/ actions/ -maxdepth 3 -name '*.ts' -not -path "*/tscommon/*" -not -path "*/__tests__/*" -type f -print0 |
+    xargs -0 grep '"fs"' ||
```
ianlewis, line 8: nit: only catch imports
```suggestion
    xargs -0 grep 'import.*"fs"' ||
```

ianlewis (approved):

.github/workflows/scripts/pre-submit.apis/verify-safefs.sh
laurentsimon: Updated to `xargs -0 grep -P '(import.*|require.*)"fs"'` to catch `require("fs")` too.

.github/actions/tscommon/README.md
laurentsimon: that's correct. I did not need to do that on my local machine, I could just use `../tscommon`. On GitHub it failed. Adding the tarball as dependency added an `integrity: sh512-` entry to tscommon dependency in the lock file, and this fixed the issue.
But if there's another way to fix it, I'd like to use it.

.github/actions/tscommon/tscommon-0.0.0.tgz
laurentsimon: `npm ci` fails on GitHub without adding the tarball as a dependency. I'm not sure why, maybe it'a a setting...?

.github/actions/tscommon/Makefile
laurentsimon: done.

.github/actions/tscommon/Makefile
```diff
@@ -0,0 +1,48 @@
...
+clean:
+	rm -rf dist lib node_modules
+
+## Testing
+#####################################################################
+
+.PHONY: unit-test
+unit-test: node_modules/.installed ## Runs all unit tests.
+	# NOTE: Make sure the package builds.
+	npm run build
```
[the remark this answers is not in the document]
laurentsimon: there's no `build` defined in the Makefile. I dont see one for other Actions either. I just copied their Makefile. Is it something we should change?

.github/actions/tscommon/Makefile
```diff
@@ -0,0 +1,48 @@
...
+	npm ci
+	touch node_modules/.installed
+
+.PHONY: action
+action: node_modules/.installed ## Builds the action.
+	npm run build
+
+.PHONY: package
+package: node_modules/.installed ## Builds the distribution package.
+	npm run all
```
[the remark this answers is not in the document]
laurentsimon: done, thanks

.github/actions/sign-attestations/package.json
```diff
@@ -10,7 +10,8 @@
     "format-check": "prettier --check '**/*.ts'",
     "package": "ncc build --source-map",
     "lint": "eslint --max-warnings 0 src/**/*.ts",
-    "all": "npm ci && npm run format && npm run lint && npm run build && npm run package"
+    "all": "npm ci && npm run format && npm run lint && npm run build && npm run package",
+    "test": "echo no tests to run"
```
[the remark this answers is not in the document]
laurentsimon: Let me remove it and see if it triggers errors. It was failing so I added it, but it might have been a local run.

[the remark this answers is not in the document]
laurentsimon: seem to work. Closing

.github/actions/tscommon/tscommon-0.0.0.tgz
laurentsimon: I'll wait for an ack before merging the PR, inn case you know of a solution for the problem above which is better than using the tarball

.github/actions/tscommon/Makefile
```diff
@@ -0,0 +1,49 @@
...
+	npm ci
+	touch node_modules/.installed
+
+.PHONY: action
+action: node_modules/.installed ## Builds the action.
+	npm run build
+
+.PHONY: package
+package: node_modules/.installed ## Builds the distribution package.
+	npm run build
```
ianlewis, line 31: nit: I think this makes more sense.
```suggestion
	npm pack
```

You might also consider keeping the tgz in a dist/ directory and packing with `mkdir -p dist/ && npm pack --pack-destination dist/` kind of like how we have done with the other packages/actions. That's a bit more consistent.

.github/actions/tscommon/README.md
ianlewis: SG

.github/actions/tscommon/tscommon-0.0.0.tgz
ianlewis: ok, I got it. Probably it's easier to just include the tgz. See my other comment about using a dist/ dir.

.github/actions/tscommon/Makefile
laurentsimon: we still need to build to "compile" TS to js. npm pack is only for the tarball, iiuc

laurentsimon: made the changes, thatnks.

ianlewis: Right. you can have it depend on build.

PR #1889, opened 2023-03-29
feat: Add a verbose option to the docker-builder

rbehjati:
Tested with `build` and `dry-run` commands.
For `build` tested with an invalid image digest, and the errors are printed to console.

### DIFF
--- a/.github/workflows/builder_docker-based_slsa3.yml
+++ b/.github/workflows/builder_docker-based_slsa3.yml
@@ -250,13 +250,15 @@ jobs:
             --build-definition-path build-definition.json \
             --builder-image "${BUILDER_IMAGE}@${BUILDER_DIGEST}" \
             --git-commit-digest "sha1:${GITHUB_SHA}" \
-            --source-repo "git+https://github.com/${GITHUB_REPOSITORY}${REF}"
+            --source-repo "git+https://github.com/${GITHUB_REPOSITORY}${REF}" \
+            --verbose
           "${GITHUB_WORKSPACE}/${BUILDER_BINARY}" dry-run \
             --build-config-path "${CONFIG_PATH}" \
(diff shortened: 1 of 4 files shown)


internal/builders/docker/commands.go
```diff
@@ -43,6 +49,8 @@ func DryRunCmd(check func(error)) *cobra.Command {
 		Use:   "dry-run [FLAGS]",
 		Short: "Generates and stores a JSON-formatted BuildDefinition based on the input arguments.",
 		Run: func(cmd *cobra.Command, args []string) {
+			defer printVerboseHint(inputOptions.Verbose)
```
rbehjati, line 52: @asraa I added this hoping to get the hint printed out, but it does not work. Is it because `os.Exit(1)` is explicitly called in the `checkExit` function? Or am I doing something wrong?

rbehjati: Yes. This works, but if I don't pass `--verbose`, I don't get "use --verbose to see detailed error". 

Or I can remove this and fix it later.

rbehjati: I removed the hint for now :)

asraa (commented):
Thanks! One more - could you add the --verbose option to the workflow build and dry-run commands?

rbehjati:
> Thanks! One more - could you add the --verbose option to the workflow build and dry-run commands?

Good point. Thanks. Done. Can you take another look?

asraa (approved):
That's perfect - I'm happy with this being an option. I think I would've been OK if it was always turned on since GH logs are long and verbose anyway, but I like this feature, just in case there's some reason that the logs are leaky or need to reason to be silenced

ianlewis (approved):

PR #1892, opened 2023-03-30
fix: verify-token read path error

laurentsimon:
See https://github.com/slsa-framework/slsa-github-generator/issues/1890


ianlewis (approved):
LGTM. But I think our secure `fs` routines should allow reading/writing to `/tmp` and maybe `/var/tmp`. We can probably ignore `$TMPDIR` since I don't think it's really used.
See: https://superuser.com/questions/332610/where-is-the-temporary-directory-in-linux

Also we should also allow reading/writing to the directory in [`$RUNNER_TEMP`](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)

asraa (approved):

laurentsimon:
> LGTM. But I think our secure `fs` routines should allow reading/writing to `/tmp` and maybe `/var/tmp`. We can probably ignore `$TMPDIR` since I don't think it's really used. See: https://superuser.com/questions/332610/where-is-the-temporary-directory-in-linux
> 
> Also we should also allow reading/writing to the directory in [`$RUNNER_TEMP`](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)

I've added read. write access to `/tmp` and `$RUNNER_TEMP`, and read-only access to GITHUB_EVENT_PATH.

laurentsimon:
Temporarily disabling the secure-upload-folder pre-submit to merge. The Action calls compute-sha256@main, so the changes in this PR don't take effect. To fix this I think we need to use the detect-env Action and checkout the repo at the right ref, similar to what we do for builder checkout in Go codebase.

ianlewis:
> Temporarily disabling the secure-upload-folder pre-submit to merge. The Action calls compute-sha256@main, so the changes in this PR don't take effect. To fix this I think we need to use the detect-env Action and checkout the repo at the right ref, similar to what we do for builder checkout in Go codebase.

Yeah, this may make things a bit harder since any change to the `tscommon` library will affect all of it's downstream dependencies.

PR #1895, opened 2023-03-31
feat: Add npm-publish input

### LINKED ISSUE #1894: [feature][nodejs] Support publishing on non-tag events
Currently publishing a package with `npm publish` is skipped if the `github.ref` doesn't start with `refs/tags/`. We should support publishing on other events, not just when a tag is pushed.

ianlewis:
Fixes #1894

Adds an `npm-publish` input to allow users to specify whether to publish the package to the npm repository.


.github/workflows/builder_nodejs_slsa3.yml
```diff
@@ -83,7 +89,7 @@ jobs:
 
   slsa-publish:
     needs: [slsa-run]
-    if: startsWith(github.ref, 'refs/tags/')
+    if: ${{ inputs.npm-publish }}
```
laurentsimon, line 92: so we don't want to automatically publish on new tags?
Would it complicate because you want to check if the tag is the same as the version declared in the package?
I think it's probably fine to do simply document that users should set to `npm-publish: startsWith(github.ref, 'refs/tags')` instead

laurentsimon (approved):

.github/workflows/builder_nodejs_slsa3.yml
ianlewis: I set the input's default to be `startsWith(github.ref, 'refs/tags')` so that it should publish by default on new tags.

ianlewis: Right now I'm not checking that the tag == the package version. I think that's going to have to be a check that the verifier does anyway.

PR #1906, opened 2023-04-02
feat: add additional information about viewing attestation

### LINKED ISSUE #1879: [feature] use search.sigstore.dev as an additional output of the provenance generation
**Is your feature request related to a problem? Please describe.**
This is a feature and not related to a problem.

Based on the recent news[^1] that Chainguard contributes the Rekor Search UI project to Sigstore, we might reconsider using this address in one of the outputs about uploading provenance to Rekor.

> This site is now served under: [search.sigstore.dev](https://search.sigstore.dev/)

As of now today, `slsa-github-generator` [outputs](https://github.com/slsa-framework/slsa-github-generator/blob/64fee9e1f91c3423606e50050d7691b738de4403/internal/builders/go/pkg/provenance.go#L190) the following message once it generated provenance:

```bash
Uploaded signed attestation to rekor with UUID 24296fb24b8ad77ad7d75dd8666c9c2d015ef7b498511222851da95142c808257ca7986691f9f38f.
```

From an end-user perspective who might not know how to search this UUID in the transparency log, I thought that we can give information about what would be the next step of using this UUID, for example, we can add an informative message like this:

```bash
You can use rekor-cli to see the log entry details:

$ rekor-cli get --uid 24296fb24b8ad77ad7d75dd8666c9c2d015ef7b498511222851da95142c808257ca7986691f9f38f 

OR

You can use search.sigstore.dev to see the log entry details:

https://search.sigstore.dev/?uuid=24296fb24b8ad77ad7d75dd8666c9c2d015ef7b498511222851da95142c808257ca7986691f9f38f
```

Similar to this issue in Tekton Chains: _https://github.com/tektoncd/chains/issues/576_

[^1]: _https://www.chainguard.dev/unchained/chainguard-contributes-rekor-search-project-to-sigstore_

developer-guy:
Fixes #1879


signing/sigstore/rekor.go
```diff
@@ -88,6 +88,10 @@ func (r *Rekor) Upload(ctx context.Context, att signing.Attestation) (signing.Lo
 	}
 
 	fmt.Printf("Uploaded signed attestation to rekor with UUID %s.\n", uuid)
+	fmt.Printf(`You could use rekor-cli to view the log entry details:"
+	"$ rekor-cli get --uuid %s --rekor_server %s",
+	"In addition to that, you could also use the Rekor Search UI:"
+	"https://search.sigstore.dev/?uuid%s`, uuid, r.rekorAddr, uuid)
```
ianlewis, line 94: I think the '"' double quotes will get printed in a raw string like this.
See: https://go.dev/play/p/6hG4rHMvJ6w

I think you need to remove the double quotes and the whitespace at the beginning of the line.

Also, for now I think it doesn't matter, but the second half is only relevant if using the public rekor instance. Maybe something like the following? (https://go.dev/play/p/7ZbUsexpgIa)

```suggestion
	fmt.Printf("You could use rekor-cli to view the log entry details:\n\n"+
		"  $ rekor-cli get --uuid \"%s\" --rekor_server \"%s\"\n\n", uuid, r.rekorAddr)
	if r.rekorAddr == DefaultRekorAddr {
		fmt.Printf("In addition to that, you could also use the Rekor Search UI: \n\n"+
			"  https://search.sigstore.dev/?uuid=%s\n\n", uuid)
	}
```

developer-guy: or for the sake of simplicity, we may ignore adding the Rekor server address to the log message.

developer-guy: I've updated accordingly: https://go.dev/play/p/EnhhKXMjRKv

ianlewis (approved):

PR #1913, opened 2023-04-03
fix: Dont validate or read event path from slsa token

### LINKED ISSUE #1900: [bug] Don't validate or read event_path from slsa-token
We're validating and reading the event_path from the token (https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/verify-token/src/predicate1.ts#L121). I'm not sure there is guarantee that the re-usable workflow's runner's path is identical to the caller's.

I think we need to remove the validation that slsa-token.event_path == GITHUB_EVENT_PATH and always read from GITHUB_EVENT_PATH.

ianlewis:
Fixes #1900 

Adds a new `event_payload_sha256` field to the BYOB SLSA token. This contains a sha256 checksum in hex format of the event payload as read from `GITHUB_EVENT_PATH`. The checksum is compared against the checksum of the same file when read by `verify-token`. This ensures that the same event payload for the user's workflow is being included in provenance.

A new function `safeFileSha256` was added to `tscommon` since it's used in a number of places. `tscommon` was updated to `0.1.0` and a new tarball was added. Adding a new tarball will help with understanding which actions depend on which version of `tscommon` without necessarily having to update every action whenever `tscommon` changes. We can remove old version of `tscommon` when we are sure they are not used.


.github/actions/verify-token/src/validate.ts
```diff
@@ -24,11 +25,11 @@ export function validateGitHubFields(gho: githubObj): void {
...
 
-  // event_path
+  // event_payload_sha256
   validateField(
-    "github.event_path",
-    gho.event_path,
-    process.env.GITHUB_EVENT_PATH
+    "github.event_payload_sha256",
+    gho.event_payload_sha256,
+    tscommon.safeFileSha256(process.env.GITHUB_EVENT_PAYLOAD || "")
```
laurentsimon, line 32: Hope we don't run into unexpected problems with this hardened code.  There is not a lot of interesting info in it https://github.com/laurentsimon/slsa-on-github-test/actions/runs/4601805495/jobs/8130089172 ifaict

.github/actions/tscommon/src/file.ts
```diff
@@ -6,11 +7,21 @@ import process from "process";
...
     return wdt;
   }
-  return process.env["GITHUB_WORKSPACE"] || "";
+  return process.env.GITHUB_WORKSPACE || "";
+}
+
+// safeFileSha256 returns the hex-formatted sha256 sum of the contents of an
+// untrusted file path.
+export function safeFileSha256(untrustedPath: string): string {
+  if (!safeExistsSync(untrustedPath)) {
```
laurentsimon, line 20: nit: you don't need this line. `safeReadFileSync()` will check that the file is safe to read. If it does not exists, the built-in read call https://github.com/slsa-framework/slsa-github-generator/blob/d10c919298e18c8afb2b5f68e1872ec98a750cfb/.github/actions/tscommon/src/file.ts#L82 will throw an exception.

.github/actions/verify-token/package.json
```diff
@@ -27,7 +27,7 @@
     "@actions/github": "^5.1.1",
     "@octokit/webhooks-types": "^6.9.0",
     "sigstore": "^1.0.0",
-    "tscommon": "file:../tscommon/tscommon-0.0.0.tgz"
+    "tscommon": "file:../tscommon/tscommon-0.1.0.tgz"
```
laurentsimon, line 30: you'll need to update https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/pre-submit.actions.yml#L26 to iterate thru the tarballs

laurentsimon (approved):

.github/actions/verify-token/package.json
ianlewis: Done.

.github/actions/verify-token/src/validate.ts
ianlewis: I tested this and the event payload doesn't change. It only includes info about what triggered the initial workflow
https://github.com/ianlewis/actions-test/actions/runs/4604876020

.github/workflows/pre-submit.actions.yml
```diff
@@ -17,13 +17,20 @@ jobs:
...
+      matrix:
+        version: ["0.0.0", "0.1.0"]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@8f4b7f84864484a7bf31766abe9204da3cbe65b3 # v3.5.0
 
       - name: Untar the package tarball
         working-directory: .github/actions/tscommon
-        run: tar xvzf tscommon-0.0.0.tgz
+        env:
```
laurentsimon, line 29: Thought about this a bit more. Since we're checking out a single repo, extracting each tarball will fail as soon as we update the code, because each tarball is generated with a different codebase. Do I understand correctly?

Issue #1901, opened 2023-03-31
[e2e]: docker-based tag main default slsa3 


ianlewis:
Repo: https://github.com/slsa-framework/example-package/tree/v35.0.2
Run: https://github.com/slsa-framework/example-package/actions/runs/4577186955
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.docker-based.tag.main.default.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.docker-based.tag.main.default.slsa3.yml
Trigger: push
Branch: v35.0.2
Date: Fri Mar 31 16:36:04 UTC 2023

ianlewis:
Repo: https://github.com/slsa-framework/example-package/tree/v35.0.3
Run: https://github.com/slsa-framework/example-package/actions/runs/4581243051
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.docker-based.tag.main.default.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.docker-based.tag.main.default.slsa3.yml
Trigger: push
Branch: v35.0.3
Date: Sat Apr  1 04:13:44 UTC 2023

ianlewis:
Repo: https://github.com/slsa-framework/example-package/tree/v35.0.4
Run: https://github.com/slsa-framework/example-package/actions/runs/4586832608
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.docker-based.tag.main.default.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.docker-based.tag.main.default.slsa3.yml
Trigger: push
Branch: v35.0.4
Date: Sun Apr  2 04:13:19 UTC 2023

ianlewis:
@rbehjati @asraa Could one of you two take a look at this?

ianlewis:
Repo: https://github.com/slsa-framework/example-package/tree/v35.0.5
Run: https://github.com/slsa-framework/example-package/actions/runs/4592814930
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.docker-based.tag.main.default.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.docker-based.tag.main.default.slsa3.yml
Trigger: push
[...]

PR #1914, opened 2023-04-03
fix: add GITHUB_BASE_REF to SLSA v1.0 creation

asraa:
Fixes  https://github.com/slsa-framework/example-package/actions/runs/4597557193/jobs/8120481712

Even though `GITHUB_BASE_REF` is not populated except for pull_request, slsa-verifier expects to check this field of the provenance. See error below:

```
Verifying artifact config.toml: FAILED: invalid DSSE envelope payload: environment type for GITHUB_BASE_REF

FAILED: SLSA verification failed: invalid DSSE envelope payload: environment type for GITHUB_BASE_REF
```

Will fix https://github.com/slsa-framework/slsa-github-generator/issues/1901


laurentsimon (commented):
Thanks

laurentsimon (approved):

Issue #1737, opened 2023-03-03
[feature][byob] Distinguish between user inputs and TRW inputs


laurentsimon:
See https://github.com/actions/runner/issues/2274

We could:
1. Ask TRW writers to give a list of inputs. This is error prone, especially when TRW writers start updating their inputs.
2. Compare the TRW inputs (in the token) with those in the workflow_dispatch event. This does not work if some TRW inputs have the same name as the workflow_dispatch
3. Read the TRW workflow and parse it

ianlewis:
> 2\. This does not work if some TRW inputs have the same name as the workflow_dispatch

What happens in this case? Is one overwritten by the other?

laurentsimon:
Looks like there is `on.workflow_call.inputs` we can use. I have not confirmed, but this would mean the only necessary change is for TRWs to use `toJson(on.workflow_call.inputs)` instead of `toJson(inputs)`

ianlewis:
> Looks like there is `on.workflow_call.inputs` we can use. I have not confirmed, but this would mean the only necessary change is for TRWs to use `toJson(on.workflow_call.inputs)` instead of `toJson(inputs)`

The reply on [the other bug ](https://github.com/actions/runner/issues/2274) is basically your option #3. This doesn't exist as a context that passed at runtime.

laurentsimon:
I think we need to the Fulcio claims to be available to do this. The certificate contains the `GITHUB_SHA` of the project, but not of the TRW, ie the workflow_sha is not present in Fulcio certs yet. We need https://github.com/sigstore/fulcio/pull/945 to land.
@haydentherapper when do you think it will be implemented?

haydentherapper:
Working on it currently! Hoping in a couple weeks, it's pretty straightforward.

laurentsimon:
Awesome, please ping us here when it's landed if you can

laurentsimon:
The PR got merged, and @haydentherapper will put it in staging, so we may be able to test it. @haydentherapper What's the fulcio URL / TUF to test against?

haydentherapper:
https://docs.sigstore.dev/cosign/keyless/#public-staging-environment - fulcio.sigstage.dev, and `tuf-root-staging` for the staging bucket.

PR #1917, opened 2023-04-04
feat: Distinguish between reusable workflow inputs and trigger workflow inputs

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/1737


.github/actions/verify-token/src/index.ts
```diff
@@ -127,29 +133,40 @@ async function run(): Promise<void> {
...
+    // they may be empty.
+    // TODO(#1780): test empty inputs.
+
+    // Extract certificate information.
+    const [toolURI, toolRepository, toolRef, toolSha, toolPath] =
+      parseCertificate(bundle);
+
+    // Extract the inputs.
+    // See https://github.com/slsa-framework/slsa-github-generator/issues/1737.
+    const rawFinalTokenObj = await filterWorkflowInputs(
```
ianlewis, line 146: nit: it's not actually final

```suggestion
    const rawFilteredTokenObj = await filterWorkflowInputs(
```

.github/actions/verify-token/src/utils.ts
```diff
@@ -22,3 +27,172 @@ export function getWorkflowPath(obj: githubObj): string {
...
+  inputs: Map<string, string | number | boolean>
+): Map<string, string | number | boolean> {
+  if (!inputs) {
+    return new Map<string, string | number | boolean>();
+  }
+  if (!(inputs instanceof Map)) {
+    return new Map<string, string | number | boolean>(Object.entries(inputs));
+  }
+  return inputs;
+}
```
ianlewis, line 41: This kind of confuses me. If `inputs` is defined as a `Map` wouldn't it already be a `Map`? Why do we need to do a conversion?

.github/actions/verify-token/Makefile
```diff
@@ -44,3 +44,4 @@ unit-test: node_modules/.installed ## Runs all unit tests.
 lint: node_modules/.installed ## Runs eslint.
 	npm run format-check
 	npm run lint
+"yaml": "^2.2.1"
```
ianlewis, line 47: This is a c&p miss right?

.github/actions/verify-token/src/inputs.ts
```diff
@@ -0,0 +1,76 @@
...
+  return updateSLSAToken(content, slsaToken);
+}
+
+export function updateSLSAToken(
+  content: string,
+  slsaToken: rawTokenInterface
+): rawTokenInterface {
+  const ret = Object.create(slsaToken);
+  const workflow: GitHubWorkflowInterface = YAML.parse(content);
+  slsaToken.tool.inputs = asMap(slsaToken.tool.inputs);
```
ianlewis, line 41: What is the type of `slsaToken.tool.inputs` initially?

.github/actions/verify-token/src/inputs.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  // No inputs field defined.
+  if (!workflow.on.workflow_call.inputs) {
+    core.info("no input defined in the workflow");
+    ret.tool.inputs = new Map();
+    return ret;
+  }
+
+  // Inputs defined.
+  const wInputsMap = new Map(Object.entries(workflow.on.workflow_call.inputs));
```
ianlewis, line 57: Would this not use `asMap` as well?

.github/actions/verify-token/src/inputs.ts
```diff
@@ -0,0 +1,76 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import * as core from "@actions/core";
+import * as YAML from "yaml";
+import { rawTokenInterface, GitHubWorkflowInterface } from "../src/types";
+import { fetchToolWorkflow, asMap } from "./utils";
+
+export async function filterWorkflowInputs(
```
ianlewis, line 19: nit: add a doc comment that explains what this function is for.

.github/actions/verify-token/src/types.ts
```diff
@@ -65,3 +65,11 @@ export interface rawTokenInterface {
     masked_inputs: string[];
   };
 }
+
+export interface GitHubWorkflowInterface {
```
ianlewis, line 69: nit: add doc comment
```suggestion
// GitHubWorkflowInterface represents a GitHub Actions reusable workflow.
// See: https://docs.github.com/en/actions/using-workflows/reusing-workflows#creating-a-reusable-workflow
export interface GitHubWorkflowInterface {
```

.github/actions/verify-token/src/utils.ts
```diff
@@ -1,4 +1,9 @@
 import { githubObj } from "./types";
+import * as core from "@actions/core";
+import * as sigstore from "sigstore";
```
ianlewis, line 3: nit: avoid the `sigstore.sigstore` stuff.

```suggestion
import { sigstore } from "sigstore";
```

.github/actions/verify-token/src/utils.ts
```diff
@@ -22,3 +27,172 @@ export function getWorkflowPath(obj: githubObj): string {
...
+    throw new Error(
+      `bundle.verificationMaterial.x509CertificateChaincertificates is empty.`
+    );
+  }
+  // NOTE: the first certificate is the client certificate.
+  const clientCertDer = Buffer.from(
+    bundle.verificationMaterial.x509CertificateChain.certificates[0].rawBytes,
+    "base64"
+  );
+  const clientCertPath = "client.cert";
```
ianlewis, line 69: nit: Maybe consider writing to `$RUNNER_TEMP` rather than the workspace?

.github/actions/verify-token/src/utils.ts
```diff
@@ -22,3 +27,172 @@ export function getWorkflowPath(obj: githubObj): string {
...
+  return [repo, ref];
+}
+
+export async function fetchToolWorkflow(
+  ghToken: string,
+  repoName: string,
+  hash: string,
+  workflowPath: string
+): Promise<string> {
+  const url = `https://raw.githubusercontent.com/${repoName}/${hash}/${workflowPath}`;
```
ianlewis, line 182: Can you not do this with the Octokit client that's part of the GitHub Actions toolkit?

I think it would be something like:

```typescript
import * as github from "@actions/github";

const octokit = github.getOctokit(ghToken);

let result = await octokit.repos.getContent({
  owner: ownerName,
  repo: repoName,
  path: workflowPath
})

// content will be base64 encoded
const body = Buffer.from(result.data.content, 'base64').toString()
```

.github/actions/verify-token/package.json
```diff
@@ -25,12 +25,15 @@
     "@actions/core": "^1.10.0",
     "@actions/github": "^5.1.1",
     "@octokit/webhooks-types": "^6.9.0",
+    "node-fetch": "^2.6.9",
```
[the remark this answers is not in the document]
ianlewis: Or even better, use octokit. See comment below.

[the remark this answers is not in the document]
laurentsimon: would love to use another dep, ill take a look at http

.github/actions/verify-token/src/utils.ts
laurentsimon: It confused me too. I thought so too, but when testing it does not work. Here's what I saw in practice:

Both `JSON.parse()` and `YAML.parse()` seem to return an object that is *not* a Map:
`const rawTokenObj: rawTokenInterface = JSON.parse(rawTokenStr);`
`const workflow: GitHubWorkflowInterface = YAML.parse(content);`

even though the definition of the `rawTokenInterface` and `GitHubWorkflowInterface` are.

I also found out that TS lets me assign `object to the Map field (hose coming from the `parse()`); I am not sure why. So I need to check whether it's a map or not. And the function above, as you say, takes as input a `Map` yet I can pass it the result of `[YAML|JSON].parse()` above which is actually not a map, and it still compiles and runs properly.

I'm not sure why this is happening or even allowed in TS... But if there's a better solution, I'd like to use it

.github/actions/verify-token/src/inputs.ts
laurentsimon: it's an `object`, even though we declare it as a `Map` in the structure

.github/actions/verify-token/src/utils.ts
laurentsimon: +1, the hash seems to be supported so I'll update. Thanks

.github/actions/verify-token/src/utils.ts
ianlewis: Lets' just document this in comments for now.

.github/actions/verify-token/src/inputs.ts
ianlewis: hmm. I guess Map and Object are basically compatible types in TS. I'll need to look up the details.

.github/actions/verify-token/src/utils.ts
laurentsimon: done, ptal

.github/actions/verify-token/package.json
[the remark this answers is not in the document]
laurentsimon: Used oktokit instead.. which uses node-fetch under the hood.

.github/actions/verify-token/src/utils.ts
laurentsimon: Added this comment https://github.com/slsa-framework/slsa-github-generator/issues/1918. Since this changes tscommon, I think we should resolve that issue first.

.github/actions/verify-token/src/inputs.ts
ianlewis: nit: you missed this comment I think.

ianlewis (approved):

.github/actions/verify-token/src/inputs.ts
laurentsimon: I did miss it. I updated asMap to be a generic, because the workflow inputs are a map string -> object, whereas the reusable workflow inputs are a map string -> bool | string | int.

laurentsimon:
@ianlewis any idea what's going on with pre-submit failures? We updated cosign or another dep recently I suppose. Not urgent to merge.

ianlewis:
> @ianlewis any idea what's going on with pre-submit failures? We updated cosign or another dep recently I suppose. Not urgent to merge.

Not sure. The only thing touching `go.mod` recently was the update to Go 1.20. Are the changes you made to `go.mod` on this PR necessary?

laurentsimon:
> > @ianlewis any idea what's going on with pre-submit failures? We updated cosign or another dep recently I suppose. Not urgent to merge.
> 
> Not sure. The only thing touching `go.mod` recently was the update to Go 1.20. Are the changes you made to `go.mod` on this PR necessary?

absolutely not. I must have tried something while in the wrong folder. Reverted.

PR #1919, opened 2023-04-05
feat: Add support for uploading to release to Node.js builder

### LINKED ISSUE #888: [feature] npm builder: upload tarball to GitHub releases
Let's try to upload the package tarball to GitHub release assets, in addition to publishing it.

ianlewis:
Fixes #888


.github/workflows/builder_nodejs_slsa3.yml
```diff
@@ -52,6 +52,24 @@ on:
...
+        required: false
+        type: boolean
+        default: true
+
+      upload-tag-name:
+        description: >
+          If non-empty and `upload-assets` is set to true, the provenance is uploaded to the GitHub
+          release identified by the tag name. If a workflow is run on a new tag and `upload-tag-name`
+          is non-empty, the new tag is ignored and the value of `upload-tag-name` is used instead to upload
+          the assets.
```
asraa, line 69: Why would this be the case? Would it be better to error out if upload-tag-name is non-empty and it's running on a new tag? I'd assume if upload-tag-name if populated, they should be running on workflow_dispatch - or else you'd get tag mismatch issues when doing slsa verification

I believe this is not the case in the impl below - if its a tag ref, with upload-tag-name, it'll upload to both

asraa (approved):
lgtm minus docstring

.github/workflows/builder_nodejs_slsa3.yml
ianlewis: > Why would this be the case?

The value of the tag used to identify the release comes from the tag name by default (if pushing a new tag). But if not, you need to specify the `upload-tag-name` input to provide a tag name to upload to. If you don't it doesn't do anything.

> Would it be better to error out if upload-tag-name is non-empty and it's running on a new tag?

I'm not sure. A user may want to create multiple tags for a single release (v1, v1.1, v1.1.4) but only create a GitHub release and upload assets to one of them (which may or may not be the one originally pushed to).

> I believe this is not the case in the impl below - if its a tag ref, with upload-tag-name, it'll upload to both

Good catch! fixed.

asraa: Hmmm interesting. I'm only I guess curious of if it runs on a tag vX and uploads to tag vY, then slsa-verifier will fail when verifying against vY.. 

I suppose this is kind of a general problem - we don't have a way to rebuild provenance at a certain tag

ianlewis: Right. IIUC, slsa-verifier will fail when verifying against any version given to `upload-tag-name` if it doesn't match the   `github.ref` no matter what. It's kind of a property of `upload-tag-name` on all the workflows FWIW

ianlewis: Created issue #1947 so we can track this.

PR #1983, opened 2023-04-12
chore: Use slsa-verifier v2.2.0 in generate-builder

### LINKED ISSUE #1996: [feature] Build slsa-github-generator binaries from release tag
Update the release.yml workflow to use the previously released verison of slsa-github-generator so that released binaries are built with an actual release at a tag.

This will make it easier to verify our binaries with `slsa-verifier` v2.2.0+ and require less special cases for our binaries.

ianlewis:
Fixes #1996


ianlewis:
Blocked by #1984

asraa (approved):

laurentsimon (approved):
The sha matches changes in this PR https://github.com/slsa-framework/slsa-verifier/pull/556

ianlewis:
Was blocked by Docker updating the `golang` base image **again**. Resolved by #1988 

This PR should be able to move forward.

ianlewis:
Blocked by #1996

ianlewis:
> Blocked by #1996

We need to wait until v1.6.0 is released and update the [builder ref](https://github.com/slsa-framework/slsa-github-generator/blob/fac372e1d8693e92834fae3140e362ffc0405d38/.github/workflows/pre-submit.actions.yml#L484) to `refs/tags/v1.6.0`

ianlewis:
> We need to wait until v1.6.0 is released and update the [builder ref](https://github.com/slsa-framework/slsa-github-generator/blob/fac372e1d8693e92834fae3140e362ffc0405d38/.github/workflows/pre-submit.actions.yml#L484) to `refs/tags/v1.6.0`

Actually we can't do that because we need this change to be merged for adversarial tests in RELEASE.md to work. So we need to temporarily disable the branch protection check for `generate-builder-generic-no-compile` and merge this change, do the v1.6.0 release which includes no-compile checks, and then after the release update the  [builder ref](https://github.com/slsa-framework/slsa-github-generator/blob/fac372e1d8693e92834fae3140e362ffc0405d38/.github/workflows/pre-submit.actions.yml#L484)  to v1.6.0 and reinstate the branch protection check.

ianlewis:
I actually managed to get around the issue by adding a new `testing` flag to `generate-builder` which sets the `SLSA_VERIFIER_TESTING` flag. This should allow us to keep the pre-submit test.

Once we release v1.6.0 we can use that release for the pre-submit test and remove the `testing` input on `generate-builder`.

PR #1988, opened 2023-04-13
chore: Remove old detect-workflow action

### LINKED ISSUE #1401: [bug] verify base image pre-submit fails
See https://github.com/slsa-framework/slsa-github-generator/actions/runs/3751444140/jobs/6372431280

I have temporarily disabled this check in the branch protection setting

ianlewis:
Fixes #1401 

Removes the detect-workflow Docker container action in favor of the detect-workflow-js action. Also removes the base image verification pre-submit as it's not needed anymore.

/cc @asraa


.github/workflows/generator_generic_slsa3.yml
```diff
@@ -115,7 +115,7 @@ jobs:
       - name: Detect the generator ref
         id: detect
         continue-on-error: true
-        uses: slsa-framework/slsa-github-generator/.github/actions/detect-workflow@main
+        uses: slsa-framework/slsa-github-generator/.github/actions/detect-workflow-js@main
```
laurentsimon, line 118: we could as well rename the js action as `detect-workflow`. Maybe a follow-up PR.

laurentsimon (approved):

.github/workflows/generator_generic_slsa3.yml
ianlewis: Yes. I was going to do it on this PR but it just made it unreadable so I'll do it in a follow up.

Issue #1918, opened 2023-04-04
[discussion] Reconsider tscommon versioning


laurentsimon:
See comment https://github.com/slsa-framework/slsa-github-generator/pull/1913/files#r1157725414
/cc @ianlewis

ianlewis:
> See comment https://github.com/slsa-framework/slsa-github-generator/pull/1913/files#r1157725414 /cc @ianlewis

SYAC

> Thought about this a bit more. Since we're checking out a single repo, extracting each tarball will fail as soon as we update the code, because each tarball is generated with a different codebase. Do I understand correctly? Maybe using multiple tarballs is adding more complexity that it solves. Let me know, I may be mis-understanding something.

Right. The issue I found is that if you don't run `npm ci` then you don't catch that `tscommon` was updated. But you're comment is right, I think pre-submits would likely catch when the code is updated as integrity checks would fail when running `npm ci`.

laurentsimon:
+1, that's also my understanding.

Let's address https://github.com/slsa-framework/slsa-github-generator/pull/1917#discussion_r1157918002 once this issue resolved.

PR #1990, opened 2023-04-13
feat: use a single version of tscommon

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/1918


.github/actions/verify-token/package.json
```diff
@@ -26,8 +26,7 @@
     "@actions/github": "^5.1.1",
     "@octokit/webhooks-types": "^6.9.0",
     "sigstore": "^1.1.1",
-    "tscommon": "file:../tscommon/tscommon-0.1.0.tgz",
-    "yaml": "^2.2.1"
```
ianlewis, line 30: Did we not need the `yaml` package anymore?

ianlewis (approved):

.github/actions/verify-token/package.json
laurentsimon: I think it was never needed. Pre-submit are passing. tscommon only uses files and crypto deps.

laurentsimon: Sorry, mis-read. I thought the dep was imported in the tscommon. Why did the pre-submit not catch it? Weird, ill send a PR

ianlewis: Yeah, I think it works since all the tests run `npm ci` which just looks at `package-lock.json`. Since it's included there everything's fine.

keeping `package.json` in sync with `package-lock.json` is mostly for us developers to make sense of the dependencies.

laurentsimon: ah, makes sense.

PR #2009, opened 2023-04-19
fix: detect-workflow-js for PRs from admins on same repo

ianlewis:
OIDC tokens may be available when a PR's source and destination are the same repo and the PR author has write access to the repo.

This updates `detect-workflow-js` to more generally handle the case of pull requests and merge queues.


.github/actions/detect-workflow-js/package.json
```diff
@@ -8,7 +8,7 @@
     "build": "tsc",
     "format": "prettier --write '**/*.ts'",
     "format-check": "prettier --check '**/*.ts'",
-    "package": "ncc build --source-map",
+    "package": "ncc build --external encoding --source-map",
```
ianlewis, line 11: Note to reviewers:

'encoding' is an optional package included as an optional "peerDependency" of `node-fetch` which is required by `octokit`. I'm not sure what makes `ncc` determine whether to include it in the dist/ or not but this option makes it so it's properly excluded.

.github/actions/detect-workflow-js/src/detect.ts
```diff
@@ -62,11 +62,11 @@ export async function detectWorkflowFromContext(
...
 
-  // If this is a slsa-github-generator repository or fork, then look
-  // for the repo and head SHA from the pull_request event value.
+  // If this is a pull request, then look for the repo and head SHA from the
+  // pull_request event value.
   if (
-    workflowData.event === "pull_request" &&
-    workflowData.repository.name === "slsa-github-generator"
+    workflowData.event === "pull_request" ||
+    workflowData.event === "merge_group"
```
laurentsimon, line 69: do we need to support PR for other repositories? SLSA 1.0 specs only requires supporting a subset of events, none of which are PR triggers.

ianlewis: The repository name isn't `slsa-framework/slsa-github-generator` for GHA runs on PRs. They run in the fork and the fork could be named anything like `my-slsa-generator-fork`.

The reason I removed this is that this forces contributors to name their fork `slsa-github-generator`. If they change the name the pre-submits would fail and the check doesn't really improve security.

ianlewis: Right. Maybe checking the `full_name` would be better.
https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository

ianlewis: Done.

.github/actions/detect-workflow-js/src/detect.ts
```diff
@@ -62,11 +62,12 @@ export async function detectWorkflowFromContext(
 
   let [repository, ref, workflow] = ["", "", ""];
 
-  // If this is a slsa-github-generator repository or fork, then look
-  // for the repo and head SHA from the pull_request event value.
+  // If this is a pull request, then look for the repo and head SHA from the
```
laurentsimon, line 65: nit: add back the comment about slsa-github-generator, since the condition checks for it.

laurentsimon (approved):

.github/actions/detect-workflow-js/src/detect.ts
ianlewis: Done.

PR #2010, opened 2023-04-20
fix: Don't record RUNNER_XXX and IMAGE_XXX variables

### LINKED ISSUE #1817: [bug] Don't record RUNNER_XXX
We record RUNNER_NAME, RUNNER_OS and RUNNER_IMAGE from the setup-token. These values are taken from the caller, and don't represent the values where the build actually takes place. 

I suggest we remove them. The runner label (if available) can be reported since our re-usable workflows decide which runner to run. (If these values can be retrieved from the workflow, we should report them.)

We may be able to report some of the runner info as resolvedDependencies if the untrusted job reports it, but not sure that's valuable since we don't trust it.

related: I created https://github.com/slsa-framework/slsa/issues/681 to ask for clarification about who populates these

@ianlewis @asraa

ianlewis:
Fixes #1817


laurentsimon (approved):
Let's create another issue and add a TODO in the code to try to add these back later if we find a reliable way to do it

ianlewis:
> Let's create another issue and add a TODO in the code to try to add these back later if we find a reliable way to do it

I think it's fairly straightforward to find out where to add it so I just created an issue: #2027

Issue #1790, opened 2023-03-10
[feature][byob] Support more secrets


ianlewis:
BYOB supports secrets but has a fixed number of secrets it can support. We should support more than the current two secrets.

laurentsimon:
What use case did you encounter for a 3rd secret?

ianlewis:
> What use case did you encounter for a 3rd secret?

So far I haven't encountered a situation where a 3rd secret would be needed, so I think it's fairly low priority but I figure we should support more as adding more secret inputs later would be a breaking change. For example, a user's internal action would fail if the delegator passed `slsa-workflow-secret3` but it wasn't defined in their action.

laurentsimon:
Looks like we need more secrets then! We can add an arbitrary number of secrets. Today we have only 2 secrets supported but it's easy to add more.

> inheritance

Secrets may be inherited only if the re-usable workflow is in the same organization

> The question is, can a PW set secrets in such a way that they get inherited all the way through the chain til they reach the callback action?

Yes. Right now we do that like in https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/nodejs-action/action.yml#L15-L23. The TRW secrets are defined in https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/delegator_generic_slsa3.yml#L33-L44. So the TRW can declare any secrets, and we forward them to the Action, which knows the mapping.

Will that work for your use case?

aalmiray:
Ah OK. So I suppose by convention a given TRW and its companion callback could map `secret1` as the value for `GPG_PASSPHRASE`. It's the callback's responsibility to route secret1 to where it needs to be. 

How many secrets would be needed? hard to tell. Here's an example kitchen-sink repository with more than 40 secrets https://github.com/jreleaser/smoketests-jreleaser/blob/main/.jreleaser/config.toml

JReleaser's official build currently relies on 17 secrets https://github.com/jreleaser/jreleaser/blob/ceea4122057c856767dc3c3d3089679433dfe5dc/.github/workflows/release.yml#L131

laurentsimon:
> Ah OK. So I suppose by convention a given TRW and its companion callback could map secret1 as the value for GPG_PASSPHRASE. It's the callback's responsibility to route secret1 to where it needs to be.

Correct, at least in our design. If you think of something better, let us know.

[...]

PR #2073, opened 2023-04-27
feat: byob: Support 15 secrets

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/1790

I don't think this will break existing callback Actions. It should only log a warning that the input is unknown when calling the Action.


.github/workflows/delegator_generic_slsa3.yml
```diff
@@ -31,16 +31,24 @@ on:
...
+      secret12:
+      secret13:
+      secret14:
+      secret15:
         description: >
-          Optional secret.
+          Optional secrets.
 
-          This argument is passed, unchanged, to the builder's Action's `build-artifacts`.
+          The arguments are passed, unchanged, to the builder's Action's `build-artifacts`.
```
ianlewis, line 51: Maybe just make this a comment. right now only `secret15` has a description.

.github/workflows/delegator_generic_slsa3.yml
```diff
@@ -31,16 +31,24 @@ on:
...
         required: false
```
ianlewis, line 52: Probably we can just omit the `required` field on `secret15` too. `required` defaults to false anyway.

ianlewis (approved):

.github/workflows/delegator_generic_slsa3.yml
laurentsimon: good idea, thanks.

PR #2075, opened 2023-04-28
feat: Add Low perms delegator

ianlewis:
Adds a new low(least?)-permissions delegator that requests minimal GITHUB_TOKEN permissions. We now use this delegator in the Node.js builder.

We also remove functionality for publishing to a GitHub release since most packages don't do that anyway, and we don't have to request `contents:write` permission.


.github/workflows/delegator_low-perms_slsa3.yml
```diff
@@ -0,0 +1,279 @@
...
+        with:
+          name: "${{ needs.rng.outputs.value }}-${{ env.SLSA_ARTIFACTS_FILE }}"
+          path: "${{ env.SLSA_ARTIFACTS_FILE }}"
+
+  # generate-provenance generates and signs the provenance.
+  generate-provenance:
+    needs: [rng, verify-token, privacy-check, build-artifacts-ubuntu]
+    outputs:
+      attestations-download-name: "${{ needs.rng.outputs.value }}-slsa-attestations"
+    permissions:
```
laurentsimon, line 203: nit: add `contents: read` for private repositories

laurentsimon (approved):

.github/workflows/delegator_low-perms_slsa3.yml
```diff
@@ -0,0 +1,279 @@
+# Copyright 2022 SLSA Authors
+#
```
laurentsimon, line 2: Should this still be called generic? `delegator-lowperms_generic_slsa3.yml`, since the workflow is still "generic"? Or we call it `delegator_package_slsa3.yml` since it builds packages

ianlewis: lowperms-generic is ok with me

.github/workflows/delegator_low-perms_slsa3.yml
ianlewis: Refresh my memory. Was `contents:read` needed for upload and download artifacts on private repos?

laurentsimon: ignore what I said. I remembered wrong.

.github/workflows/builder_nodejs_slsa3.yml
```diff
@@ -108,35 +95,26 @@ jobs:
...
 
   slsa-run:
     needs: [slsa-setup]
     permissions:
       id-token: write # For signing.
-      contents: write # For asset uploads.
-      packages: write # For publishing to GitHub packages.
-    uses: slsa-framework/slsa-github-generator/.github/workflows/delegator_generic_slsa3.yml@main
+      contents: read # For repo checkout.
+      actions: read # For getting workflow run on private repos.
```
laurentsimon, line 110: i don't know for sure which perm is needed to get the run https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/verify-token/src/predicate02.ts#L38

What's not clear to me is why we need to get use this API to get the run https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/verify-token/src/predicate02.ts#L77C12-L79

Actor name and id are present in the env variables `GITHUB_ACTOR` and `GITHUB_ACTOR_ID`, so we should be able to get them for free without permissions?

ianlewis: > i don't know for sure which perm is needed to get the run https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/verify-token/src/predicate02.ts#L38
The permission required is documented here:
https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#actions

I was able to test that adding this permission works for a private repo and removing it causes the request to fail.

> What's not clear to me is why we need to use this API to get the run https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/verify-token/src/predicate02.ts#L77C12-L79
> 
> Actor name and id are present in the env variables `GITHUB_ACTOR` and `GITHUB_ACTOR_ID`, so we should be able to get them for free without permissions and without calling the API?

triggering actor id is apparently the actor who triggered the re-run. So user A could trigger a workflow run, and then user B could trigger the re-run. The actor_id would be user A and triggering_user_id would be user B in that case. They currently don't have an env var or github context field for the triggering_actor_id though there is a triggering_actor (username) on the github context.

Mark seemed to think it was important enough to include. 
https://github.com/slsa-framework/github-actions-buildtypes/blob/bc8206c3c0f38e4925285ac9091a18431b010783/workflow/v1/README.md#internal-parameters

I suppose that we could just wait to include it until GHA provides an env var or a github context field but `actions: read` does give us the flexibility to get it if we need to (we also just seem to have it by default for public repos so it's not a lot of extra permissions).

.github/workflows/delegator_lowperms-generic_slsa3.yml
```diff
@@ -129,8 +129,6 @@ jobs:
       artifacts-layout-sha256: ${{ steps.upload.outputs.sha256 }}
     runs-on: ubuntu-latest
     permissions:
-      # NOTE: ideally we should avoid these permissions for build systems that 'run' dependency code
-      # during 'compilation': e.g., npm's install scripts.
       contents: read # For repo checkout.
```
laurentsimon, line 132: you'll need `actions:read` to read run info for private repos?

.github/workflows/delegator_generic_slsa3.yml
```diff
@@ -126,8 +126,7 @@ jobs:
       artifacts-layout-sha256: ${{ steps.upload.outputs.sha256 }}
     runs-on: ubuntu-latest
     permissions:
-      # NOTE: ideally we should avoid these permissions for build systems that 'run' dependency code
-      # during 'compilation': e.g., npm's install scripts.
+      # TODO(#2076): Use dynamic GITHUB_TOKEN permissions.
       contents: write # For release assets.
```
laurentsimon, line 130: you'll need `contents:read` to checkout private repos?

.github/workflows/builder_nodejs_slsa3.yml
laurentsimon: gotcha. Did not realize triggering actor was not set on re-run.

PR #2091, opened 2023-05-02
fix: Update inputs to match softprops/action-gh-release

### LINKED ISSUE #2090: [bug] Workflows update pre-release and draft flags for existing releases
When the release that we upload to already exist the new `draft-release` and `prerelease` flags update the existing release when not passed. We need a way to not update the release if those inputs are not passed to the workflows.

ianlewis:
Fixes #2090

Changes the `prerelease` and `draft-release` flags to strings and passes them directly to `softprops/action-gh-release` in order to preserve the logic of that action.

Also updates `upload-tag-name` to pass that value directly to `softprops/action-gh-release`.


.github/workflows/builder_go_slsa3.yml
```diff
@@ -74,16 +74,20 @@ on:
         required: false
         type: boolean
         default: false
+      # NOTE: `prerelease` and `draft-release` flags have somewhat complicated logic.
+      # - "true" = set to prerelease (case-sensitive, new and existing)
+      # - any other non-empty value = set to regular release (new and existing)
+      # "" = no change for existing releases, false for new releases.
       prerelease:
-        description: "If true, GitHub Release is created as a pre-release."
+        description: "Identify the release as a prerelease. Same as softprops/action-gh-release prerelease flag."
```
laurentsimon, line 82: nit: give an example?

laurentsimon (approved):

.github/workflows/builder_go_slsa3.yml
ianlewis: Updated the description

Issue #1571, opened 2023-01-25
[feature][byob] secure upload unsigned attestations


laurentsimon:

laurentsimon:
This could be using the secure-upload-folder too

PR #2096, opened 2023-05-02
feat: byob: secure upload attestation and public download Action

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/1571


actions/delegator/secure-attestations-download/action.yml
```diff
@@ -0,0 +1,23 @@
+name: "Secure attestations download"
+description: "Download a set of attestations and verify their SHA256"
```
laurentsimon, line 2: is this path acceptable for a public Action? Will "delegator" be confusing to end-users?

actions/nodejs/secure-attestations-download/action.yml
```diff
@@ -0,0 +1,23 @@
+name: "Secure attestations download"
```
laurentsimon, line 1: Creating a dedicated Action that matches the "nodejs" builder end-users will be familiar with. I think this also allows us to update a builder without breaking others.

actions/nodejs/secure-attestations-download/action.yml
```diff
@@ -0,0 +1,23 @@
+name: "Secure attestations download for nodejs builder"
```
ianlewis, line 1: I realize we only have the nodejs builder right now but I think most BYOB workflows could use this, no?

Also what situation were we expecting folks to use this? I guess if they need to do npm publish on their own?

.github/workflows/scripts/pre-submit.delegators/expected.diff
```diff
@@ -0,0 +1,16 @@
+15c15,19
+< name: SLSA builder delegator
```
laurentsimon, line 2: ignore this file, I need to rebase

actions/nodejs/secure-attestations-download/action.yml
laurentsimon: yes all folks can use it. I'm hesitant to create one for all ecosystems. If we ever need to change this, at least we'll be able to break only one of the builders, not all.

ianlewis: Yeah, it's fine for now I guess.

.github/workflows/builder_nodejs_slsa3.yml
```diff
@@ -108,7 +108,7 @@ jobs:
       id-token: write # For signing.
       contents: read # For repo checkout of private repos.
       actions: read # For getting workflow run on private repos.
-    uses: slsa-framework/slsa-github-generator/.github/workflows/delegator_lowperms-generic_slsa3.yml@main
+    uses: laurentsimon/slsa-github-generator/.github/workflows/delegator_lowperms-generic_slsa3.yml@feat/sec-upload-att
```
ianlewis, line 111: FIXME

.github/workflows/builder_nodejs_slsa3.yml
```diff
@@ -127,9 +127,10 @@ jobs:
       # NOTE: The attestation name has a rng value in it's name so we don't need to check for file collisions.
       - name: Download attestations
         id: attestation-download
-        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+        uses: laurentsimon/slsa-github-generator/actions/nodejs/secure-attestations-download@feat/sec-upload-att
```
ianlewis, line 130: FIXME

.github/workflows/builder_nodejs_slsa3.yml
```diff
@@ -127,9 +127,11 @@ jobs:
       # NOTE: The attestation name has a rng value in it's name so we don't need to check for file collisions.
       - name: Download attestations
         id: attestation-download
-        uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
+        uses: slsa-framework/slsa-github-generator/actions/nodejs/secure-attestations-download@main
         with:
```
laurentsimon, line 131: note: end-uses will need to use a similar command. We need to tell them that the name of the folder containing all the attestations is `${{ needs.slsa-run.outputs.attestations-download-name }}`.

@ianlewis is that OK in terms of UX?

laurentsimon: you said you'll provide an Action as well to simplify

ianlewis: Yeah, this is fine. I have some plans for simplification.

I'd like to make the artifact uploads and the file names in the upload match if possible, and provide an action for publishing to make it easier.

ianlewis (approved):
No delegator diff change?

laurentsimon:
> No delegator diff change?

apparently not.. because I changed the exact same lines in both workflows so they match.

laurentsimon:
> > No delegator diff change?
> 
> apparently not.. because I changed the exact same lines in both workflows so they match.

And I'm obviously wrong. pre-submit is failing :-)

laurentsimon:
I marked the delegator pre-submit as required in branch protection settings.

PR #2097, opened 2023-05-02
feat: pre-submit to detect differences between delegator workflows

laurentsimon:


ianlewis (approved):
Thanks! I guess we'll probably have to update the diff each time but I guess that's ok.

.github/workflows/scripts/pre-submit.delegators/compare-diff.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+set -euo pipefail
+
+cat <<EOF >DIFF
```
ianlewis, line 5: nit: Put the diff in a .diff file and read the file here. That will make it easier to update it.

PR #2098, opened 2023-05-03
fix: Update the use of SLSA v1 provenance in the docker-based tool

rbehjati:
This PR updates the Docker-based builder to use the new SLSA v1 structs.


rbehjati:
@asraa Does this change make sense? 

I ran `go get github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v1`, but it seems like it has affected some unrelated packages.

asraa (commented):
Yes it does! It should be pulling in the new version of go-securesystemslib - I think it may be that the sigstore dep needs to also be updated to latest as well because of the securesystemslib update.

rbehjati:
I think this is no longer relevant.

Issue #2079, opened 2023-04-28
[feature] Add more tests for TRW's specific commit sha1


laurentsimon:
We need to verify that that generated  provenance is correct. Unit tests and scheduled tests within this repo.
The feature was introduced in https://github.com/slsa-framework/slsa-github-generator/pull/2078

laurentsimon:
this should include verification for both v0.2 and v1.0 predicates

laurentsimon:
Removing this issue from the BYOB milestone, because we want to support v1.0 first.

PR #2135, opened 2023-05-17
feat: add test for checkout's sha1 option of verify-token

laurentsimon:
This is https://github.com/slsa-framework/slsa-github-generator/issues/2079 for v0.2 provenance.


.github/actions/verify-token/src/utils.ts
```diff
@@ -58,14 +60,30 @@ export function getTriggerRef(rawTokenObj: rawTokenInterface): string {
...
 
+function validateSha1(sha1: string): void {
+  // Length of sha1 in hex format must be 40.
+  if (sha1.length !== 40) {
+    throw new Error(`invalid sha1 length: ${sha1.length}`);
+  }
+
+  // Check if the string only contains hexadecimal characters.
+  for (const c of sha1) {
+    if (!/[a-fA-F0-9]/.test(c)) {
```
ianlewis, line 71: you can add the length check to the regex
```suggestion
    if (!/[a-fA-F0-9]{40}/.test(c)) {
```

.github/actions/verify-token/dist/index.js
```diff
@@ -50997,14 +51015,28 @@ function getTriggerRef(rawTokenObj) {
     return rawTokenObj.github.ref;
 }
 exports.getTriggerRef = getTriggerRef;
+function validateSha1(sha1) {
```
ianlewis, line 51018: not a real issue I think but strange that this shows up twice in the compiled dist file.

ianlewis (approved):

.github/actions/verify-token/dist/index.js
laurentsimon: I re-generated but still appearing multiple times.

PR #2132, opened 2023-05-16
feat: Add BYOB Gradle builder

AdamKorcz:
Adds a BYOB Gradle builder.

This is currently added to sigstore-java: https://github.com/sigstore/sigstore-java/blob/main/.github/workflows/byob-slsa.yaml

Latest Sigstore-java action for the Gradle builder: https://github.com/AdamKorcz/sigstore-java/actions/runs/4991352365 which builds the artifacts and generates the provenance statement.

Issue #2060, opened 2023-04-25
[feature][byob] Remove private inputs from provenance


laurentsimon:
We had an issue about it, but I could not find it so I'm creating this new one.

1We need to mask private fields of GH context for privacy reasons: username, etc. Or, alternatively, we don't record the entire context since only a few triggers are meant to be supported by v1.0 and we can record each event's specific information instead. Note: removing the GH context has implications for branch / tag verification, so the logic would need to change.

ianlewis:
We have a number of TODOs in the code pointing to #1575 relating to masking data in provenance. For v1.0 provenance we currently include the entire event object which is problematic.
https://github.com/slsa-framework/slsa-github-generator/blob/8f12c7b08b0fde2eb6580658797e23e56198c9c0/.github/actions/verify-token/src/predicate1.ts#L109-L115

laurentsimon:
https://github.com/slsa-framework/slsa-github-generator/issues/1575 was more about TRW-specific sensitive inputs. The GH payload we record is the one we need to address next before BYOB release.

PR #2144, opened 2023-05-21
feat: Golang slsa module 0.2 to 1.0 migration

houdini91:
HI all,
Trying to lend a hand by migrating the golang slsa module to SLSA 1.0.
Basically i tried to follow the https://slsa.dev/provenance/v1  section Migrating from 0.2 to 1.0. 

I am a first time contributor so take your time reviewing my code and i appreciate any feedback or comments.

Fixes https://github.com/slsa-framework/slsa-github-generator/issues/1525

### DIFF
--- a/internal/builders/go/pkg/provenance.go
+++ b/internal/builders/go/pkg/provenance.go
@@ -24,6 +24,8 @@ import (
 
 	intoto "github.com/in-toto/in-toto-golang/in_toto"
 	slsacommon "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/common"
+	slsa1 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v1"
+
 	"github.com/slsa-framework/slsa-github-generator/github"
 	"github.com/slsa-framework/slsa-github-generator/internal/utils"
 	"github.com/slsa-framework/slsa-github-generator/slsa"
@@ -150,22 +152,22 @@ func GenerateProvenance(name, digest, command, envs, workingDir string,
 	//
 	// NOTE: map is a reference so modifying invEnv modifies
 	// p.Predicate.Invocation.Environment.
-	invEnv, ok := p.Predicate.Invocation.Environment.(map[string]interface{})
+	invEnv, ok := p.Predicate.BuildDefinition.InternalParameters.(map[string]interface{})
 	if !ok {
-		panic(fmt.Sprintf("converting %T to map[string]interface{}", p.Predicate.Invocation.Environment))
+		panic(fmt.Sprintf("converting %T to map[string]interface{}", p.Predicate.BuildDefinition.InternalParameters))
 	}
 	invEnv["arch"] = os.Getenv("RUNNER_ARCH")
 	invEnv["os"] = os.Getenv("ImageOS")
 
 	// Add details about the runner's OS to the materials
-	runnerMaterials := slsacommon.ProvenanceMaterial{
+	runnerMaterials := slsa1.ResourceDescriptor{
 		// TODO: capture the digest here too
 		URI: fmt.Sprintf(
 			"https://github.com/actions/virtual-environments/releases/tag/%s/%s",
 			os.Getenv("ImageOS"), os.Getenv("ImageVersion"),
 		),
 	}
-	p.Predicate.Materials = append(p.Predicate.Materials, runnerMaterials)
+	p.Predicate.BuildDefinition.ResolvedDependencies = append(p.Predicate.BuildDefinition.ResolvedDependencies, runnerMaterials)
 
 	if utils.IsPresubmitTests() {
 		fmt.Println("Pre-submit tests detected. Skipping signing.")
--- a/slsa/buildtype.go
+++ b/slsa/buildtype.go
@@ -23,8 +23,7 @@ import (
 
 	intoto "github.com/in-toto/in-toto-golang/in_toto"
 	slsacommon "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/common"
-	slsa "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
-	slsa02 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v0.2"
+	slsa1 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v1"
 
 	"github.com/slsa-framework/slsa-github-generator/github"
 )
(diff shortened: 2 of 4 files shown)


ianlewis:
Thanks for this!

A couple general comments:

Some downstream consumers of the provenance might still need to consume v0.2. Could we make this an option (similar to the [`slsa-version` input](https://github.com/slsa-framework/slsa-github-generator/blob/1790bc7b4aa7debcc81f43bb3cfe204550301ee1/actions/delegator/setup-token/action.yml#L6-L9) in `slsa-token` action)? /cc @laurentsimon 

Please add "Fixes #1525" to your PR description.

houdini91:
@ianlewis Do you think this versioning input flag capability should be part of this PR or another?

I am probably two new to this project to answer how this should be done

But anyhow here is are my thoughts ..

1) Have multiple builder options for each version of slsa in the repo side by side, Action Input flag select between the two implementations.

2) In workflow ref selection process , For each version of SLSA autoselect the ref (Branch, Tag?) the builder uses to bulid the generator binary (i am not sure how to autoselect this).. 

3) Document the action tag to use for each version ... not really a input flag like you wanted..

4)  Add the https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/generic As a Embedded Sumodule versioning it aligned to the SLSA versions. Builder selects the submodule for build using the input_flag directly.

laurentsimon:
Hi @houdini91, thanks for sending this PR. Migrating to v1.0 seems like a seamless change, but it's more complicated that it seems. There are a few things that need to happen for a complete solution, which probably needs a series of PRs
1. I think we need a new workflow for the migration. Otherwise we won't be able to release other workflows / updates until migration is complete - because it may be considered a breaking change
2. We currently don't explicitly "separate" inputs based on the trigger, but the examples for v1.0 does https://github.com/slsa-framework/github-actions-buildtypes/tree/main/workflow/v1. We may want to follow this.
3. Remove privacy inputs https://github.com/slsa-framework/slsa-github-generator/issues/2060
4. We need to record the inputs to the builder itself (https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/builder_go_slsa3.yml#L38-L98), because that's the interface for v1.0. We no longer need to record the compiler commands run by the compiler, which will greatly simplify the workflow.

The good news is that we have a new framework to do all this for all builders at once https://github.com/slsa-framework/slsa-github-generator/milestone/11. This means we need to re-write the workflow. You can seen an example for nodejs https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/builder_nodejs_slsa3.yml and the underlying workflow it's calling to handle provenance generation https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/delegator_lowperms-generic_slsa3.yml. The SLSA predicates end up being generated in https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/verify-token/src/index.ts#L171-L188. Points 2, 3 and 4 above are taken care of in this code (or will) so you don't have to worry about them yourself. All you need is use this framework to implement point 1.

If you're interested, we can share a Google doc with you to help you understand how how the framework works (we also have examples for other ecosystems, e.g. [JReleaser](https://github.com/jreleaser/jreleaser-java-slsa-generator/blob/main/.github/workflows/trw.yml), [Gradle](https://github.com/slsa-framework/slsa-github-generator/pull/2132)). We're working on documentation for this framework which hopefully will be ready next month.

houdini91:
Ok i totally missed this https://github.com/slsa-framework/github-actions-buildtypes/tree/main/workflow/v1
I try and write down what i understood (more for me that for you guys).

> 2. We currently don't explicitly "separate" inputs based on the trigger
Not sure i understood? Filter out any event that is not in the `Support event table` (for v1), should this be done in the workflow or code level? 

> 3.  Remove privacy inputs
Understood i think, Some filter of the event and inputs needs to be done somehow.

4.  We need to record the inputs to the builder itself 
Understood i think, So we need to add the Workflow input to the v1 prov.

> The good news is that we have a new framework to do all this for all builders at once
> Points 2, 3 and 4 above are taken care of in this code
What do you mean in this code ? (This PR, The Code you refer to for ts builder?, BYOB?)
So you mean the BYOB will take of these points or i should update this PR to match these requirements.

> If you're interested, we can share a Google doc with you to help you understand ...
Please do, i am a bit lost.

General question:
BYOB - This framework is going to allow SLSA-lv3 using there own build action by providing workflow that wraps the build action with SLSA prov generation (key and generation and so on) ?
As Part of the framework will already provide workflows that separate different versions of SLSA and builders, Did i understand correctly?

Personal Side note:
Sorry for the probably stupid questions, I guess i just starting to understand how magnificently complex SLSA lv3 requirements have on actually build system.
In my head LV3 was only forcing the builder to attest to the Artefacts Built apon it, not actually building the Artifact.

laurentsimon:
> Ok i totally missed this https://github.com/slsa-framework/github-actions-buildtypes/tree/main/workflow/v1 I try and write down what i understood (more for me that for you guys).
> 
> > 2. We currently don't explicitly "separate" inputs based on the trigger
> 
> Not sure i understood? Filter out any event that is not in the `Support event table` (for v1), should this be done in the workflow or code level?

The provenance output needs to be:
```json
externalParameters{
  inputs: {
    <name>: <value>
  }
}
```
and currently it's:
```json
externalParameters{
  <name>: <value>
}
```
> > 3. Remove privacy inputs
> 
> Understood i think, Some filter of the event and inputs needs to be done somehow.

Correct, we have a tracking issue to remove usernames https://github.com/slsa-framework/slsa-github-generator/issues/2060

> > 4. We need to record the inputs to the builder itself

> Understood i think, So we need to add the Workflow input to the v1 prov.

Correct. The BYOB framework makes this easy. You just do https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/builder_nodejs_slsa3.yml#L99
 
> > The good news is that we have a new framework to do all this for all builders at once
> > Points 2, 3 and 4 above are taken care of in this code
> 
> What do you mean in "this code" ? (This PR, The Code you refer to for ts builder?, BYOB?) So you mean the BYOB will take of these points or i should update this PR to match these requirements.

code = BYOB framework. Sorry for the confusion.

> > If you're interested, we can share a Google doc with you to help you understand ...
> 
> Please do, i am a bit lost.

Do you have an email address I can share the doc with?

> General question: BYOB - This framework is going to allow SLSA-lv3 using there own build action by providing workflow that wraps the build action with SLSA prov generation (keys, delegation and other magic) ? did i understand the premis? 

Correct

> As Part of the framework will already provide workflows that separate different versions of SLSA and builders, Did i understand correctly?
not sure what you mean exactly. But I think yes it will take care of supporting multiple version and creating the right format. See https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/verify-token/src/predicate1.ts for v1, and https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/verify-token/src/predicate02.ts for v0.2

> Personal Side note: Sorry for the probably stupid questions, I guess i just starting to understand how magnificently complex SLSA lv3 requirements have on actually build system. In my head LV3 was only forcing the builder to attest to the Artefacts Built apon it, not actually building the Artifact.

no problem. I think we can help you design the workflow for the Go builder to use BYOB.. and get v1.0 support. If you're up for it we're happy to chat and provide guidance.

laurentsimon:
Closing this PR since these changes are more complicated.

PR #2156, opened 2023-05-23
feat: Support go-version-file

### LINKED ISSUE #1807: [feature] Support go-version-file in generate-builder
Support the [`go-version-file`](https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file) input in `generate-builder` action to get the Go version from the `go.mod` file.

We have a lot of references to the Go version so it would be good to reduce that and manage it from the go.mod file.

ianlewis:
Fixes #1807


laurentsimon (approved):

.github/workflows/builder_go_slsa3.yml
```diff
@@ -37,7 +37,11 @@ on:
     inputs:
       go-version:
         description: "The go version to use."
-        required: true
+        required: false
```
ianlewis, line 40: @laurentsimon Do we even really care too much about specifying the Go version for project checkout? I guess just so that it's recorded as an input to the Go builder?

Do we record the inputs for the Go builder separately from the user workflow?

laurentsimon: > @laurentsimon Do we even really care too much about specifying the Go version for project checkout? 

It's good to have users hardcode the version at least, it's similar to pinning. Otherwise once we switch to v1.0 provenance, the info will be lost.

> I guess just so that it's recorded as an input to the Go builder?

Right now it's implicitly recorded via the compiler command we record. In v1, we will need to record these inputs properly. In v0.2 it was not needed, and we opted to record the lower-level representation with the compiler commands instead.

> Do we record the inputs for the Go builder separately from the user workflow?

In v0.2 no. In v1,0 we will need to.

Let me know if this makes sense or not. I could be forgetting something...

ianlewis: > Let me know if this makes sense or not. I could be forgetting something...

No this is my understanding as well. Just wanted to make sure.

PR #2158, opened 2023-05-23
chore: Add actionlint linter

### LINKED ISSUE #1896: [feature][ci] Add an actionlint pre-submit
https://github.com/rhysd/actionlint

Actionlint will run shellcheck on inline scripts and has checks for script injection that might be useful for us.

### LINKED ISSUE #2056: [feature] Remove testing flag from generate-builder
After v1.6.0 is released we should remove the `testing` flag from `generate-builder` and use the `v1.6.0` release for pre-submit testing.

ianlewis:
Fixes #1896
Fixes #2056 

Adds the [`actionlint`](https://github.com/rhysd/actionlint) linter. This allows us to check beyond just yaml formatting and lint Actions workflows against the [Github Actions workflow syntax](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions) and other [checks](https://github.com/rhysd/actionlint/blob/v1.6.24/docs/checks.md). Also runs `shellcheck` for inlined bash code in workflows.

Currently doesn't support `actions.yml` format for composite actions yet though.

Adds a new `Makefile` target.

```shell
make actionlint
```


.github/workflows/delegator_lowperms-generic_slsa3.yml
```diff
@@ -176,7 +176,7 @@ jobs:
       - name: Checkout the project repository
         uses: slsa-framework/slsa-github-generator/.github/actions/secure-project-checkout@main
         with:
-          fetch-depth: ${{ toJson(needs.verify-token.outputs.slsa-verified-token).source.checkout.fetch_depth }}
```
laurentsimon, line 179: good catch! was this not working before then? But we have some scheduled test...

.github/workflows/pre-submit.actions.yml
```diff
@@ -498,17 +499,14 @@ jobs:
         uses: ./.github/actions/detect-workflow-js
       - uses: ./.github/actions/generate-builder
         with:
-          repository: ${{ steps.detect.output.repository }}
+          repository: ${{ steps.detect.outputs.repository }}
```
laurentsimon, line 502: good catch!

laurentsimon (approved):

.github/workflows/pre-submit.actions.yml
ianlewis: Yeah, these are the kinds of things that the `actiontest` tool can find so it's pretty useful.

.github/workflows/delegator_lowperms-generic_slsa3.yml
laurentsimon: in this workflow https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/e2e.verify-token.reusable.yml#L55 which is verified in https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/e2e.verify-token.reusable.yml#L54

same for `slsa-checkout-sha1`

ianlewis: Yeah, I suppose this passes because the input was basically ignored and it was checked out with the default fetch-depth value. We don't seem to actually verify that the checkout happened at the proper depth in a wrapped builder action or anything as part of the tests.

PR #2165, opened 2023-05-23
feat: remove `externalParameters.workflow` for builders

laurentsimon:


.github/actions/verify-token/src/index.ts
```diff
@@ -62,11 +62,22 @@ async function run(): Promise<void> {
     if (!ghToken) {
       throw new Error("token not provided");
     }
+    const builderInterfaceType = core.getInput("builder-interface-type");
```
ianlewis, line 65: You can pass a required option to get input and it will throw an error if it wasn't supplied

.github/actions/verify-token/action.yml
```diff
@@ -31,7 +31,11 @@ inputs:
     default: ${{ github.token }}
 
   output-predicate:
-    description: "The filename to place the output SLSA predicate based on the verified token"
+    description: "The filename to place the output SLSA predicate based on the verified token."
+    required: true
+
+  builder-interface-type:
```
ianlewis, line 37: Could the default just be "builder"?

ianlewis (approved):

.github/actions/verify-token/src/index.ts
laurentsimon: done, thanks

.github/actions/verify-token/action.yml
laurentsimon: we could, but I thought making it explicit was better. Wdut?

Issue #2164, opened 2023-05-23
[feature][byob] Record inputs for generators


laurentsimon:
We currently don't record the inputs for generators in the `externalParameters` in `verify-token`

PR #2174, opened 2023-05-24
feat: disable v1 support for generators

laurentsimon:
This will ensure we don't forget to add support when adding support for generators in v1.0 in the generator and verifier.


.github/actions/verify-token/dist/index.js
```diff
@@ -581,6 +581,9 @@ function createPredicate(rawTokenObj, toolURI, token, isGenerator) {
                 // TODO(#2164): record the inputs, depending on the type of trigger events.
                 inputs: {},
             };
+            // Throw an error for now. We have no generators using v1.0 yet
+            // and it's not supported in the slsa-verifier.
+            throw new Error("not supported: #2164, #2164");
```
ianlewis, line 586: #2164, #2164? Was this supposed to be two different issues?

ianlewis (approved):

.github/actions/verify-token/dist/index.js
laurentsimon: good catch. Fixed.

Issue #2182, opened 2023-05-26
[feature] Use SLSA resource description's annotations to specify whether the resolvedDependencies were a source or other material


asraa:
The container-based builder uses two resolved deps: one is the source, and one is the builder binary from this repository. 

They are both represented as a ResourceDescriptor in the unordered `ResolvedDependencies` field of the SLSA v1.0 provenance.

My suggestions is to use an annotation "source" to hint at which was the `source`, so the slsa-verifier can make an informed decision.

We can keep the default behavior the same - that is, if there is only 1 resolved dep, use that as the source. if there are more than 1, we can use the annotations to narrow the list down.

This is a string-string map, so we can do something like

```
      "resolvedDependencies": [
        {
          "uri": "git+https://github.com/slsa-framework/slsa-github-generator@refs/heads/main",
          "digest": {
            "sha256": "d5a35d5f711ac09f4cc661ccc42a9c3f9a29a3b140b59239d6c193606063de45"
          }
        },
        {
          "uri": "git+https://github.com/slsa-framework/example-package@refs/heads/main",
          "digest": {
            "sha1": "b11ac9395b936cc5afcbc45112f48d66d85788b7"
          },
          "annotations":{
                "source": "true"
          }
         }
      ],
```

the verifier can check presence of a `source` key in the annotation map with the value `true`.

What do you think?

laurentsimon:
SG. Let's add the same for BYOB then

PR #2183, opened 2023-05-26
fix: add source annotations to container-based predicate's resolveddependencies

asraa:
See https://github.com/slsa-framework/slsa-github-generator/issues/2182

This allows verifiers to detect which resolved dependency corresponds to a source material. It adds an annotation like the following:

```
      "resolvedDependencies": [
        {
          "uri": "git+https://github.com/asraa/slsa-github-generator@refs/heads/main",
          "digest": {
            "sha1": "f6233c5dcfcfb8c7ef8cce768959c11771a3072a"
          },
          "annotations": {
            "source": "true"
          }
        }
      ],
```

This also fixes the e2e tests in conjunction with the verifier PR.

Also took the opportunity to clean up the predicate's internal paremeters - no more github runner or other invalid fields.


laurentsimon (commented):
LGTM overall. I'm curious why we need an annotation if the buildType always adds it as the first resolvedDependencies instead.

asraa:
> LGTM overall. I'm curious why we need an annotation if the buildType always adds it as the first resolvedDependencies instead.

We can do that, but is that too brittle? the resolved dependencies is explicitly supposed to be an unordered collection according to slsa spec.

asraa:
Per resolution offline, let's keep uniformly as first resolvedDeps until we have a universal solution.

laurentsimon (approved):

Issue #2186, opened 2023-05-27
[feature] Ass `source:URI` under externalParameters`


laurentsimon:

asraa:
I think it should be a resource descriptor - we get the URI and the hash explicitly, and can also add new fields.

I think they should rephrase the migration as 

```
"source": {
   "uri": old.invocation.configSource.uri,
}
```

That migration note was made before they switched to ResourceDescriptors.

kpk47:
I agree with @asraa that source should be a resourceDescriptor. The provenance spec recommends string values in externalParameters, but they aren't required. 

> The externalParameters and internalParameters are the top-level inputs to the template, meaning inputs not derived from another input. Each is an arbitrary JSON object, though it is RECOMMENDED to keep the structure simple with string values to aid verification. The same field name SHOULD NOT be used for both externalParameters and internalParameters.

Since there's a benefit to using a resourceDescriptor, go for it. Just make sure to keep the fields in sync between the generator and the verifier.

laurentsimon:
> I think it should be a resource descriptor - we get the URI and the hash explicitly, and can also add new fields.
> 
> I think they should rephrase the migration as
> 
> ```
> "source": {
>    "uri": old.invocation.configSource.uri,
> }
> ```
> 

and maybe also have:
```json
"source": {
    "uri": old.invocation.configSource.uri,
    "digest": old.invocation.configSource.digest,
}
```

kpk47:
I support that change. I've opened an issue on the SLSA spec repo to get input.

ianlewis:
Just to be clear, this is a totally separate Resource Descriptor that's separate from the entry in `resolvedDependencies`?

asraa:
> Just to be clear, this is a totally separate Resource Descriptor that's separate from the entry in resolvedDependencies?

I understood it to be duplicated, since it is (1) a user input and (2) a resolved dependency pulled into the environment by the builder.

But yeah, actually just the source URI might be a user input, while the resolved dependency may contain the extra info like the digest that's resolved when the builder pulls it. So dupes doesn't really feel right either.

ianlewis:
> I understood it to be duplicated, since it is (1) a user input and (2) a resolved dependency pulled into the environment by the builder.
> 
> But yeah, actually just the source URI might be a user input, while the resolved dependency may contain the extra info like the digest that's resolved when the builder pulls it. So dupes doesn't really feel right either.

[...]

PR #2193, opened 2023-05-30
feat: Add `externalParameters.source` for v1 specs for BYOB

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/2186


.github/workflows/scripts/schedule.actions/verify-verified-token.sh
```diff
@@ -40,5 +40,6 @@ e2e_verify_predicate_v1_runDetails_builder_id "$PREDICATE_CONTENT" "https://gith
 e2e_verify_predicate_v1_buildDefinition_externalParameters_workflow "$PREDICATE_CONTENT" "$(e2e_this_file_full_path)" "$GITHUB_REF" "git+https://github.com/$GITHUB_REPOSITORY"
 
 # Verify external parameters source
-e2e_verify_predicate_v1_buildDefinition_externalParameters_source "$PREDICATE_CONTENT" "{\"uri\":\"git+https://github.com/$GITHUB_REPOSITORY@$GITHUB_REF\",\"digest\":{\"sha1\":\"$GITHUB_SHA\"}}"
+digest=$(e2e_get_source_sha1)
+e2e_verify_predicate_v1_buildDefinition_externalParameters_source "$PREDICATE_CONTENT" "{\"uri\":\"git+https://github.com/$GITHUB_REPOSITORY@$GITHUB_REF\",\"digest\":{\"sha1\":\"$digest\"}}"
```
laurentsimon, line 44: this check should have been failing. The daily runs were not working due to permission errors https://github.com/slsa-framework/slsa-github-generator/actions/runs/5125842921
I've tryied addrssing this in this PR by changing the default perms to `permissions: {}`.

ianlewis:
We had a discussion about adding an identifier for the source repo to the `annotations` field in `resolvedDependencies`? What do you think about that idea?

asraa:
> We had a discussion about adding an identifier for the source repo to the annotations field in resolvedDependencies?

I think if we do, it'd be great to have a universal annotation that's agreed upon between multiple builders, so that parsing the arbitrary `object` can be shared logic.

Since annotations are maps of strings to objects, maybe it can be `usage: <some-source-type-uri>`?

ianlewis:
> I think if we do, it'd be great to have a universal annotation that's agreed upon between multiple builders, so that parsing the arbitrary `object` can be shared logic.

Yeah, I agree. It should probably get into the https://github.com/slsa-framework/github-actions-buildtypes repo at the very least.

> Since annotations are maps of strings to objects, maybe it can be `usage: <some-source-type-uri>`?

For reference, Kubernetes has a similar type of annotations (though string: string) and they use URIs in the keys with "official" annotations starting with using a `kubernetes.io` domain.
https://kubernetes.io/docs/reference/labels-annotations-taints/

laurentsimon:
> Maybe SLSA needs a few of these well known annotations.

I'm curious what's the point of these annotation in practice. iiuc, the verification is only required to look at `externalParameters.source`, and anything in material is to do recursive verification of dependencies...

laurentsimon:
There is WIP for specs to resolve this, so closing for now

PR #2197, opened 2023-05-31
refactor: Update ResourceDescriptor to in-toto attestation v1.0

ianlewis:
Updates the `ResourceDescriptor` type to the in-toto attestation v1.0 spec.
https://github.com/in-toto/attestation/blob/main/spec/v1/resource_descriptor.md


asraa (approved):

.github/actions/verify-token/src/predicate1.ts
```diff
@@ -81,6 +81,7 @@ export async function createPredicate(
       },
       resolvedDependencies: [
         {
+          name: "sourceRepo",
```
laurentsimon, line 84: is this a standard annotation in the specs?

asraa: Sort of, it can be:

```
Machine-readable identifier for distinguishing between descriptors.

The semantics are up to the producer and consumer. The name SHOULD be stable, such as a filename, to allow consumers to reliably use the name as part of their policy.
```

So `sourceRepo` can distinguish between descriptors, and can allow verifiers to check against policy. It is not however the filename convention mentioned in there

ianlewis: It's not super important so we could just leave it off. The rest of the change is still valid.

.github/actions/create-container_based-predicate/src/predicate.ts
```diff
@@ -37,12 +42,19 @@ export interface Metadata {
...
-  uri: string;
-  digest: DigestSet;
-  localName?: string;
+  name?: string;
+  uri?: string;
+  digest?: DigestSet;
+  content?: Uint8Array; // content is bytes.
   downloadLocation?: string;
   mediaType?: string;
+  annotations?: { [key: string]: string };
```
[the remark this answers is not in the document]
ianlewis: Yep you're right. I did see that. It seems that a JSON object would preclude values from being strings though. Updated.

ianlewis:
> Shall we wait to see if we need these attestations before implementing? iiuc, we don't need them for our builders, do we?

You're right it's not a change that will affect us yet since we currently only use `uri` and `digest` and those fields haven't changed. I figured we could just go ahead and update the type anyway. Setting the `name` field isn't really the best way to identify the source dependency so we can wait for a definitive answer on that issue.

.github/actions/verify-token/src/predicate1.ts
laurentsimon: I vote for not adding it and see if we need it in the future.
The intoto specs gives us the ability, but it'd be nice if SLSA standardized some of the common names, and give some guidance about how it's supposed to be used @kpk47

ianlewis: Yep. Already removed it.

PR #2203, opened 2023-06-02
fix: Allow no inputs to TRW

### LINKED ISSUE #1780: [feature][byob] Add tests for empty TRW inputs
Ensure it works with empty TRW inputs, e.g. with and e2e test
I don't think this case would happen in practice, though

ianlewis:
Fixes #1780


laurentsimon (approved):
do you know what the provenance will look like in this case? `inputs:{}`?

.github/actions/verify-token/__tests__/inputs.test.ts
```diff
@@ -20,6 +20,19 @@ import { updateSLSAToken } from "../src/inputs";
 import { rawTokenInterface } from "../src/types";
 
 describe("updateSLSAToken", () => {
+  it("no inputs", async () => {
+    const inputs = JSON.parse("{}");
+    const token = createToken(inputs);
+    const content = `
```
laurentsimon, line 26: can you add a unit tests with no workflow_call? I don't think we have one

laurentsimon: ignore me. We do have one https://github.com/slsa-framework/slsa-github-generator/blob/b25fcf20ac7cc3869ebf1572bc578552a3548d20/.github/actions/verify-token/__tests__/inputs.test.ts#L178-L180

ianlewis:
> do you know what the provenance will look like in this case? `inputs:{}`?

Yep.

PR #2212, opened 2023-06-02
feat: rename setup-token to setup-generic (slsa-framework#2176)

enteraga6:
Gradle and Maven builders referenced setup-token. Changed to use setup-generic now.


laurentsimon (approved):
Good catch! Thank you!

PR #2213, opened 2023-06-02
fix: verify-token daily runs

laurentsimon:


ianlewis (approved):
You'll want to address the linter errors. Mostly adding `type` to inputs and deduping the `name1` input.

PR #2220, opened 2023-06-05
fix: set extension of attestation as ".build.slsa"

laurentsimon:
It's currently just "<fn>.sigstore"

closes https://github.com/slsa-framework/slsa-github-generator/issues/2230


asraa (commented):
Is this right? I think it shouldn't be an intoto.jsonl file -it's a sigstore file, with an intoto statement containted inside one of the fields of the sigstore bundle message.

laurentsimon:
ah, you're right.  dont know what we should call these. My worry is that extensions keep changing (used to be `intoto.jsonl`, now we use `.sigstore`) and may confuse users: they should not have to care about how the attestation is represented, should they? 

I'd like to have a `.<whatever>.slsa`, so `.intoto.jsonl.slsa`, `.sigstore.slsa`, etc... or even a `.slsa`...

@TomHennen any thoughts?

asraa:
> they should not have to care about how the attestation is represented, should they?

I think they might need to here: we're actually distributing a signed bundle, that contains the attestation inside - so if they want to extract it out and treat it like normal attestation data they would need to know the bundle format. WDYT?

That being said, I heard that maybe @haydentherapper is proposing the sigstore bundle more formally as a signature bundle wrapper?

haydentherapper:
I’ve seen it called “.bundle” but I’d avoid that so that when we start outputting protobuf bundles, we can use it for that. Don’t have a great idea for the ext though. 

Yea, we were chatting about using a dsse envelope to store Sigstore metadata, but I wouldn’t block on that.

ianlewis:
I think `.sigstore` is ok, though I think it would be nice if it was `.sigstore.json` or `.sigstore.jsonl` to denote the encoding.

ianlewis:
> I don't know how feasible any of this is, but I know that I'd like to be able to publish in-toto attestations from various signers (not just sigstore) to registries and other storage locations. Do we have any story for how to do that?

Currently it's a bit flimsy.

Our currently released builders/generators support uploading the `.intoto.jsonl` (correct in this case) files as a Workflow run artifact or GitHub release asset. This is probably the nicest story so far since the files can be moved and copied around easily and the format isn't particularly tied to sigstore (though that's the only signing method we support at the moment). However, we do have to have a bit of a bespoke way to get the tlog entries as there are no hints in the intoto attestation, and an "offline" version of the tlog entry is not included.

BYOB, which all of our "next-generation" builders are based off of, is currently working with sigstore bundles. This is primarily because we are using `sigstore-js` and that works with sigstore bundles, [supports "offline" verification only](https://github.com/sigstore/sigstore-js/blob/3e770f048ed57b70ce7b4ae60ef3b08de9de3718/packages/client/src/tlog/verify/index.ts#L30), and requires the tlog entry to be included in the bundle (unless we omit tlog entry verification altogether). As it happens, this is also what `npm` uses and the official registry supports sigstore bundles in the package metadata so it was also easy to integrate for the Node.js builder.

laurentsimon:
> > they should not have to care about how the attestation is represented, should they?
> 
> I think they might need to here: we're actually distributing a signed bundle, that contains the attestation inside - so if they want to extract it out and treat it like normal attestation data they would need to know the bundle format. WDYT?

I don't know how many users will be doing this manually. Could they not be using a tool to extract what they want, e.g. the slsa-verifier? 
I am only looking at it from a UX standpoint, and from the point of view a majority of users who generate attestations and don't (want to) know the internal implementation details. The naming convention has already become part of semantic versioning. Users will need to change their behavior, downstream tooling will have to change their behavior (there are folks who download attestation upon release to verify them automatically), and this trickles down. It would be really nice if something could just have a "signed SLSA attestation" extension, which allows downstream users to verify simply by listing `*.slsa` and passing that to a verification tool. Otherwise every time we come up with a new format we will break everyone.

I understand it has some limitation, in the sense that we lose the underlying format, which is why I was proposing calling them `.<format>.slsa` - if that's useful to end-users.

laurentsimon:
> That reasoning makes sense to me but I wouldn't anchor too much on just fetching the 'slsa' attestation. SLSA also has VSAs, and plans to add new tracks which may add new attestations (in addition to whatever in-toto already provides). 

re: VSA -> `.vsa`
re: multiple tracks: `.<track>-slsa`
re: intoto. It's not uncommon to rename formats when they have a common underlying format. SARIF is json based, and folks give it the `.sarif` extension. `/etc/mime.types` has plenty of other examples.

> I think as long as there's a way to transmit multiple attestations where the downstream user can fetch all of them that should be ok. But if we wind up with something that just works for a single predicate or where the user has to know the exact predicate type to fetch I don't think it will work out well.

Users who generate multiple attestations will need to have them stored in a single place. if they have a VSA, a code review attestation, a build attestation, is their only choice to merge them together into a single "bundle"? These attestation will be generated by different entities, so we can't ask users to be aware of builder internals, IMO. I don't see a big problem having `<mypackage>.build-slsa`, `<mypackage>.review-slsa`, `<mypackage>.vsa-slsa`. I actually think it's self explanatory for users. 

Probably better to move this discussion to the main slsa repo?

laurentsimon:
Created https://github.com/slsa-framework/slsa-github-generator/issues/2230 for tracking on our side.

laurentsimon:
We decided to use `build.slsa`, sea reasoning in https://github.com/slsa-framework/slsa-github-generator/issues/2230

ianlewis (approved):

PR #2229, opened 2023-06-06
fix: fix verify-token SLSA v1.0 source resourcedescriptor

asraa:
The predicate output uses a `gitCommit` instead of `sha1`


.github/actions/verify-token/dist/index.js
```diff
@@ -552,7 +552,7 @@ function createPredicate(rawTokenObj, toolURI, token, isGenerator) {
                     {
                         uri: sourceURI,
                         digest: {
-                            gitCommit: sourceSha1,
```
ianlewis, line 555: I think technically `gitCommit` is ok. Though it kind of depends on what we want to do.
https://github.com/in-toto/attestation/blob/ac23735e1571fb2f442086d7e8e152d65c4dc0fd/spec/v1/digest_set.md#gitcommit-gittree-gitblob-gittag

Was there an issue that prompted you to make the change?

asraa: Oh, interesting, I was going off ResourceDescriptor, and for some reason that does not use DigestSet:

https://github.com/in-toto/attestation/blob/ac23735e1571fb2f442086d7e8e152d65c4dc0fd/spec/v1/resource_descriptor.md#resourcedescriptor-field-type-specification

The verify-token scripts are expecting `sha1` - so I can always update them.

PR #2241, opened 2023-06-07
docs: Hardening BYOB

laurentsimon:

PR #2267, opened 2023-06-09
feat: Add BYOB Bazel Builder

enteraga6:
@mihaimaruseac 

Adds a BYOB Bazel Builder.

Support for artifacts with runtime files and e2e tests coming soon.


enteraga6:
btw: did not mean to try to auto-merge, thought it wanted to update my branch with recent commits from main!

ianlewis:
> btw: did not mean to try to auto-merge, thought it wanted to update my branch with recent commits from main!

no worries. Just disable it and all is good 😄

.github/workflows/builder_bazel_slsa3.yml
```diff
@@ -0,0 +1,69 @@
...
+        description: "Artifacts to build"
+        required: true
+        type: string
+      flags:
+        description: "Flags for the bazel build command"
+        required: false
+        type: string
+        default: ""
+
+    outputs:
```
ianlewis, line 35: Are there any other artifacts that we can glean from the bazel config and provide here in outputs?

I imagine this is pretty hard but maybe the `bazel` command is able to tell us what it's outputs are somehow?

.github/workflows/builder_bazel_slsa3.yml
```diff
@@ -0,0 +1,69 @@
...
+      attestations-download-name:
+        description: >
+          Name of the artifact to download all the attestations.
+
+          When run on a `pull_request` trigger, attestations are not signed and have an ".intoto" extension.
+          When run on other triggers, attestations are signed and have an "intoto.sigstore" extension.
+        value: ${{ jobs.slsa-run.outputs.attestations-download-name }}
+jobs:
```
ianlewis, line 43: nit: blank line between sections.
```suggestion
        value: ${{ jobs.slsa-run.outputs.attestations-download-name }}

jobs:
```

.github/workflows/builder_bazel_slsa3.yml
```diff
@@ -0,0 +1,69 @@
...
+on:
+  workflow_call:
+    inputs:
+      rekor-log-public:
+        description: "Allow publication of your repository name on the public Rekor log"
+        required: false
+        type: boolean
+        default: false
+      targets:
+        description: "Artifacts to build"
```
ianlewis, line 26: Can multiple targets be specified? Maybe include in the description how to do that? (e.g. "comma separated", "semicolon separated" or whatever)

internal/builders/bazel/README.md
```diff
@@ -0,0 +1,252 @@
...
+Bazel build artifacts produced using the identity of the GitHub workflow. This can be used to
+create a positive attestation to an artifact coming from your repository.
+
+That means that once your users verify the artifact they have downloaded they can
+be sure that it was created by your repository's workflow and hasn't been
+tampered with.
+
+## Development status
+
+The Bazel builder is currently in beta. The API could change while approaching
```
ianlewis, line 45: Is this in `beta`? or are we planning to release it very soon?

If not we may want to just say it's under development.

internal/builders/bazel/README.md
```diff
@@ -0,0 +1,252 @@
...
+| rekor-log-public  | No       | false              | Set to true to opt-in to posting to the public transparency log. Will generate an error if false for private repositories. This input has no effect for public repositories. See [Private Repositories](#private-repositories).<br>Default: `false` |
+| targets           | Yes      |                    | A space separated list of targets to build and generate artifacts for. See [targets](https://bazel.build/concepts/build-ref#targets) for more information. \                                                                                        |
+| flags             | No       | ""                 | A space separated list of flags to modify the build by. See [flags](https://bazel.build/docs/user-manual#build-options) for more information. \                                                                                                     |
```
ianlewis, line 160: I wonder if we have done any kind of exploration as to what allowing arbitrary flags to bazel would allow a user to do?

For example, we can change the `--package_path` which might open it up to some strange exploits. I suppose this is mostly fine as long as we record the `flags` input.

.github/workflows/builder_bazel_slsa3.yml
```diff
@@ -0,0 +1,69 @@
...
+        required: true
+        type: string
+      flags:
+        description: "Flags for the bazel build command"
+        required: false
+        type: string
+        default: ""
+
+    outputs:
+      attestations-download-name:
```
ianlewis, line 36: nit: use "provenance" to be consistent with other builders.
```suggestion
      provenance-download-name:
```

internal/builders/bazel/build.sh
```diff
@@ -0,0 +1,34 @@
...
+# Build with respect to entire arrays of flags and targets
+bazel build "${BUILD_FLAGS[@]}" "${BUILD_TARGETS[@]}"
+
+# Using target string, copy artifact to binaries dir
+for CURR_TARGET in "${BUILD_TARGETS[@]}"; do
+  CD_PATH=${CURR_TARGET%:*}
+  CD_PATH=${CD_PATH////}
+  BINARY_NAME=${CURR_TARGET#*:}
+  cp "bazel-bin/$CD_PATH/$BINARY_NAME" ./binaries
+done
```
ianlewis, line 34: Does this allow us do use other rules packages to build artifacts (e.g. [rules_python](https://github.com/bazelbuild/rules_python))? Do output artifacts always get picked up this way?

I suspect that most popular rules are sufficiently hermetic and just generate files without any side-effects but I'm not 100% sure.

ianlewis: If not, it would be nice to update the docs to include which types of rules we support and which we don't.

internal/builders/bazel/build.sh
```diff
@@ -0,0 +1,34 @@
...
+IFS=' ' read -r -a BUILD_TARGETS <<< "${TARGETS}"
+
+# Build with respect to entire arrays of flags and targets
+bazel build "${BUILD_FLAGS[@]}" "${BUILD_TARGETS[@]}"
+
+# Using target string, copy artifact to binaries dir
+for CURR_TARGET in "${BUILD_TARGETS[@]}"; do
+  CD_PATH=${CURR_TARGET%:*}
+  CD_PATH=${CD_PATH////}
+  BINARY_NAME=${CURR_TARGET#*:}
```
ianlewis, line 32: I know we're not always good about it but bash is often difficult to understand/remember. Consider adding a comment as to what each of these lines are doing.

internal/builders/bazel/generate-layout.sh
```diff
@@ -0,0 +1,53 @@
...
+# "version" and "attestations" fields:
+echo -e -n "{\n  \"version\": 1,\n  \"attestations\": [" >> "$SLSA_OUTPUTS_ARTIFACTS_FILE"
+
+num_binary_files=$(find ./binaries -type f | wc -l)
+counter=1
+
+# Add one attestation per binary:
+find ./binaries -type f -print0 | while read -r -d $'\0' fname
+do
+
```
ianlewis, line 28: nit: delete blank line
```suggestion
```

internal/builders/bazel/build.sh
```diff
@@ -0,0 +1,34 @@
...
+IFS=' ' read -r -a BUILD_FLAGS <<< "${FLAGS}"
+IFS=' ' read -r -a BUILD_TARGETS <<< "${TARGETS}"
+
+# Build with respect to entire arrays of flags and targets
+bazel build "${BUILD_FLAGS[@]}" "${BUILD_TARGETS[@]}"
+
+# Using target string, copy artifact to binaries dir
+for CURR_TARGET in "${BUILD_TARGETS[@]}"; do
+  CD_PATH=${CURR_TARGET%:*}
+  CD_PATH=${CD_PATH////}
```
ianlewis, line 31: Is this right? I think this removes all '/' characters from the target. IIUC, This would be ok for targets like `//foo:bar` but won't work for targets like `//foo/bar:baz`. In that case `CD_PATH` would end up being `foobar:baz` rather than `foo/bar:baz`.

You also might want to consider relative targets that aren't referencing the workspace root. i.e. `foo/bar:baz` How would we handle getting the right path for those targets?

internal/builders/bazel/generate-layout.sh
```diff
@@ -0,0 +1,53 @@
...
+      echo -n "," >> "$SLSA_OUTPUTS_ARTIFACTS_FILE"
+    fi
+
+    counter="$((counter +1))"
+done
+
+# Close "attestations" and "version":
+echo -e "\n  ]" >> "$SLSA_OUTPUTS_ARTIFACTS_FILE"
+echo "}" >> "$SLSA_OUTPUTS_ARTIFACTS_FILE"
+cat "$SLSA_OUTPUTS_ARTIFACTS_FILE"
```
ianlewis, line 53: nit: newline at end of file
```suggestion
cat "$SLSA_OUTPUTS_ARTIFACTS_FILE"

```

ianlewis (commented):
Thanks for doing this! Looks pretty good!

internal/builders/bazel/action.yml
```diff
@@ -0,0 +1,91 @@
...
+
+    # rng generates a random number to avoid name collision in artifacts
+    # when multiple workflows run concurrently.
+    - name: Generate random 16-byte value (32-char hex encoded)
+      id: rng
+      uses: slsa-framework/slsa-github-generator/.github/actions/rng@main
+
+    - name: Generate Artifacts
+      id: generate-artifacts
+      uses: actions/upload-artifact@0b7f8abb1508181956e8e162db84b466c27e18ce # v3.1.2
```
laurentsimon, line 69: you'll eventually need to use the public `actions/secure-upload-artifacts` from https://github.com/slsa-framework/slsa-github-generator/pull/2241/files.
Please create an issue on our repo and add a comment in the code here `// TODO(#issue-number): Use secure upload folder Action`

.github/workflows/builder_bazel_slsa3.yml
enteraga6: What do you mean by any other artifacts? Some artifacts will need runfiles along with it in order to run, and I am working on implementing that as well. Do you mean something else though?

internal/builders/bazel/README.md
enteraga6: Right! Will change to say it's under development, thank you for the catch!

internal/builders/bazel/README.md
enteraga6: I have not done any exploration on that front, but will mark it down for further review!

internal/builders/bazel/build.sh
enteraga6: No, I am sorry this is bugged. Caught it as well on a more complex directory test. Will fix it and look into relative targets!

internal/builders/bazel/build.sh
enteraga6: Definitely, I hear you and agree! Will add comments!

.github/workflows/builder_bazel_slsa3.yml
ianlewis: I meant "could we provide the created artifacts in an output somehow?" I think it would be useful to output artifact name(s) for the binaries that were created.

internal/builders/bazel/build.sh
enteraga6: I am unsure about other rule packages, and I will look into it further and update the readme with my findings!

.github/workflows/builder_bazel_slsa3.yml
enteraga6: Right. The binaries are uploaded as a folder of artifacts, but to support runfiles, I am making each artifact be its own folder with the binary it represents as well as the runfiles it needs to run. Perhaps some beneficial outputs would be the names of the artifact folders and the name of the binary zip.

internal/builders/bazel/action.yml
enteraga6: Created it and added the comment!

.github/workflows/builder_bazel_slsa3.yml
laurentsimon: @mihaimaruseac

internal/builders/bazel/build.sh
mihaimaruseac: I think it should support the other rules, but will probably have to test

internal/builders/bazel/build.sh
enteraga6: Fixed parsing bug, will test it further though and look into relative targets!

internal/builders/bazel/build.sh
enteraga6: Comments added on new bash script.

.github/workflows/builder_bazel_slsa3.yml
enteraga6: Updated targets description.

.github/workflows/builder_bazel_slsa3.yml
enteraga6: Just to add: the runfile support will come later and I was hypothesizing future constructive outputs. Right now, the current structure is that the artifact that is uploaded is a folder called, "binaries", which is filled with the artifacts constructed from the build.

internal/builders/bazel/build.sh
enteraga6: Bazel requires all targets to referenced with respect to repo: 

> Relative labels cannot be used to refer to targets in other packages; the repository identifier and package name must always be specified in this case.

so we do not need to worry getting the right path for those cases

ianlewis: So you can definitely do things like `bazel build target:foo` which will be relative to the current directory (even though you are running it in the workspace root).

For example:
```
git clone https://github.com/bazelbuild/examples
cd examples/cpp-tutorial/stage2
bazel build main:hello-world
```

In this case we are calling the `hello-world` target in the `main` directory. So in this case I would expect it to work even though the target we are referring to doesn't start with `//`.

internal/builders/bazel/build.sh
enteraga6: I tested the python rule set and it will be supported when the runfile feature gets added. Runfile feature is being developed currently and is coming soon.

internal/builders/bazel/README.md
enteraga6: > For example, we can change the `--package_path` which might open it up to some strange exploits. I suppose this is mostly fine as long as we are recording the `flags` input.

What exploits were you thinking that --package_path could produce? I have been looking over all the flag options [here](https://bazel.build/reference/command-line-reference#options-common-to-all-commands), and I think the flags that could be exploited would just hurt the user that calls the builder in their workflow.

The user could alter different flags to the builder such that their build is delayed, but wouldn't that just hurt them?

The flags input is recorded in the provenance generated.

.github/workflows/builder_bazel_slsa3.yml
enteraga6: Additional outputs will be in future runfile PR, which will include secure-upload action for binaries folder that has outputs to send back to user

internal/builders/bazel/build.sh
enteraga6: I agree with the example you provided. Since users will input their targets in their workflow job which references the Bazel Builder, I will put into the documentation to reference everything from the root, using // syntax. In the builder it clones the repo and the context is from the root, so all the targets will be relative to the root of the repo. If a target was passed into the workflow in form of `main:hello-world`, it would require additional complexity on the builder side to find the relative target or additional constraints on target name uniqueness.

Do you think that there is a big push for being able to reference targets relatively? I think having to reference the targets from the root is fine, and will document that constraint in the readme.

internal/builders/bazel/build.sh
ianlewis: Cool! I look forward to seeing some docs and examples in the future!

internal/builders/bazel/README.md
ianlewis: Yeah, I'm not sure that this is an issue we need to resolve on this PR. Just musing really.

BTW is there a way to specify the bazel version you use? Some flags might be supported on this or that version so it might be good if the user could specify that. Maybe something to consider on a later PR.

internal/builders/bazel/build.sh
ianlewis: Yeah, you can document that constraint. I think it's pretty common though for folks to run things like `bazel build :main` or `bazel build main:main` so you would want to make that really clear in the documentation that it isn't supported.

internal/builders/bazel/README.md
```diff
@@ -0,0 +1,253 @@
...
+    if: startsWith(github.ref, 'refs/tags/')
+    uses: slsa-framework/slsa-github-generator/.github/workflows/builder_bazel_slsa3.yml@v1.7.0
+    with:
+      targets: "//src:fib //src:hello"
+      flags: "--strip=always"
+```
+
+The `targets` are a set of space separated build targets to be built. The targets are assume to
+be referenced from the root of the directory. Because of this the each target inputted should be referenced
+as `//path/from/root/to/target:your_target`.
```
ianlewis, line 96: We should also mention that the target identifier needs to be specified. Bazel allows you to specify just a package name and defaults to a target id with the same name as the package name (e.g. `bazel build //test` will build `//test:test`)
```suggestion
The `targets` are a set of space separated build targets to be built. Each target must include the `//` workspace root identifier and package target identifier (`:your_target`). Because of this each target should be of the form `//path/from/root/to/target:your_target`.
```

internal/builders/bazel/README.md
```diff
@@ -0,0 +1,253 @@
...
+- `push` (including new tags)
+- `release`
+- Manual run via `workflow_dispatch`
+
+However, in practice, most triggers should work with the exception of
+`pull_request`. If you would like support for `pull_request`, please tell us
+about your use case on
+[issue #358](https://github.com/slsa-framework/slsa-github-generator/issues/358). If
+you have an issue in all other triggers please submit a
+[new issue](https://github.com/slsa-framework/slsa-github-generator/issues/new/choose).
```
ianlewis, line 149: These are the event types BYOB supports based on https://github.com/slsa-framework/github-actions-buildtypes/tree/main/workflow/v1

```suggestion
Only the following [event types] are supported:

| Supported event type  | Event description                          |
| --------------------- | ------------------------------------------ |
| [`create`]            | Creation of a git tag or branch.           |
| [`release`]           | Creation or update of a GitHub release.    |
| [`push`]              | Creation or update of a git tag or branch. |
| [`workflow_dispatch`] | Manual trigger of a workflow.              |

`pull_request` events are currently not supported. If you would like support for
`pull_request`, please tell us about your use case on
[issue #358](https://github.com/slsa-framework/slsa-github-generator/issues/358). If
you have an issue in all other triggers please submit a
[new issue](https://github.com/slsa-framework/slsa-github-generator/issues/new/choose).

[event types]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows
[`create`]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#create
[`release`]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release
[`push`]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push
[`workflow_dispatch`]: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch
```

internal/builders/bazel/build.sh
```diff
@@ -0,0 +1,44 @@
...
+IFS=' ' read -r -a BUILD_TARGETS <<< "${TARGETS}"
+
+# Build with respect to entire arrays of flags and targets
+bazel build "${BUILD_FLAGS[@]}" "${BUILD_TARGETS[@]}"
+
+# Using target string, copy artifact to binaries dir
+for CURR_TARGET in "${BUILD_TARGETS[@]}"; do
+  # Take out the first two // in CURR_TARGET
+  # "//src/internal:fib" --> "src/internal:fib"
+  CD_PATH=$(echo "$CURR_TARGET" | cut -d'/' -f3-)
```
ianlewis, line 32: If we require the targets be referenced from the workspace root this might fail (or even succeed) in some unexpected ways if users don't do that. For example, if there is a `a/b/c:main` target and a `abc:main` target this would turn `a/b/c:main` into `abc:main` and upload a different target than was intended.

Maybe we should validate it and throw an error if the target doesn't begin with `//`? Maybe also validate that the target id is specified? e.g. check for the `:`?

internal/builders/bazel/build.sh
```diff
@@ -0,0 +1,44 @@
...
+
+  # Removes field after and including the colon
+  # "src/internal:fib" --> "src/internal"
+  CD_PATH=$(echo "$CD_PATH" | cut -d':' -f1)
+
+  # Removes everything up to and including the first colon
+  # "//src/internal:fib" --> "fib"
+  BINARY_NAME=${CURR_TARGET#*:}
+
+  #Copy the binary to artifact directory, binaries
```
ianlewis, line 42: nit: extra space after the `#`
```suggestion
  # Copy the binary to artifact directory, binaries
```

ianlewis (approved):
Feel free to create issues for the comments and address on follow up PRs

internal/builders/bazel/build.sh
ianlewis: Created https://github.com/slsa-framework/slsa-github-generator/issues/2308

ianlewis:
> Feel free to create issues for the comments and address on follow up PRs

Actually I'll just go ahead create the issues myself and merge. Great job on this PR!

ianlewis (approved):

ianlewis:
I guess since I made commits it still requires an approval... @laurentsimon

ianlewis:
> I guess since I made commits it still requires an approval... @laurentsimon

Actually no, never mind. It seems it's waiting for one more job that is taking forever.

internal/builders/bazel/build.sh
enteraga6: I agree, validation should be added in next PR.

`a/b/c:main` --> `c:main` after `CD_PATH=$(echo "$CURR_TARGET" | cut -d'/' -f3-)` since it cuts off the first two `/`'s and keeps everything after the second `/`.

The target beginning with `//` is crucial if the path to the target is longer as we can see above. Validation for target id can be added with it to check for `:` and string after it.

Thanks for this idea!

internal/builders/bazel/README.md
enteraga6: Thank you for the clarification!

PR #2324, opened 2023-06-21
fix: resolve target edge cases in Bazel Builder with cquery

enteraga6:
cc: @mihaimaruseac 

Changed getting the targets from parsing in shell script to using Bazel's built in `cquery` command.

This resolves edge cases that resulted from the parsing schema. Particularly, more generic glob patterns are supported such as `//src/...` and `//src/internal:all`. Additionally, using cquery to obtain target paths allows the user to have more flexibility in the flags that they can use to configure the build. Users that pass the flag to optimize the build process or users on different types of machines will have a different path to the artifacts within the bazel-out directory. Using cquery, gives the path to artifacts solving this problem.

However, cquery does not have runfile support as it is based on the build graph. Runfile support for artifacts is currently being developed.


internal/builders/bazel/README.md
```diff
@@ -93,6 +93,9 @@ jobs:
 
 The `targets` are a set of space separated build targets to be built. Each target must include the `//` workspace root identifier and package target identifier (`:your_target`). Because of this each target should be of the form `//path/from/root/to/target:your_target`.
 
+Targets can also be referred to with general glob patterns such as `//src/...` or `//src/internal:all`. Note however, that support for artifacts that
+require runfiles is still currently in development and not available at this time.
```
ianlewis, line 97: It might be nice to track this on an issue.

internal/builders/bazel/build.sh
```diff
@@ -25,20 +25,23 @@ IFS=' ' read -r -a BUILD_TARGETS <<< "${TARGETS}"
...
-for CURR_TARGET in "${BUILD_TARGETS[@]}"; do
-  # Take out the first two // in CURR_TARGET
-  # "//src/internal:fib" --> "src/internal:fib"
-  CD_PATH=$(echo "$CURR_TARGET" | cut -d'/' -f3-)
-
-  # Removes field after and including the colon
-  # "src/internal:fib" --> "src/internal"
-  CD_PATH=$(echo "$CD_PATH" | cut -d':' -f1)
+# Use associative array as a set to increase efficency in avoiding double copying the target
+declare -A FILES_SET
```
ianlewis, line 29: If you're using variables in this file only then I think we should use lowercase by convention. Declaration and usage of variables can then be tracked by `shellcheck`.

Uppercase variable names are assumed to be global, cross-script environment variables. e.g. the ones set as inputs to this script by the workflow (`TARGETS`, `FLAGS`)

This goes for other variables too, like `BILDE_TARGETS`, `BUILD_FLAGS`, `CURR_TARGET` etc.

Apologies for not commenting on this sooner.
```suggestion
declare -A files_set
```

internal/builders/bazel/build.sh
```diff
@@ -25,20 +25,23 @@ IFS=' ' read -r -a BUILD_TARGETS <<< "${TARGETS}"
...
+    FILES_SET["${file}"]="1"
+  done <<< "$bazel_generated"
+done
 
-  # Copy the binary to artifact directory, binaries
-  cp "bazel-bin/$CD_PATH/$BINARY_NAME" ./binaries
+# Copy set of unique targets to binaries. Without !, it would give values not keys
+for file in "${!FILES_SET[@]}"; do
+  # Remove the symbolic link and copy
+  cp -L "$file" ./binaries
```
ianlewis, line 46: Also, consider copying files to upload into a temp directory rather than `./binaries` since this is in the `GITHUB_WORKSPACE` which is typically a checkout of the user's repo so you technically could run into name collisions.

Could be on a later PR.

ianlewis (approved):

internal/builders/bazel/build.sh
enteraga6: Thanks for the idea! I will implement that later. Tracking it on this [issue](https://github.com/slsa-framework/slsa-github-generator/issues/2331).

internal/builders/bazel/README.md
enteraga6: I agree! Tracking it on this [issue](https://github.com/slsa-framework/slsa-github-generator/issues/2332)

internal/builders/bazel/build.sh
enteraga6: Thanks for letting me know! Just switched them to the lowercase convention :)

ianlewis (approved):

PR #2326, opened 2023-06-22
feat: sha1 option for checkout

laurentsimon:


.github/actions/secure-project-checkout/action.yaml
```diff
@@ -38,6 +43,7 @@ runs:
       uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
       with:
         fetch-depth: ${{ inputs.fetch-depth }}
+        ref: ${{ inputs.checkout-sha1 }}
```
ianlewis, line 46: If someone sets `checkout-sha1` to something like `main` then I guess this would succeed but would fail on the sha check below?

I wonder if we shouldn't validate it's a sha before doing anything.

ianlewis (approved):

.github/actions/secure-project-checkout/action.yaml
laurentsimon: We validate it in verify-token https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/verify-token/src/utils.ts#L61.

But I've added a check in bash. Good call.

Issue #2273, opened 2023-06-12
[bug][byob] Remove `.intoto` from attestation name 


laurentsimon:
Builders built with BYOB should not set `.intoto` in their name.
We need to update the gradle, maven, nodejs and bazel builders

See comment https://github.com/slsa-framework/slsa-github-generator/pull/2266#issuecomment-1586403799

PR #2330, opened 2023-06-22
feat: remove intoto extension for BYOB

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/2273


internal/builders/gradle/create_attestation.sh
```diff
@@ -31,7 +31,7 @@ do
           
     echo -n "
```
laurentsimon, line 32: note: the maven builder calls this file as well

ianlewis (approved):

PR #2336, opened 2023-06-23
feat: Include error cause for Sigstore Internal Errors

enteraga6:
cc: @mihaimaruseac 

This expands the information the user will receive when there is a Sigstore Internal Error during the jobs of a BYOB workflow or other workflows that utilize the `sign-attestations` Github Action. Before, there was just the assigned error message, but after this, there will be the error message and cause, if it is an `InternalError`.

Updated sigstore version from `"^1.1.1"` --> `"1.5.1"` to have access to the class export of `InternalError`.


laurentsimon (approved):
LGTM, thanks!

Issue #2332, opened 2023-06-22
[feature][bazel] Add support for artifacts that need runfiles


enteraga6:
### Issue
Currently, the Bazel Builder does not have support for artifacts that require runfiles in order to execute. The bazel builder produces artifacts, but if runfiles are required they are not provided with the artifact itself, making it impossible for it to run.

### Proposal
Add support for runfiles by letting the user pass a flag into the Bazel Builder for if they want their artifacts to include any runfiles that Bazel generates for it. This will allow users to pick, based on if they need the runfiles or not.

PR #2362, opened 2023-07-07
feat: Runfile and java artifact support for Bazel Builder

enteraga6:
cc: @mihaimaruseac 

This will provide support for artifacts that also need runfiles. Users can use a flag, `needs-runfiles`, to also package the artifact alongside the runfiles Bazel generates if the user deems it as necessary. 

Also, support for Java artifacts is made available by packaging the JARs to be standalone through a format in Bazel called `_deploy.jar`. Alongside the `_deploy.jar`, there will be a modified `run-script` that allows the users that download the artifact to run the run-script by using a flag called `local_javabin`, where they put the path to their own java bin such that it is utilized by the run-script. This run-script is generated by Bazel from a template and later modified in the `build.sh` in the internal part of the builder to add this flag for the users. More information is available on the readme.

Java targets will automatically be converted to their `_deploy.jar` with this. Three flags are used for users that have java targets:

`includes-java`: if true then adds an additional flag to build command as well as rule for local java repo in WORKSPACE in order to utilize the `--singlejar` capability of run-script for `_deploy.jar` such that the remote jdk does not need to be included in runfiles. Doing it like this prevents massive bloat when attesting.

`user-java-distribution` and `user-java-version`: let the user specify the exact java they want to use to build

When users run the run-script they will include an additional flag `local_javabin` which they will set to their local javabin that the run-script will utilize to run the `_deploy.jar`

A combination of `bazel query` and `bazel cquery` were used to resolved edge cases with the implementation.


laurentsimon (commented):
There seems to be changes to files that are out of scope of the PR. Is it a git / merge problem?

enteraga6:
Yes, everything is synced up with the main branch now after resolving the conflicts. Sorry about that.

enteraga6:
closes #2332

internal/builders/bazel/action.yml
```diff
@@ -51,10 +51,24 @@ runs:
...
+      id: java
+      uses: actions/setup-java@5ffc13f4174014e2d4d4572b3d74c3fa61aeb2c2 # v3.11.0
+      with:
+        distribution: "${{ fromJson(inputs.slsa-workflow-inputs).user-java-distribution }}"
+        java-version: "${{ fromJson(inputs.slsa-workflow-inputs).user-java-version }}"
+
+    - name: Set Env Var $JAVA_HOME
+      shell: bash
+      run: |
+        export JAVA_HOME="$JAVA_HOME"
```
ianlewis, line 64: What does this do? Isn't `JAVA_HOME` set by `setup-java`?

ianlewis (approved):

internal/builders/bazel/action.yml
enteraga6: This was supposed to export JAVA_HOME manually so that it could be used to make the local Java repository rule within the WORKSPACE file. I did it to solve this issue:
https://github.com/actions/runner/blob/master/docs/adrs/0278-env-context.md#dont-populate-the-env-context-with-environment-variables-from-runner-machine

However, I just tested it, and it is not the case anymore that it must be manually exported, so this step is actually un-needed. Thanks!

Issue #845, opened 2022-09-11
[bug] Alternative ways required to pass large attestation subjects


behnazh-w:
**Describe the bug**
For complex projects that include multiple modules or packages, the number of subjects and size of `base64-subjects` can easily grow larger than [the allowed size](https://stackoverflow.com/questions/1078031/what-is-the-maximum-size-of-a-linux-environment-variable-value) of command-line argument and environment strings. For example the size of the subject hashes in a project I tried recently is about 200KB which results in the following error when the generator tries to pass it [as command-line argument](https://github.com/slsa-framework/slsa-github-generator/blob/v1.2.0/.github/workflows/generator_generic_slsa3.yml#L120).
```
 Error: An error occurred trying to start process '/usr/bin/bash' with working directory '/home/runner/work/x'. Argument list too long
 ```

To get around that, possible [solutions](https://stackoverflow.com/questions/33051108/how-to-get-around-the-linux-too-many-arguments-limit) require making changes to the GH runner, which is not a good idea probably.

Would it be possible to pass the subjects in a different way and not as an environment string through command-line argument as done [here](https://github.com/slsa-framework/slsa-github-generator/blob/v1.2.0/.github/workflows/generator_generic_slsa3.yml#L120)?

**Additional context**

- The documentation of kernel call [execve()](https://man7.org/linux/man-pages/man2/execve.2.html) in section `Limits on size of arguments and environment` is helpful.
- Here is another [related question](https://stackoverflow.com/questions/73657514/setting-env-variable-to-a-large-value-succeeds-then-fails) on Stack Vverflow.

ianlewis:
Yes, I think the right way to do this would need to be via a config file that is committed to the repo. The workflow could read this file from the repository to get the subjects needed. I'm open to other ideas though.

Just out of curiosity, what was it that made the base64 of the subjects so large? was it the sheer number of subjects? or were the subject names very long?

behnazh-w:
> Yes, I think the right way to do this would need to be via a config file that is committed to the repo. The workflow could read this file from the repository to get the subjects needed. I'm open to other ideas though.

[...]

PR #2365, opened 2023-07-10
feat: Support large subjects file in generic generator

laurentsimon:
closes https://github.com/slsa-framework/slsa-github-generator/issues/845


.github/workflows/generator_generic_slsa3.yml
```diff
@@ -281,3 +330,12 @@ jobs:
           set -euo pipefail
           echo "outcome=$([ "$SUCCESS" == "true" ] && echo "success" || echo "failure")" >> "$GITHUB_OUTPUT"
           [ "$CONTINUE" == "true" ] || [ "$SUCCESS" == "true" ] || exit 27
+
+  # cleanup deletes internal artifacts used by the generator workflow
+  cleanup:
+    runs-on: ubuntu-latest
+    needs: [detect-env, generator, upload-assets]
+    steps:
+      - uses: geekyeggo/delete-artifact@54ab544f12cdb7b71613a16a2b5a37a9ade990af # v2.0.0
```
ianlewis, line 339: Is it really necessary to delete the artifact? I'm worried a bit about introducing another 3rd party dependency.

internal/builders/generic/attest.go
```diff
@@ -51,9 +52,14 @@ run in the context of a Github Actions workflow.`,
 			ghContext, err := github.GetWorkflowContext()
 			check(err)
 
+			// If not sujects are provided, read from the file instead.
```
ianlewis, line 55:
```suggestion
			// If not subjects are provided, read from the file instead.
```

internal/builders/generic/attest.go
```diff
@@ -132,10 +138,13 @@ run in the context of a Github Actions workflow.`,
 		&attPath, "signature", "g", "",
 		"Path to write the signed provenance.",
 	)
+	c.Flags().StringVarP(
+		&subjectsFilename, "subjects-filename", "f", "",
+		"Filename containing a formatted list of subjects in the same format as sha256sum (base64 encoded).",
+	)
 	c.Flags().StringVarP(
 		&subjects, "subjects", "s", "",
```
ianlewis, line 146: Do we even use the subjects flag anymore? Can we remove it?

.github/workflows/generator_generic_slsa3.yml
laurentsimon: +1, created https://github.com/slsa-framework/slsa-github-generator/issues/2382 to delete without the Action.
Note that this Action is already used in delegators. It's has 2 releases since 2020 so it won't incur additional dependabot PRs. But folks who use the Action allow-list settings won't like this.

internal/builders/generic/attest.go
laurentsimon: +1

laurentsimon:
ptal, I addressed the comments

ianlewis (approved):

PR #2380, opened 2023-07-13
feat: add maven verification plugin

AdamKorcz:
Adds a Maven plugin that makes it easy to verify the provenance files of all dependencies of a Maven-based project.

The plugin can either run during the Maven build cycle, or users can invoke it manually from within a project directory.

The plugin checks whether the dependencies in a `pom.xml` file are released with provenance attestations, and invokes the [slsa-verifier](https://github.com/slsa-framework/slsa-verifier) against the dependencies that have provenance files.

@loosebazooka @laurentsimon


internal/builders/maven/plugins/verification-plugin/src/main/java/io/github/adamkorcz/SlsaVerificationMojo.java
```diff
@@ -0,0 +1,189 @@
...
+            executeMojo(
+                plugin(
+                    groupId("org.codehaus.mojo"),
+                    artifactId("exec-maven-plugin"),
+                    version("3.1.0")
+                ),
+                goal("exec"),
+                configuration(
+                    element(name("executable"), "go"),
+                    element(name("commandlineArgs"), "install github.com/slsa-framework/slsa-verifier/v2/cli/slsa-verifier@latest")
```
laurentsimon, line 92: latest means latest version, not latest commit, correct?
An alternative here is to download the binary from GitHub and verify its provenance, instead of recompiling.
However, if users run the script multiple times, the go command will re-use the existing installation instead of recompiling.

Which option is best in your opinion?

Also, shall we be pinning to an immutable version?

AdamKorcz: I believe it pulls the latest commit, but I am having second thoughts about getting the Maven plugin to install the slsa-verifier. I think it is better to require the user the install it manually.

laurentsimon: re: latest commit. If e install anything, let's install at a version / tag.
Seems like a good to let users install themselves for this PoC.

internal/builders/maven/plugins/verification-plugin/src/main/java/io/github/adamkorcz/SlsaVerificationMojo.java
```diff
@@ -0,0 +1,189 @@
...
+                getLog().info("Skipping slsa verification for " + artifactStr + ": No slsa file found.");
+                continue;
+            }
+
+            // Verify slsa file
+            try {
+                // Run slsa verification on the artifact and print the result
+                // It will never fail the build process
+                String arguments = "verify-artifact --provenance-path ";
+                arguments += "${project.build.directory}/slsa/" + artifact.getArtifactId() + "-" + artifact.getVersion() + "-jar.intoto.build.slsa ";
```
laurentsimon, line 162: nit: remote intoto from the name. The builder at v1.70 does have this extension but it's been removed. v1.80 won't have it.

internal/builders/maven/plugins/verification-plugin/src/main/java/io/github/adamkorcz/SlsaVerificationMojo.java
```diff
@@ -0,0 +1,189 @@
...
+            } catch(MojoExecutionException e) {
+                getLog().info("Skipping slsa verification for " + artifactStr + ": No slsa file found.");
+                continue;
+            }
+
+            // Verify slsa file
+            try {
+                // Run slsa verification on the artifact and print the result
+                // It will never fail the build process
+                String arguments = "verify-artifact --provenance-path ";
```
laurentsimon, line 161: Can we be careful about script injection? Is there an exec (non-shell) command we can use instead?

ianlewis: Looks like the arguments are split by the exec-maven-plugin using `CommandLineUtils.tranlateCommandLine` which is implemented here: https://github.com/sonatype/plexus-utils/blob/5ba6cfcca911200b5b9d2b313bb939e6d7cbbac6/src/main/java/org/codehaus/plexus/util/cli/CommandLineUtils.java#L302-L383

It doesn't look obviously vulnerable to me but it does make me wary that the command is taken as a String rather than a String array and then split this way.

internal/builders/maven/plugins/verification-plugin/src/main/java/io/github/adamkorcz/SlsaVerificationMojo.java
```diff
@@ -0,0 +1,189 @@
...
+                arguments += " --source-uri ./ ${project.build.directory}/slsa/" + artifact.getArtifactId() + "-" + artifact.getVersion() + ".jar";
+                executeMojo(
+                    plugin(
+                        groupId("org.codehaus.mojo"),
+                        artifactId("exec-maven-plugin"),
+                        version("3.1.0")
+                    ),
+                    goal("exec"),
+                    configuration(
+                        element(name("executable"), gohome + "/bin/slsa-verifier"),
```
laurentsimon, line 172: if we let the user installs, is there a consistent way to get the path to the slsa-verifier?

internal/builders/maven/plugins/verification-plugin/src/main/java/io/github/adamkorcz/SlsaVerificationMojo.java
AdamKorcz: The plugin now requires the slsa-verifier to be pre-installed.

internal/builders/maven/plugins/verification-plugin/src/main/java/io/github/adamkorcz/SlsaVerificationMojo.java
AdamKorcz: Would prefer to catch up on this in a separate PR to be sure that we properly guard against command injections here. I have added an inline comment with a TODO.

internal/builders/maven/plugins/verification-plugin/pom.xml
```diff
@@ -0,0 +1,47 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
```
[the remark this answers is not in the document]
AdamKorcz: Yes

internal/builders/maven/plugins/verification-plugin/src/main/java/io/github/adamkorcz/SlsaVerificationMojo.java
laurentsimon: Thanks!

internal/builders/maven/plugins/verification-plugin/src/main/java/io/github/adamkorcz/SlsaVerificationMojo.java
AdamKorcz: The user now passes a parameter when using the plugin. This sets an environment variable that contains the path to the slsa-verifier.

internal/builders/maven/plugins/verification-plugin/src/main/java/io/github/adamkorcz/SlsaVerificationMojo.java
laurentsimon: sg, let's differ to the follow-up PR. Please create a tracking issue and update the comment to `TODO(#number)`

internal/builders/maven/plugins/verification-plugin/src/main/java/io/github/adamkorcz/SlsaVerificationMojo.java
laurentsimon: if the binari is in the `$PATH`, the user need no provide anything, correct? For xeample, if a user install using the GitHub action in a workflow, they won't know where it was installed. They just know it's in the path and they can run `$ slsa-verifier`. In Golang, you can use https://pkg.go.dev/os/exec#LookPath to get the full path to a binary. Is there something similar for Java?

AdamKorcz:
The failing tests look unrelated to this PR.

internal/builders/maven/plugins/verification-plugin/src/main/java/io/github/adamkorcz/SlsaVerificationMojo.java
laurentsimon: gotcha... let's discuss in follow-up PR then.

laurentsimon (approved):

laurentsimon:
the failing pre-submits are due to the RC release @ianlewis is going to cut. We can merge after it's cut

PR #2397, opened 2023-07-18
[chore][bazel] use secure-folder-upload and temp dir

### LINKED ISSUE #2276: [feature] Use Secure Upload folder action on BYOB Bazel Builder
Need to replace upload-artifact for directory in BYOB Bazel Builder with secure upload folder action.

Secure Upload folder action located here: https://github.com/slsa-framework/slsa-github-generator/pull/2241/files

### LINKED ISSUE #2331: [feature][bazel] Use a temp directory for artifacts
### Issue
The directory that the artifacts gets copied to is `./binaries` which is in the `GITHUB_WORKSPACE` that includes a checkout of the user's repo. Because of this, it is possible to run into name collisions on the copy.

### Proposal
Instead of copying to `./binaries`, copy to a temporary directory instead to avoid running into potential name collisions.

enteraga6:
closes #2276 
closes #2331 

This switches the folder upload from the actions/upload-artifact to our secure-upload-folder @ v1.7.0.

This also strengthens the name of the folder that is uploaded to avoid collisions. Since the path to folder cannot be in /tmp with the secure-upload-folder, the name of the folder being upload was changed from `binaries` to `bazel_builder_binaries_to_upload_to_gh` to avoid any potential name collisions and be able to utilize the secure-upload-folder action.

### DIFF
--- a/internal/builders/bazel/build.sh
+++ b/internal/builders/bazel/build.sh
@@ -17,7 +17,7 @@
 set -euo pipefail
 
 # TODO(Issue #2331): switch copy to binaries to a temp dir
-mkdir binaries
+mkdir bazel_builder_binaries_to_upload_to_gh
 
 # Transfer flags and targets to their respective arrays
 IFS=' ' read -r -a build_flags <<< "${FLAGS}"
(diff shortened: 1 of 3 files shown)


enteraga6:
cc/ @laurentsimon

internal/builders/bazel/build.sh
```diff
@@ -16,8 +16,7 @@
 
 set -euo pipefail
 
-# TODO(Issue #2331): switch copy to binaries to a temp dir
-mkdir binaries
+mkdir bazel_builder_binaries_to_upload_to_gh_7bc972367cb286b7f36ab4457f06e369
```
laurentsimon, line 19: nit: add a comment why there's a long string in the name, ie avoid collision.
You may as well make it a variable and use it in your code:
```shell
my_name="blabla"
...
"./${my_name}
```

laurentsimon (approved):

laurentsimon (approved):

PR #2458, opened 2023-07-24
feat: add documentation to gradle builder

AdamKorcz:
Adds documentation for the Gradle builder. 

Also removes the need for write-permissions for packages.

### DIFF
--- a/internal/builders/gradle/README.md
+++ b/internal/builders/gradle/README.md
@@ -1,0 +1,121 @@
+# Generation of SLSA3+ provenance for `.jar` artifacts built with Gradle
+
+This document explains how to generate SLSA provenance for artifacts built with Gradle.
+
+This can be done by adding a step to your Github Actions workflow to call a
+[reusable
+workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows)
+to build the package and generate SLSA provenance. We'll call this
+workflow the "Gradle builder" from now on.
+
+---
+
+<!-- markdown-toc --bullets="-" -i README.md -->
+
+<!-- toc -->
+
+- [Benefits of Provenance](#benefits-of-provenance)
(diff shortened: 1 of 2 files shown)


internal/builders/gradle/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+permissions: read-all
+
+jobs:
+  usetrw:
+    permissions:
+      contents: read
+      id-token: write
+      actions: read
+    uses: slsa-framework/slsa-github-generator/.github/workflows/builder_gradle_slsa3.yml@main
```
laurentsimon, line 83: Can you use `@v1.7.0`, same for Maven.
We pin at tag for other builders, see https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/generic.

I think we have a script to verify the tag is the latest. @ianlewis

laurentsimon (approved):

internal/builders/gradle/README.md
```diff
@@ -0,0 +1,121 @@
...
+
+jobs:
+  usetrw:
+    permissions:
+      id-token: write
+      contents: read
+      actions: read
+    uses: slsa-framework/slsa-github-generator/.github/workflows/builder_gradle_slsa3.yml@v1.7.0
+    with:
+      rekor-log-public: true
```
ianlewis, line 85: nit: I'd like to leave this off examples since folks are likely to copy and paste them, and we would really like them to understand what this option does before setting it (we already have docs in the Private Repositories section).

laurentsimon: Let me send a PR to remove these in all examples. Good catch

laurentsimon: https://github.com/slsa-framework/slsa-github-generator/pull/2466

PR #2459, opened 2023-07-24
feat: reduce package permissions in maven builder / publisher

AdamKorcz:
Write permissions are not needed for the lowperms delegator, so removing that from the Maven builder.

### DIFF
--- a/.github/workflows/publish_maven.yml
+++ b/.github/workflows/publish_maven.yml
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+permissions: read-all
+
 on:
   workflow_call:
     inputs:
--- a/internal/builders/maven/README.md
+++ b/internal/builders/maven/README.md
@@ -84,7 +84,6 @@ jobs:
       contents: write
       id-token: write
       actions: read
-      packages: write
     uses: slsa-framework/slsa-github-generator/.github/workflows/builder_maven_slsa3.yml@main
     with:
       rekor-log-public: true
--- a/internal/builders/maven/README.md
+++ b/internal/builders/maven/README.md
@@ -81,10 +81,10 @@ env:
 jobs:
   build:
     permissions:
-      contents: write
       id-token: write
+      contents: read
       actions: read
-    uses: slsa-framework/slsa-github-generator/.github/workflows/builder_maven_slsa3.yml@main
+    uses: slsa-framework/slsa-github-generator/.github/workflows/builder_maven_slsa3.yml@v1.7.0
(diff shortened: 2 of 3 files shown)


internal/builders/maven/README.md
```diff
@@ -84,7 +84,6 @@ jobs:
       contents: write
```
laurentsimon, line 84: is contents:write needed?

internal/builders/maven/README.md
```diff
@@ -81,11 +81,10 @@ env:
 jobs:
   build:
     permissions:
```
laurentsimon, line 83: the publisher call still uses write permissions:
```yaml
permissions:
      contents: write
      id-token: write
      actions: read
      packages: write
```, 

None of them are needed, correct?

**Review by ianlewis — COMMENTED** [review]
This LGTM but we should address @laurentsimon's comments.

File: internal/builders/maven/README.md (hunk 2 above)
Comment (ianlewis) [reply internal/builders/maven/README.md:83]: > None of them are needed, correct?

What do you mean here? IIUC we need these permissions after `packages: write` is removed from the call to the lowperms delegator.

```yaml
permissions:
  id-token: write
  contents: read
  actions: read
```

laurentsimon: sorry, that was meat as a comment for the publish, not the build. The publish should not need permissions

.github/workflows/publish_maven.yml
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+permissions: read-all
```
laurentsimon, line 15: added this

internal/builders/maven/README.md
```diff
@@ -97,12 +96,7 @@ You can also release artifacts to Maven Central by adding the following step to
 ```yaml
   publish:
     needs: build
-    permissions:
```
laurentsimon, line 100: I removed the permissions.

internal/builders/maven/README.md
```diff
@@ -81,11 +81,10 @@ env:
 jobs:
   build:
     permissions:
-      contents: write
       id-token: write
+      contents: read
       actions: read
-      packages: write
-    uses: slsa-framework/slsa-github-generator/.github/workflows/builder_maven_slsa3.yml@main
+    uses: slsa-framework/slsa-github-generator/.github/workflows/builder_maven_slsa3.yml@v1.7.0
```
laurentsimon, line 87: I updated the ref

laurentsimon (approved):

PR #2465, opened 2023-07-24
fix: name of Gradle Action

laurentsimon:


.github/workflows/publish_maven.yml
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+permissions: read-all
```
ianlewis, line 15: nit: Probably it would be ok with just
```suggestion
permissions:
  contents: read
```

ianlewis (approved):

PR #2475, opened 2023-07-26
fix: verify-token allow empty base_ref

laurentsimon:
https://github.com/slsa-framework/slsa-github-generator/pull/2471 added the base_ref in verify-token but this value may be empty.

This PR allows the base_ref to be empty string.


kpk47 (approved):

.github/actions/verify-token/dist/index.js
```diff
@@ -1027,13 +1027,14 @@ exports.validateFieldAnyOf = validateFieldAnyOf;
  * @param name - the name of the value
  * @param actual - the actual value of the field
  * @param expected - the expected value of the field
+ * @param allow_empty - whether the value may be empty
  * @throws Error - if actual and expected don't match or are empty.
  */
-function validateField(name, actual, expected) {
+function validateField(name, actual, expected, allow_empty = false) {
```
ianlewis, line 1033: We could also separate into `validateValue` and `validateNonEmpty`.

ianlewis (approved):

PR #2534, opened 2023-08-04
fix: Remove setup-node from the Node.js publish action

### LINKED ISSUE #2522: [feature][nodejs] Remove `setup-node` from `publish` action
For the Node.js builder users are meant to call `setup-node` themselves for the `publish` action. The `publish` action shouldn't call the `setup-node` action itself.

We however still need to make sure that a recent enough version of `npm` is installed that supports the `--provenance-path` flag. The `publish` flag currently already does this.
 
Related #2320

ianlewis:
Fixes #2522 

Removes `setup-node` from the `publish` action since users are expected to run `setup-node` themselves. This is because `setup-node` makes changes to the local job and may adversely affect the user's later steps.

### DIFF
--- a/actions/nodejs/publish/action.yml
+++ b/actions/nodejs/publish/action.yml
@@ -55,12 +55,6 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Setup Node
-      uses: actions/setup-node@e33196f7422957bea03ed53f6fbb155025ffc7b8 # v3.7.0
-      with:
-        node-version: ${{ inputs.node-version }}
-        node-version-file: ${{ inputs.node-version-file }}
-
     - name: Create temp dir
       id: temp-dir
       shell: bash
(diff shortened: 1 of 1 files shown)


laurentsimon (approved):
shall we do the same for the Maven / Gradle publish Action? 

/cc @AdamKorcz

ianlewis:
> shall we do the same for the Maven / Gradle publish Action?
> 
> /cc @AdamKorcz

I think it's a good idea. Having users run `setup-*`  themselves gives them more control over the state of the job.

laurentsimon:
> > shall we do the same for the Maven / Gradle publish Action?
> > /cc @AdamKorcz
> 
> I think it's a good idea. Having users run `setup-*` themselves gives them more control over the state of the job.

@AdamKorcz please send us PR. I created https://github.com/slsa-framework/slsa-github-generator/issues/2535 for tracking. Thanks

PR #2538, opened 2023-08-04
feat: Add directory input to Maven builder

AdamKorcz:
Adds another `input` for the Maven builder to allow the user to specify the project directory. The current problem this solves is to make https://github.com/slsa-framework/example-package/pull/253 work.

This was suggested by @laurentsimon in https://github.com/slsa-framework/example-package/pull/253#discussion_r1282455298


laurentsimon (approved):
Thanks. You could do some verification on the path to make sure it's under the workspace, something like https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/secure-download-artifact/action.yml#L42-L48

.github/workflows/builder_maven_slsa3.yml
```diff
@@ -27,6 +27,11 @@ on:
         required: false
         default: 17
         type: number
+      directory:
+        description: "Sub-directory to launch the build from. Must be under the workspace. Relative from the root of the file directory when invoking the builder. Should not include a '/' at the end."
```
laurentsimon, line 31: by taking the `realpath` of this path, it should eliminate the trailing `/`. This way users need not have this additional  constraints.

internal/builders/maven/action.yml
```diff
@@ -68,12 +68,17 @@ runs:
       shell: bash
       env:
         SLSA_OUTPUTS_ARTIFACTS_FILE: ${{ inputs.slsa-layout-file }}
+        PROJECT_ROOT: ${{ fromJson(inputs.slsa-workflow-inputs).directory }}
       run: |
+        gh_root_dir=$(pwd)
```
laurentsimon, line 73: let's validate this path

.github/workflows/builder_maven_slsa3.yml
AdamKorcz: Yes, let's do that.

AdamKorcz:
@laurentsimon Could you have another look, please?

ianlewis:
> Thanks. You could do some verification on the path to make sure it's under the workspace, something like https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/actions/secure-download-artifact/action.yml#L42-L48

TBF, I believe the chdir would happen in the "untrusted" build's job anyway. So I don't think this is super important (though necessary if it's used elsewhere).

internal/builders/maven/action.yml
```diff
@@ -68,12 +68,39 @@ runs:
       shell: bash
       env:
         SLSA_OUTPUTS_ARTIFACTS_FILE: ${{ inputs.slsa-layout-file }}
+        PROJECT_ROOT: ${{ fromJson(inputs.slsa-workflow-inputs).directory }}
```
ianlewis, line 71: It's usually a good idea to put an "UNTRUSTED" prefix on user inputs just to make it clear.
```suggestion
        UNTRUSTED_PROJECT_ROOT: ${{ fromJson(inputs.slsa-workflow-inputs).directory }}
```

Then after validation you should explicitly set it to a trusted value.
```shell
validate_path "${UNTRUSTED_PROJECT_ROOT}"

# Directory was validated. Explicitly trust it.
project_root="${UNTRUSTED_PROJECT_ROOT}"
```

internal/builders/maven/action.yml
```diff
@@ -68,12 +68,39 @@ runs:
...
       run: |
+        # Ensure no directory traversal.
+        # NOTE: the actions/download-artifact Action only creates files
+        # in the workspace directory, but this may change in the future.
+        # TODO(#1893): Consolidate directory traversal checks
+        validate_path() {
+          untrusted_path=$1
+          resolved_dir=$(readlink -m "$untrusted_path")
+          wd=$(readlink -m "${GITHUB_WORKSPACE}")
+          if [[ "${wd}" != "" ]] && [[ "${resolved_dir}" != "${wd}"/* ]] && [[ "${resolved_dir}" != "${wd}" ]]; then
```
ianlewis, line 81: I realize this is copied from the `secure-download-artifact` action but do we really need the `"${wd}" != ""` check? What situation does this cover? IIUC this means that if `GITHUB_WORKSPACE` is `""` for some reason this check just passes.

/cc @laurentsimon

laurentsimon: I dont remember why we was needed this at some point. I don't think we need it anymore

AdamKorcz: removed it.

internal/builders/maven/action.yml
AdamKorcz: Done.

internal/builders/maven/action.yml
```diff
@@ -68,12 +68,39 @@ runs:
...
+        validate_path "${gh_root_dir}"
+        validate_path "${project_root}"
+
         mv ./__BUILDER_CHECKOUT_DIR__ ../__BUILDER_CHECKOUT_DIR__ \
           && cd ../__BUILDER_CHECKOUT_DIR__/actions/maven/publish/slsa-hashing-plugin \
           && mvn clean install \
           && cd - \
+          && cd "${project_root}" \
           && mvn package  -Drun.hash.jarfile=true
+        mv $(dirname ${SLSA_OUTPUTS_ARTIFACTS_FILE}) "${gh_root_dir}/../"
```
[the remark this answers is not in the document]
AdamKorcz: Changed. Could you double check that is what you meant?

internal/builders/maven/action.yml
```diff
@@ -68,12 +68,39 @@ runs:
...
+              if [[ "${resolved_dir}" != /tmp/* ]] && [[ "${resolved_dir}" != "/tmp" ]]; then
+                echo "Path is not in the workspace or temp directory: $untrusted_path"
+                exit 1
+              fi
+            fi
+          fi
+        }
+
+        project_root=$(realpath "${PROJECT_ROOT}")
+        gh_root_dir=$(pwd)
```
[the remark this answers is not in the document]
AdamKorcz: Done.

.github/workflows/builder_maven_slsa3.yml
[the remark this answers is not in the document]
AdamKorcz: Removed it.

internal/builders/maven/action.yml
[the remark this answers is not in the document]
laurentsimon: I think I meant that gh_root_dir can be removed and set to GITHUB_WORKSPACE in all places.

laurentsimon: nit: you're missing `"` around `${SLSA_OUTPUTS_ARTIFACTS_FILE}`

AdamKorcz: Fixed.

internal/builders/maven/action.yml
```diff
@@ -68,12 +68,41 @@ runs:
...
+                exit 1
+              fi
+            fi
+          fi
+        }
+
+        validate_path "${UNTRUSTED_PROJECT_ROOT}"
+        project_root=$(realpath "${UNTRUSTED_PROJECT_ROOT}")
+
+        # gh_root_dir is GITHUB_WORKSPACE and is trusted.
```
laurentsimon, line 94: let's at least verify pwd is GITHUB_WORKSPACE..

internal/builders/maven/action.yml
```diff
@@ -68,12 +68,41 @@ runs:
...
+              if [[ "${resolved_dir}" != /tmp/* ]] && [[ "${resolved_dir}" != "/tmp" ]]; then
+                echo "Path is not in the workspace or temp directory: $untrusted_path"
+                exit 1
+              fi
+            fi
+          fi
+        }
+
+        validate_path "${UNTRUSTED_PROJECT_ROOT}"
+        project_root=$(realpath "${UNTRUSTED_PROJECT_ROOT}")
```
laurentsimon, line 92: since we validated the path, why do we need to `realpath`? If it's for removing the `/`? Let's add a comment

internal/builders/maven/action.yml
```diff
@@ -68,12 +68,41 @@ runs:
...
+        # It does not require validation.
+        gh_root_dir=$(pwd)
+
         mv ./__BUILDER_CHECKOUT_DIR__ ../__BUILDER_CHECKOUT_DIR__ \
           && cd ../__BUILDER_CHECKOUT_DIR__/actions/maven/publish/slsa-hashing-plugin \
           && mvn clean install \
           && cd - \
+          && cd "${project_root}" \
           && mvn package  -Drun.hash.jarfile=true
+        mv $(dirname "${SLSA_OUTPUTS_ARTIFACTS_FILE}") "${GITHUB_WORKSPACE}/../"
```
laurentsimon, line 104: is there a reason we sometimes use `gh_root_dir` and sometimes GITHUB_WORKSPACE in the code? Let's be consistent?

laurentsimon:
@AdamKorcz friendly ping to update this PR

internal/builders/maven/action.yml
AdamKorcz: I changed `gh_root_dir` and use `GITHUB_WORKSPACE` instead.

internal/builders/maven/action.yml
AdamKorcz: Added a comment

internal/builders/maven/action.yml
AdamKorcz: I removed `gh_root_dir` altogether since it is `GITHUB_WORKSPACE`. This also removes the `pwd` call which should resolve this comment.

AdamKorcz:
> @AdamKorcz friendly ping to update this PR

Thanks for the ping! Resolved the comments.

internal/builders/maven/action.yml
```diff
@@ -68,12 +68,39 @@ runs:
...
+        # remove trailing "/"'s with `realpath`
+        project_root=$(realpath "${UNTRUSTED_PROJECT_ROOT}")
+
         mv ./__BUILDER_CHECKOUT_DIR__ ../__BUILDER_CHECKOUT_DIR__ \
           && cd ../__BUILDER_CHECKOUT_DIR__/actions/maven/publish/slsa-hashing-plugin \
           && mvn clean install \
           && cd - \
+          && cd "${project_root}" \
           && mvn package  -Drun.hash.jarfile=true
+        mv $(dirname "${SLSA_OUTPUTS_ARTIFACTS_FILE}") "${GITHUB_WORKSPACE}/../"
```
laurentsimon, line 102: last question: why do we need to this `mv` command? `SLSA_OUTPUTS_ARTIFACTS_FILE` file is not expected to be changed by the Action.

AdamKorcz: At this line the context is:
- The builder may not be at `GITHUB_WORKSPACE`
- `slsa-hashing-plugin` has created the artifact file at the location from the `SLSA_OUTPUTS_ARTIFACTS_FILE`. 

The delegator sets  `SLSA_OUTPUTS_ARTIFACTS_FILE` (`slsa-layout-file` below) to a relative path, so if the builder is not at `$GITHUB_WORKSPACE` then we need to move it to `$GITHUB_WORKSPACE/..` so that the builder can find it later.

laurentsimon: thanks, that makes sense! Ill add a comment in your branch and merge.

PR #2622, opened 2023-08-10
docs: fix edit link

ianlewis:


README.md
```diff
@@ -108,7 +108,7 @@ Below is a non-exhaustive list of projects that use the builders in this reposit
...
 
-[Edit this file](/edit/main/README.md) to add your repository!
+[Edit this file](https://github.com/slsa-framework/slsa-github-generator/edit/main/README.md) to add your repository!
```
laurentsimon, line 111: note: this list is purposely non-exhaustive, and tried to list only the top-star repos. Shall we add a line about this? Maybe we make this list a top-30, and say that if your repo has more stars, they can ask to be added?

laurentsimon (approved):

README.md
ianlewis: Yeah, fair enough. I think we can evaluate it on a case-by-case basis. Though I think the value of folks reaching out in some way will be helpful (at least right now). If we start getting too many inappropriate requests we can reevaluate.

Issue #2662, opened 2023-08-20
[bug] Maven / Gradle builders fail to run concurrently


laurentsimon:
If two jobs within the same workflow run the Maven builder, e.g. on two projects hosted on the same repo, the target folders will have a name collision, due to lack of randomization in https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/gradle/action.yml#L120-L123 and similarly in https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/maven/action.yml#L108-L111

The artifact name (not the name on disk) needs to be randomized, as in 
https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/nodejs/action.yml#L84-L94

This requires an update to:
1. The internal Action. It must add randomization and return a new output
2. The publish / download Actions. They need an additional input for the randomized name.
4. e2e tests

NOTE: The nodejs builder has the relevant code as example.

laurentsimon:
Relevant part of the BYOB docs https://github.com/slsa-framework/slsa-github-generator/blob/main/BYOB.md#update-tca

AdamKorcz:
1 and 2 are fixed here: https://github.com/slsa-framework/slsa-github-generator/pull/2665

PR #2665, opened 2023-08-21
chore: make build dirs of java builders unique

AdamKorcz:
Fixes the following from https://github.com/slsa-framework/slsa-github-generator/issues/2662:

- The internal Action. It must add randomization and return a new output
- The publish / download Actions. They need an additional input for the randomized name.

### DIFF
--- a/.github/workflows/builder_gradle_slsa3.yml
+++ b/.github/workflows/builder_gradle_slsa3.yml
@@ -50,17 +50,17 @@ on:
         description: "The sha256 of the provenance attestation uploaded to the workflow run."
         value: "${{ jobs.slsa-run.outputs.attestations-download-sha256 }}"
 
-      target-download-name:
+      build-download-name:
         description: "The name of the build directory uploaded to the workflow run."
         # NOTE: This is an "untrusted" value returned from the build. Technically
         # the build could provide a build directory that doesn't match the
         # provenance but it would fail validation.
-        value: "${{ fromJSON(jobs.slsa-run.outputs.build-artifacts-outputs).package-download-name }}"
+        value: "${{ fromJSON(jobs.slsa-run.outputs.build-artifacts-outputs).build-download-name }}"
 
-      target-download-sha256:
-        description: "The sha256 of the target uploaded to the workflow run."
+      build-download-sha256:
+        description: "The sha256 of the build directory uploaded to the workflow run."
         # NOTE: This is an "untrusted" value returned from the build.
-        value: "${{ fromJSON(jobs.slsa-run.outputs.build-artifacts-outputs).target-download-sha256 }}"
+        value: "${{ fromJSON(jobs.slsa-run.outputs.build-artifacts-outputs).build-download-sha256 }}"
 jobs:
   slsa-setup:
     permissions:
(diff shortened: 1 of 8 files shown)


internal/builders/gradle/action.yml
```diff
@@ -47,6 +47,10 @@ outputs:
       Users should verify the download against this digest to prevent tampering.
     value: ${{ steps.upload-build-dir.outputs.sha256 }}
 
+  package-download-name:
```
laurentsimon, line 50: should this be called `build-download-name` for consistency?

internal/builders/maven/action.yml
```diff
@@ -47,6 +47,10 @@ outputs:
       Users should verify the download against this digest to prevent tampering.
     value: ${{ steps.upload-target.outputs.sha256 }}
 
+  package-download-name:
```
laurentsimon, line 50: nit: let's call it `target-download-name` for consistency?

.github/workflows/builder_gradle_slsa3.yml
```diff
@@ -50,6 +50,13 @@ on:
         description: "The sha256 of the provenance attestation uploaded to the workflow run."
         value: "${{ jobs.slsa-run.outputs.attestations-download-sha256 }}"
 
+      target-download-name:
```
laurentsimon, line 53: is there a reason the gradle builder uses 'build' rather than `target`(Maven) internally for naming the output build directory? Is this a naming convention followed by gradle users? If build is more common, shall we rename `target-` to `build-` in the gradle builder?

/cc @loosebazooka

AdamKorcz: That is exactly the reason. Gradle creates a `build` directory and Maven creates a `target` directory. Will change the naming convention in the builder to reflect that.

AdamKorcz: Should have fixed all names.

internal/builders/maven/action.yml
AdamKorcz: Should have fixed all names.

internal/builders/gradle/action.yml
AdamKorcz: Should have fixed all names.

actions/maven/publish/action.yml
```diff
@@ -86,7 +90,10 @@ runs:
         GPG_KEY_PASS: "${{ inputs.gpg-key-pass }}"
         SLSA_DIR: "${{ inputs.provenance-download-name }}"
         PROVENANCE_FILES: "${{ inputs.provenance-download-name }}"
+        TARGET_DIR: "${{ inputs.target-download-name }}"
       run: |
+        # Change name of downloaded target directory
+        mv "${TARGET_DIR}" ./target
```
laurentsimon, line 96: did you verify this change is needed? I think the secure-download Action will download artifact in a folder that has the name of the original folder uploaded, not the randomized name.

AdamKorcz:
@laurentsimon PTAL again.

laurentsimon:
Thanks for the fast turn around @AdamKorcz !
Let's merge and see if it works in the e2e tests.

AdamKorcz:
@laurentsimon thanks, once it has been merged, I will propagate the updates on the e2e-tests side.

laurentsimon:
some GitHub problems 503 in pre-submits, etc. Re-running

laurentsimon:
Can you force push to trigger the pre-submit again?

PR #2669, opened 2023-08-21
chore: v1.9.0-rc.0

laurentsimon:
#label:release v1.9.0-rc.0

### DIFF
--- a/.github/actions/generate-builder/action.yml
+++ b/.github/actions/generate-builder/action.yml
@@ -62,7 +62,7 @@ runs:
   using: "composite"
   steps:
     - name: Checkout builder repository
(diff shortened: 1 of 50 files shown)


kpk47 (approved):

laurentsimon:
Forcing re-runs because I had the wrong version in the PR description.

laurentsimon:
Looks like we were missing some ref updates, e.g. https://github.com/slsa-framework/slsa-github-generator/blob/v1.8.0/internal/builders/nodejs/action.yml#L90

I updated the RELEASE.md. The pre-submits were updated 1 week after the latest nodejs builder release https://github.com/slsa-framework/slsa-github-generator/pull/2563/files, which is why it was not caught.

PR #2709, opened 2023-09-07
feat: enhance the ko workflow

developer-guy:
To summarize the PR:
- added the `--bare` option since we are defining the whole image name in `KO_DOCKER_REPO`, if we didn't specify the `--bare` option, the result would be different than the specified in that variable.
- used the `--image-refs` flag

### DIFF
--- a/internal/builders/container/README.md
+++ b/internal/builders/container/README.md
@@ -347,11 +347,11 @@ This section explains how to generate non-forgeable SLSA provenance with existin
          echo "${KO_PASSWORD}" | ko login ghcr.io --username "$KO_USER" --password-stdin
 
          # Build & push the image. Save the image name.
-         image_and_digest=$(ko build --tags="${tag}" .)
+         ko build --bare --tags="${tag}" --image-refs .digest
 
          # Output the image name and digest so we can generate provenance.
-         image=$(echo "${image_and_digest}" | cut -d'@' -f1 | cut -d':' -f1)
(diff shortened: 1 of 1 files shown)


internal/builders/container/README.md
```diff
@@ -347,13 +347,13 @@ This section explains how to generate non-forgeable SLSA provenance with existin
...
+         ko build --bare --tags="${tag}" --image-refs .digest
 
          # Output the image name and digest so we can generate provenance.
-         image=$(echo "${image_and_digest}" | cut -d'@' -f1 | cut -d':' -f1)
-         digest=$(echo "${image_and_digest}" | cut -d'@' -f2)
+         image=$(cat .digest | cut -d'@' -f1 | cut -d':' -f1)
+         digest=$(cat .digest| cut -d'@' -f2)
          echo "image=$image" >> "$GITHUB_OUTPUT"
-         echo "digest=$digest" >> "$GITHUB_OUTPUT"
+         echo "digest=$(digest)" >> "$GITHUB_OUTPUT"
```
laurentsimon, line 356: should be `${digest}` instead of `$(digest)`, correct?

laurentsimon (commented):
Thanks @developer-guy

internal/builders/container/README.md
developer-guy: correct, fixed, thanks for correcting me!

laurentsimon (approved):

PR #2758, opened 2023-10-04
fix(detect-workflow-js): workflow ref may contain `@`

sodatea:
We recently encountered several failures when publishing Vite core packages via GitHub Actions.

https://github.com/vitejs/vite/actions/runs/6406860658/job/17392698897
https://github.com/vitejs/vite-plugin-vue/actions/runs/6381049226/job/17317262102

The error log from the first link:
```
npm notice publish Signed provenance statement with source and build information from GitHub Actions
npm notice publish Provenance statement published to transparency log: https://search.sigstore.dev/?logIndex=40473220
npm ERR! code E422
npm ERR! 422 Unprocessable Entity - PUT https://registry.npmjs.org/create-vite - Error verifying sigstore provenance bundle: Invalid GitHub Actions SLSA v1 provenance predicate:
npm ERR! buildDefinition.externalParameters.workflow.ref: refs/tags/create-vite does not match expected value for SourceRepositoryRef: refs/tags/create-vite@5.0.0-beta.0
```

This is because we use tag names like `create-vite@5.0.0-beta.0` for releases, so the resulting `job_workflow_ref` contains multiple `@`[^1]. Somehow GitHub throws away the content after the second `@`.

I don't yet understand how exactly NPM provenance via GitHub Actions works, but after some code search, I think the `.split("@", 2)` in this repository might be one source of the bug, so it might be worth fixing.

[^1]: in this example, it's `vitejs/vite/.github/workflows/publish.yml@refs/tags/create-vite@5.0.0-beta.0`


laurentsimon (commented):
Thanks for the PR! 

Could you create a dedicated function like:
```js
getWorkflowRef()
```

This way we can add a few unit tests to validate the change works and won't break again in the future? Tests are stored under this folder https://github.com/slsa-framework/slsa-github-generator/tree/main/.github/actions/detect-workflow-js/__tests__. We could crate a detect.test.ts file for the tests

sodatea:
Sure thing!
But I saw a test case for the `detectWorkflowFromOIDC` function already exists. So I just added another test case with slightly different inputs, would that be okay?

laurentsimon:
> Sure thing! But I saw a test case for the `detectWorkflowFromOIDC` function already exists. So I just added another test case with slightly different inputs, would that be okay?

You're right that should work too!

.github/actions/detect-workflow-js/src/detect.ts
```diff
@@ -46,7 +46,9 @@ export async function detectWorkflowFromOIDC(
     return Promise.reject(Error("job_workflow_ref missing from OIDC token."));
   }
 
-  const [workflowPath, workflowRef] = jobWorkflowRef.split("@", 2);
+  const firstAtIndex = jobWorkflowRef.indexOf("@");
```
laurentsimon, line 49: please add a comment why we need this, with your example `"vitejs/vite/.github/workflows/publish.yml@refs/tags/create-vite@5.0.0-beta.0"`

laurentsimon (approved):
Thank you!

@ianlewis can you take a quick final look?

ianlewis (approved):
LGTM Thanks!

ianlewis:
Could you run `make dist`  in the `detect-workflow-js` directory and commit the updated files?

sodatea:
Done. Thanks for the guidance!

unicornware:
@ianlewis

is this project supporting [npm provenance for github actions][1]? if so, is this released? i'm getting the same error, but the only "documentation" i've come across is this repo / pr, so i'm unsure. if not, sorry for the noise 😅

[1]: https://github.blog/changelog/2023-09-26-npm-provenance-general-availability/

laurentsimon:
Ah, we have not release after the fix. @ianlewis are we ready for a new release?

ianlewis:
I think we may want to revert https://github.com/slsa-framework/slsa-github-generator/pull/2918 before making a release.  But other than that I think it's ok.

Issue #2863, opened 2023-10-16
[bug] setup-go cache restore fails due to non-default path


bobcallaway:
**Describe the bug**
error while using the go builder:
```
rekor-cli-build (darwin, arm64) / builder
Restore cache failed: Dependencies file is not found in /home/runner/work/rekor-slsa/rekor-slsa. Supported file pattern: go.sum
```

**To Reproduce**
See annotations at https://github.com/bobcallaway/rekor-slsa/actions/runs/6527334459

**Expected behavior**
go module cache should be used 

**Screenshots**

**Additional context**

PR #2864, opened 2023-10-16
tell builder and build-dry about go.sum for go caching

bobcallaway:
Fixes: #2863

### DIFF
--- a/.github/actions/generate-builder/action.yml
+++ b/.github/actions/generate-builder/action.yml
@@ -79,6 +79,7 @@ runs:
       uses: actions/setup-go@fac708d6674e30b6ba41289acaab6d4b75aa0753 # v4.0.1
       with:
         go-version: ${{ inputs.go-version }}
+        cache-dependency-path: "__BUILDER_CHECKOUT_DIR__/go.sum"
 
     - name: Generate builder
       shell: bash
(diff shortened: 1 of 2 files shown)


bobcallaway:
A contaminated cache with the default settings would poison all binaries,
and contamination would not be limited to only that workflow run IIUC.

You could disable the cache that is on by default in setup-go and use a
unique cache key (see
https://github.com/actions/cache#creating-a-cache-key) to limit the blast
radius but still get reuse across the various jobs depending on how you
configure it.

ianlewis:
As @laurentsimon is alluding to, I think our main concern would be how caching affects the build and if the cache could be poisoned from the user's job. I assume the cache is for the whole repo and is reused across jobs? I kind of feel like reusable workflows *should* be operating from their own cache but could be wrong.

@bobcallaway What is the impact to you and rekor? Does not using cache greatly affect your build times?

It's also worth noting we have this same issue for other builders like the Node.js builder.

bobcallaway:
If you’re using setup-go, caching is on by default in v4+; if you don’t
want that, you should disable it IMO. I think we should talk to GitHub
about the scope of caches (org vs project vs workflow) to design it
properly.

Given that rekor will need to generate 13 variants of two binaries, caching
go modules seems like a meaningful optimization.

laurentsimon:
So maybe that's an option we can enable in the workflow to let teams opt-in, like we do for `go-version` and `go-version-file`?

>  I think we should talk to GitHub
about the scope of caches (org vs project vs workflow) to design it
properly.

+1

PR #2867, opened 2023-10-19
docs: Add hishtory as another user of SLSA L3 via Github Actions

ddworken:
If you're ok with another proud user of SLSA. :) hishtory uses it and integrates verificiation as part of an easy to use update flow.


ianlewis (approved):
Thanks!

Issue #2873, opened 2023-10-23
[e2e]: container-based workflow_dispatch main default slsa3 


ianlewis:
Repo: https://github.com/slsa-framework/example-package/tree/main
Run: https://github.com/slsa-framework/example-package/actions/runs/6608223067
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.container-based.workflow_dispatch.main.default.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.container-based.workflow_dispatch.main.default.slsa3.yml
Trigger: workflow_dispatch
Branch: main
Date: Mon Oct 23 03:09:02 UTC 2023

ianlewis:
Repo: https://github.com/slsa-framework/example-package/tree/main
Run: https://github.com/slsa-framework/example-package/actions/runs/6621832493
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.container-based.workflow_dispatch.main.default.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.container-based.workflow_dispatch.main.default.slsa3.yml
Trigger: workflow_dispatch
Branch: main
Date: Tue Oct 24 03:09:33 UTC 2023

ianlewis:
Repo: https://github.com/slsa-framework/example-package/tree/main
Run: https://github.com/slsa-framework/example-package/actions/runs/6635244332
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.container-based.workflow_dispatch.main.default.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.container-based.workflow_dispatch.main.default.slsa3.yml
Trigger: workflow_dispatch
Branch: main
Date: Wed Oct 25 03:22:08 UTC 2023

**Tests are passing now. Closing this issue.**

Issue #2912, opened 2023-10-24
[e2e]: delegator-generic release main checkout slsa3 


ianlewis:
Repo: https://github.com/slsa-framework/example-package/tree/v50.0.90
Run: https://github.com/slsa-framework/example-package/actions/runs/6621520714
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.delegator-generic.release.main.checkout.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.delegator-generic.release.main.checkout.slsa3.yml
Trigger: release
Branch: v50.0.90
Date: Tue Oct 24 02:19:21 UTC 2023

ianlewis:
Repo: https://github.com/slsa-framework/example-package/tree/v50.0.91
Run: https://github.com/slsa-framework/example-package/actions/runs/6634934865
Workflow file: https://github.com/slsa-framework/example-package/tree/main/.github/workflows/e2e.delegator-generic.release.main.checkout.slsa3.yml
Workflow runs: https://github.com/slsa-framework/example-package/actions/workflows/e2e.delegator-generic.release.main.checkout.slsa3.yml
Trigger: release
Branch: v50.0.91
Date: Wed Oct 25 02:26:25 UTC 2023

**Tests are passing now. Closing this issue.**

PR #2913, opened 2023-10-24
fix(deps): sigstore revert from v1.9 to v1.8

enteraga6:
fixes issues #2873 through #2912 (e2e test failures)

Sigstore v1.9 introduces errors on retrieving the token as if the workflow is not seen as GHA but as a browser.

Example workflow using 1.9: https://github.com/enteraga6/bazel_docker_test/actions/runs/6631189448
It fails on slsa-setup with "error retrieving identity token"

Example workflow using 1.8: https://github.com/enteraga6/bazel_docker_test/actions/runs/6631199842
Token generation and retrieval completes successfully.

What causes the issue is currently unknown, but I think that it is a change heading towards the major version bump which will require refactoring of which packages are used to get the token. See here: https://github.com/sigstore/sigstore-js/releases/tag/sigstore%402.0.0, especially [bullet pt 3](https://github.com/sigstore/sigstore-js/commit/d5060c0610d54f0c396b6c2b17bdd5fa24fd3fdc) which conflicts with current signing attestation action.

This PR reverts Sigstore dependencies to 1.8 to allow the workflows to properly run and not fail on setup.


laurentsimon (approved):

ianlewis:
@enteraga6 Can you maybe create an issue on the `sigstore-js` repo with this info? I think that this backwards compatibility breakage might be unintentional?

At least we'll be able to get more clarification.

enteraga6:
@ianlewis Certainly! Tracking the issue here: https://github.com/sigstore/sigstore-js/issues/833

Issue #2956, opened 2023-11-09
[feature]Add COSIGN_REPOSITORY input for the generator_container_slsa3 workflow


saisatishkarra:
**Is your feature request related to a problem? Please describe.**
Cosign allows publishing attestations and provenance to a repository other than the image repository using the environment variable `COSIGN_REPOSITORY`. 
- This will be helpful in scenarios where a single digest/container image is tagged with multiple repositories and would help publish provenance to an external repository to avoid pollution in all image repositories for the same digest.
- Reduces migration effort of tags when moving signatures / provenance attestations for multiple images with same digest
- Helps to provide a single point of repository to external customers during verification of multiple images of same digest

**Describe the solution you'd like**
`COSIGN_REPOSITORY ` must be allowed as an input and set as env variable for cosign command
When set, use this repository to publish provenance instead of image repository.
When unset/empty, publish provenances as per the parsed image repository (default cosign behavior)

**Describe alternatives you've considered**

**Additional context**

laurentsimon:
Thanks. I think we'll need to also have an option in the verifier tool https://github.com/slsa-framework/slsa-verifier, correct?

ianlewis:
Docs seem to be here: https://docs.sigstore.dev/system_config/registry_support/#specifying-registry

I don't see any info about authentication in the docs.

ianlewis:
@saisatishkarra  I think we can add the necessary options here and then add them to the verifier later. I think we just need to make sure we get the implementation right. See our comments on the PR.

laurentsimon:
Yes, please add the necessary options in the PR you already opened. We can help you for slsa-verifier after that. I think the main question we have on the PR is for authentication to the separate registry.

saisatishkarra:
@laurentsimon I have added the Authentication to a separate provenance registry if specified to the existing PR. Please review and help guide the changes needed for the slsa-verifier (client)

saisatishkarra:
@laurentsimon I am also looking forward to a direction to add an explicit option to the verifier and repo/code details to get started. Can you also clarify if the verifier can work with the `COSIGN_REPOSITORY` env variable without the explicit CLI option?

laurentsimon:
Completed in https://github.com/slsa-framework/slsa-github-generator/pull/2962

PR #2962, opened 2023-11-10
feat: Set cosign-repository input and env

saisatishkarra:
Solves: #2956


.github/workflows/generator_container_slsa3.yml
```diff
@@ -70,6 +70,11 @@ on:
         required: false
         type: string
         default: ""
+      cosign-repository:
```
laurentsimon, line 73: Let's rename this to a neutral (non-cosign). Nothing against cosign (we use it!) but it's an implementation detail which need not be exposed to users, I think.
How about `attestation-registry`? 

Also, update the text accordingly: `pushed to this registry instead of image registry`

Does that make sense?

ianlewis: I like the name `attestation-store` but I think most of our workflows use `provenance` for the terminology in inputs etc. so maybe `provenance-store` is better?

.github/workflows/generator_container_slsa3.yml
```diff
@@ -29,9 +29,9 @@ on:
   workflow_call:
     secrets:
       registry-username:
-        description: "Username to log into the container registry."
+        description: "Username to log into the container registry. Username must be specific to `cosign-repository` input when set"
```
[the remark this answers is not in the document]
ianlewis: I assume you need to run `cosign login` for both registries so we would need a username and password for the second registry as well so we can pass it to `cosign login`.
https://github.com/slsa-framework/slsa-github-generator/blob/f9b3411a5e1d95e0b4e9a4a3eee27122e7b6c134/.github/workflows/generator_container_slsa3.yml#L197

.github/workflows/generator_container_slsa3.yml
laurentsimon: provenance is fine with me, good call. I'm thinking that probably `provenance-registry` is better than `provenance-store`, mostly because there's no standard interface to "talk" to an attestation store. If someone wanted to support storing attestations in a GCP bucket, we'd have to parse the `provenance-store` value to identify which kind of store it is. Using `provenance-registry` clearly indicates that "kind" of registry. If we want to support other "kinds" of store, we would add `provenance-bucket`, `provenance-xxx`. Or we keep `provenance-store` but require a scheme in the uri, like "container://docker.io/xxx". Another option is to have a seperate `store-type`.

Wdut?

ianlewis: SGTM. `provenance-registry` it is then.

.github/workflows/generator_container_slsa3.yml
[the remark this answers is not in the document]
saisatishkarra: Added [step](https://github.com/slsa-framework/slsa-github-generator/pull/2962/files#diff-9a1e807670a1f36539cff3ae0d05a3fe21c470a35a0a15e469499ec4b6d7558aR213-R255) for provenance auth when parameters are set.

.github/workflows/generator_container_slsa3.yml
saisatishkarra: resorted to use `provenance-registry`

.github/workflows/generator_container_slsa3.yml
```diff
@@ -34,6 +34,12 @@ on:
         description: "Password to log in the container registry."
       image:
         description: "The OCI image name. This must not include a tag or digest."
+      provenance-registry-username:
+        description: "Username when publishing to provenance registry instead of image registry"
+      provenance-registry-password:
+        description: "Password when publishing to provenance registry instead of image registry"
+      provenance-registry:
+        description: "Provenance is pushed to this registry instead of image registry"
```
laurentsimon, line 42: nit: `If set, provenance is pushed ...`

.github/workflows/generator_container_slsa3.yml
```diff
@@ -70,6 +76,14 @@ on:
         required: false
         type: string
         default: ""
+      provenance-registry-username:
```
laurentsimon, line 79: Using the secret as username should be discourage. Can you add a note to this option and the `registry-username`: 
"This should only be used for high entropy values such as AWS Access Key. See https://github.com/saisatishkarra/slsa-github-generator/tree/feat/cosign-repository-env/internal/builders/container#workflow-inputs" for more information.

.github/workflows/generator_container_slsa3.yml
```diff
@@ -70,6 +76,14 @@ on:
...
+          description: "Username when publishing to provenance registry instead of image registry"
```
laurentsimon, line 80: nit: please add "." at the end of descriptions

.github/workflows/generator_container_slsa3.yml
```diff
@@ -34,6 +34,12 @@ on:
         description: "Password to log in the container registry."
       image:
         description: "The OCI image name. This must not include a tag or digest."
+      provenance-registry-username:
+        description: "Username when publishing to provenance registry instead of image registry"
```
laurentsimon, line 38: nit: `Username when publishing to provenance registry (option 'prove') instead of image registry`

.github/workflows/generator_container_slsa3.yml
```diff
@@ -70,6 +76,14 @@ on:
         required: false
         type: string
         default: ""
+      provenance-registry-username:
+          description: "Username when publishing to provenance registry instead of image registry"
+          required: false
+          type: string
+      provenance-registry:
```
laurentsimon, line 83: Can we remove this from secrets? There's already one in the `inputs` section, correct? Users can use `VAR` to avoid having the registry name hardcoded in their repo https://docs.github.com/en/actions/learn-github-actions/variables

.github/workflows/generator_container_slsa3.yml
```diff
@@ -195,6 +209,50 @@ jobs:
...
+          UNTRUSTED_SECRET_PROVENANCE_REGISTRY: "${{ secrets.provenance-registry }}"
+          UNTRUSTED_INPUT_PROVENANCE_USERNAME: "${{ inputs.provenance-registry-username }}"
+          UNTRUSTED_SECRET_PROVENANCE_USERNAME: "${{ secrets.provenance-registry-username }}"
+          UNTRUSTED_PROVENANCE_PASSWORD: "${{ secrets.provenance-password }}"
+          GCP_ACCESS_TOKEN: "${{ steps.auth.outputs.access_token }}"
+        if: ${{ env.UNTRUSTED_SECRET_PROVENANCE_REGISTRY != '' || env.UNTRUSTED_PROVENANCE_REGISTRY != '' }}
+        run: |
+          set -euo pipefail
+
+          # NOTE: Some registries are of the form <org>/<name>
```
laurentsimon, line 227: Let's add a TODO(xxx) to consolidate this code into a script, since it's a duplicate of the code used for signing into the main registry. Can you create a tracking issue for it?

laurentsimon (commented):
Thanks. A few questions and nits, but looking good!

saisatishkarra:
@laurentsimon Is there a way to autogenerate the readme with updated paramters for the workflow? If so, can you guide on `How-To`?

.github/workflows/generator_container_slsa3.yml
saisatishkarra: #2974

laurentsimon:
> @laurentsimon Is there a way to autogenerate the readme with updated paramters for the workflow? If so, can you guide on `How-To`?

There's no auto-generate :/ 
You can edit the README directly and it should work. We may have a presubmit https://github.com/slsa-framework/slsa-github-generator/blob/main/.github/workflows/scripts/pre-submit.markdown/markdown-toc.sh which runs the markdown-toc command (it's javascript package). It should not be necessary in your case because you're not updating the table of content.

Does that help?

.github/workflows/generator_container_slsa3.yml
saisatishkarra: fixed

.github/workflows/generator_container_slsa3.yml
saisatishkarra: Added note and also readme parameters for both `provenance-registry-username` and `registry-username`

.github/workflows/generator_container_slsa3.yml
saisatishkarra: fixed

.github/workflows/generator_container_slsa3.yml
saisatishkarra: done

.github/workflows/generator_container_slsa3.yml
saisatishkarra: Added todo

.github/workflows/generator_container_slsa3.yml
saisatishkarra: removed `provenance-registry` secret

.github/workflows/generator_container_slsa3.yml
```diff
@@ -28,12 +28,20 @@ defaults:
 on:
   workflow_call:
     secrets:
+      # Note: "This should only be used for high entropy values such as AWS Access Key.
```
laurentsimon, line 31: nit: remove the double quotes

.github/workflows/generator_container_slsa3.yml
```diff
@@ -28,12 +28,20 @@ defaults:
...
+      # See https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/container#workflow-inputs" for more information.
       registry-username:
         description: "Username to log into the container registry."
       registry-password:
         description: "Password to log in the container registry."
       image:
         description: "The OCI image name. This must not include a tag or digest."
+      # Note: "This should only be used for high entropy values such as AWS Access Key.
```
laurentsimon, line 39: nit: remove the double quotes

.github/workflows/generator_container_slsa3.yml
```diff
@@ -70,6 +78,14 @@ on:
         required: false
         type: string
         default: ""
+      provenance-registry-username:
+          description: "Username when publishing to provenance registry (option 'prove') instead of image registry."
```
laurentsimon, line 82: typo: option 'provenance-registry'

.github/workflows/generator_container_slsa3.yml
```diff
@@ -70,6 +78,14 @@ on:
...
+          required: false
+          type: string
+      provenance-registry:
+        description: "If set, Provenance is pushed to this registry instead of image registry."
```
laurentsimon, line 86: nit: Provenance -> provenance

.github/workflows/generator_container_slsa3.yml
```diff
@@ -204,14 +267,17 @@ jobs:
...
         run: |
           set -euo pipefail
 
           # Generate a predicate only.
           predicate_name="predicate.json"
           "$GITHUB_WORKSPACE/$BUILDER_BINARY" generate --predicate="$predicate_name"
 
-          COSIGN_EXPERIMENTAL=1 cosign attest --predicate="$predicate_name" \
+          COSIGN_EXPERIMENTAL=1 \
+          COSIGN_REPOSITORY="${UNTRUSTED_PROVENANCE_REGISTRY}" \
```
laurentsimon, line 279: if the `COSIGN_REPOSITORY` is set to an empty string, will cosign fallback to the image's registry?
Or do we need to check if the value is empty? Might be safer to check?

internal/builders/container/README.md
```diff
@@ -214,14 +214,18 @@ Inputs:
...
 | `gcp-workload-identity-provider` | The full identifier of the Workload Identity Provider, including the project number, pool name, and provider name. If provided, this must be the full identifier which includes all parts:<br>`projects/123456789/locations/global/workloadIdentityPools/my-pool/providers/my-provider` |
 | `gcp-service-account`            | Email address or unique identifier of the Google Cloud service account for which to generate credentials. For example:<br>`my-service-account@my-project.iam.gserviceaccount.com`                                                                                                       |
+| `provenance-registry-username`   | Username when publishing to provenance registry (option 'prove') instead of image registry. Either `provenance-registry-username` input or `provenance-registry-username` secret is required.                                                                                                                                                                                                                 |
```
laurentsimon, line 217: typo: option `provenance-registry`

laurentsimon (commented):
LGTM, thank you!

@ianlewis can you take a second look before we merge?

.github/workflows/generator_container_slsa3.yml
saisatishkarra: I have tested this locally and if `COSIGN_REPOSITORY` is set to an empty string it will fallback to the image registry. However, it is better to make this explicit by adding a check!!

.github/workflows/generator_container_slsa3.yml
saisatishkarra: done

.github/workflows/generator_container_slsa3.yml
saisatishkarra: done

.github/workflows/generator_container_slsa3.yml
saisatishkarra: fixed

.github/workflows/generator_container_slsa3.yml
saisatishkarra: fixed

.github/workflows/generator_container_slsa3.yml
saisatishkarra: Added an if condition to check for non-empty string and explicitly export `COSIGN_REPOSITORY`

internal/builders/container/README.md
saisatishkarra: fixed

.github/workflows/generator_container_slsa3.yml
saisatishkarra: [Lines](https://github.com/slsa-framework/slsa-github-generator/pull/2962/files#diff-9a1e807670a1f36539cff3ae0d05a3fe21c470a35a0a15e469499ec4b6d7558aR278-R280) added

saisatishkarra:
cc: @laurentsimon  / @ianlewis  for some eyes!! Looking to make similar changes on the verifier with some guidance on expected direction!

laurentsimon:
Take a look at the linter failures. Should be easy to fix

saisatishkarra:
> Take a look at the linter failures. Should be easy to fix

Fixed linting issues and indentation. Waiting for the checks to run!!

internal/builders/container/README.md
```diff
@@ -214,14 +214,18 @@ Inputs:
...
 | `gcp-service-account`            | Email address or unique identifier of the Google Cloud service account for which to generate credentials. For example:<br>`my-service-account@my-project.iam.gserviceaccount.com`                                                                                                       |
+| `provenance-registry-username`   | Username when publishing to provenance registry (option 'provenance-registry') instead of image registry. Either `provenance-registry-username` input or `provenance-registry-username` secret is required.                                                                                                                                                                                                                 |
+| `provenance-registry`            | If set, provenance is pushed to this registry instead of image registry.                                                                                          |
```
ianlewis, line 218: nit: Add an example.
```suggestion
| `provenance-registry`            | If set, provenance is pushed to this registry instead of image registry. (e.g. `gcr.io/my-new-repo`)                                                                                          |
```

ianlewis (approved):
LGTM

curiositycasualty:
Thanks for your time spent on this @laurentsimon & @ianlewis

Issue #3015, opened 2023-12-02
[bug]: macOS error with `base64` on `create-base64-subjects-from-file` action


sgammon:
**Describe the bug**
When running [`create-base64-subjects-from-file`](https://github.com/sgammon/slsa-github-generator/tree/main/actions/generator/generic/create-base64-subjects-from-file) on a macOS-based runner, the command `base64 -w0` is issued; the `base64` command on macOS does not support this flag.

[`man base64`](https://linux.die.net/man/1/base64) on Linux says:
```
-w, --wrap=COLS
Wrap encoded lines after COLS character (default 76). Use 0 to disable line wrapping.
```

`man base64` on macOS (Ventura 13.5.1):
```
NAME
     base64 – Encode and decode using Base64 representation

SYNOPSIS
     base64 [-h | -D | -d] [-b count] [-i input_file] [-o output_file]

...

OPTIONS
     The following options are available:
     -b count
     --break=count                   Insert line breaks every count characters.  Default is 0, which generates an unbroken stream.

     -d
     -D
     --decode                        Decode incoming Base64 stream into binary data.

     -h
     --help                          Print usage summary and exit.

     -i input_file
     --input=input_file              Read input from input_file.  Default is stdin; passing - also represents stdin.

     -o output_file
     --output=output_file            Write output to output_file.  Default is stdout; passing - also represents stdout.
```

The `-w0` flag is used to avoid line wraps. The equivalent flag on macOS would be `-b0`, but that is already the default:
```
     -b count
     --break=count                   Insert line breaks every count characters.  Default is 0, which generates an unbroken stream.
```

**To Reproduce**

1. Create a Github Actions workflow which `runs-on` any macOS version (for example, `macos-latest`)
2. Generate some assets to stamp
3. Use the SLSA generic generator
4. For certain code paths, observe the failure, which exits with code `64`:

```
Artifact release_digests.txt-[...] has been successfully uploaded!
Run set -euo pipefail
{
  "artifact_name": "release_digests.txt-[...]",
  "sha256": "[...]",
  "filename": "release_digests.txt"
}
base64: invalid argument -w0
Usage:	base64 [-hDd] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -Dd, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)
Error: Process completed with exit code 64.
```

[...]

PR #3016, opened 2023-12-02
fix: generic generator on macos runners

sgammon:
Fixes an issue with the `base64` call to generate subject data on macOS-based runners.

Fixes and closes slsa-framework/slsa-github-generator#3015.

### DIFF
--- a/actions/generator/generic/create-base64-subjects-from-file/action.yml
+++ b/actions/generator/generic/create-base64-subjects-from-file/action.yml
@@ -66,7 +66,7 @@ runs:
 
         object="{\"artifact_name\": \"${UNTRUSTED_ARTIFACT_NAME}\", \"sha256\": \"${SHA256}\", \"filename\": \"${UNTRUSTED_FILENAME}\"}"
 
-        if test "$(uname)" = "Darwin"
+        if test "$RUNNER_OS" = "macOS"
         then
             base64_object=$(echo "$object" | base64)
         else
--- a/actions/generator/generic/create-base64-subjects-from-file/action.yml
+++ b/actions/generator/generic/create-base64-subjects-from-file/action.yml
@@ -65,7 +65,12 @@ runs:
         set -euo pipefail
 
         object="{\"artifact_name\": \"${UNTRUSTED_ARTIFACT_NAME}\", \"sha256\": \"${SHA256}\", \"filename\": \"${UNTRUSTED_FILENAME}\"}"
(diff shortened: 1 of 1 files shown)


laurentsimon (approved):
Thanks!

actions/generator/generic/create-base64-subjects-from-file/action.yml
```diff
@@ -65,7 +65,12 @@ runs:
         set -euo pipefail
 
         object="{\"artifact_name\": \"${UNTRUSTED_ARTIFACT_NAME}\", \"sha256\": \"${SHA256}\", \"filename\": \"${UNTRUSTED_FILENAME}\"}"
-        echo "$object" | jq
-        base64_object=$(echo "$object" | base64 -w0)
 
+        if test "$(uname)" = "Darwin"
```
laurentsimon, line 69: can we use `RUNNER_OS` instead?

actions/generator/generic/create-base64-subjects-from-file/action.yml
```diff
@@ -65,7 +65,12 @@ runs:
...
+        then
+            base64_object=$(echo "$object" | base64)
```
ianlewis, line 71: I think `base64` on MacOS just doesn't have the `-w` option so we'll still need to remove the newlines as GitHub Actions doesn't support newlines in inputs.

We might be able to use `tr` instead regardless of the `RUNNER_OS`?

```shell
base64_object=$(echo "$object" | base64 | tr -d \\n)
```

sgammon: @ianlewis From the linked issue:

> The -w0 flag is used to avoid line wraps. The equivalent flag on macOS would be -b0, but that is already the default

Is that sufficient, in your view, or should we add a second check?

sgammon:
@ianlewis / @laurentsimon Thank you for the quick review! We are using this on our fork and all seems to be working fine. Here's a sample intoto signature produced from this revision

[buildless.darwin-amd64.intoto.jsonl.zip](https://github.com/slsa-framework/slsa-github-generator/files/13573815/buildless.darwin-amd64.intoto.jsonl.zip)

It was produced in a private pipeline, but published as part of [this release](https://github.com/buildless/cli/releases/tag/1.0.0-beta9). As far as I can tell, all seems good, but I'm rather new to SLSA.

Thank you guys for all your hard work on this framework!

PR #3298, opened 2024-02-28
fix: ensure no self-hosted runners for generic generators

### LINKED ISSUE #1868: [feature] Generator build level
The new v1.0 specs, iiuc, no longer has a "provenance" level 3. I think this means the generators would become level 2.
We could probably make them level 3 if the generator is able to check tat the caller workflow uses only GitHub-hosted runners.

Is there a way to do that?

The list of GitHub runners is available at https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.

Is there an API to retrieve that list at runtime? If so, we could compare it to the list of runners used in the workflow.

There's also this API https://stackoverflow.com/questions/70027430/get-a-list-of-github-runners, which I have not tried. It may require org-level tokens to query

ramonpetgrave64:
# Update

We're putting this PR on hold until we can get an answer in 
 - https://github.com/orgs/community/discussions/111347

# Summary

Fixes #1868 

Fails the generator workflows when they detect that other jobs in the calling workflow are targeting self-hosted runners.

A new function first fetches a list of all jobs in the calling workflow, examines their specified target `runs-on` labels, and then compares against any known self-hosted runners' labels. 

We do this extra comparison because a user need not specify the target `runs-on` label `"self-hosted"` in order to actually run on their self-hosted runners: If they have a runner with a custom labels `"cloudtop"` or even `"ubuntu-latest"`, they need only specify `runs-on: "cloudtop"` or `rune-on: "ubuntu-latest"` to make their job run on their self-hosted runner.

## Personal Access Token

The call to check the list of self-hosted runners requires a Personal Access Token with administration:read permissions. This is a breaking change, and will also require updates to many of our e2e tests bth in this repo and example-package.

## Defeatable

A repo owner with bad intentions could still get past this check: after the actual artifact build and before our checks, they can edit the properties of their self-hosted runner to remove the label in question.

## Testing Process

- Added unit tests for a new `ensureOnlyGithubHostedRunners()` in `detect.ts`.
- Added more unit tests to `main.ts`, wich calls `ensureOnlyGithubHostedRunners()`
- Added another workflow `pre-submit.actions-for-pr.yml` that does a checkout on the PR's copy of of the code before doing a test on each of the github actions.

* [ ] TODO: add e2e tests to slsa-framework/example-package 
* OR
* [ ] TODO: add a pre-submit test to this workflow that attempts to run a job on a self-hosted runner (after we make one) and checks the (`needs.detect-env.outputs.outcome`).

Forked slsa-github-generator and ran the generator within my won custom workflow that specifies various combinations of self-hosted anad non self-hosted runner labels.
 * https://github.com/ramonpetgrave64/my-example-gradle-project/actions/runs/8085634419/job/22093629374?pr=2

## Checklist

- [x] Add changes to the [CHANGELOG](./../CHANGELOG.md) if applicable.


ramonpetgrave64:
I think I'm currently having an issue in CI because I'm using the `token` as a secret, and because my PR is coming from a fork.
https://github.com/actions/toolkit/issues/465#issuecomment-630227277

```
Failed to retrieve OIDC token. This may be due to missing id-token: write permissions.
Error: Parameter token or opts.auth is required
Error: No repository detected.
```

I may be able to workaorund this with

```
with:
  token: ${{ secrets.token || github.token }}
```

laurentsimon:
> We do this extra comparison because a user need not specify the target `runs-on` label `"self-hosted"` in order to actually run on their self-hosted runners: If they have a runner with a custom labels `"cloudtop"` or even `"ubuntu-latest"`, they need only specify `runs-on: "cloudtop"` or `rune-on: "ubuntu-latest"` to make their job run on their self-hosted runner.

Does this not contradict https://github.com/slsa-framework/slsa-github-generator/issues/1868#issuecomment-1966251208? Shall we create an issue on https://github.com/actions/runner to ask to update the doc?

.github/workflows/generator_container_slsa3.yml
```diff
@@ -42,6 +42,9 @@ on:
         description: "Username when publishing to provenance repository (option 'provenance-repository') instead of image repository."
       provenance-registry-password:
         description: "Password when publishing to provenance repository instead of image repository."
+      token:
```
laurentsimon, line 45: the token is available without the need to pass it explicitly, so I don't think we need this option?

.github/workflows/generator_generic_slsa3.yml
```diff
@@ -129,6 +133,8 @@ jobs:
         id: detect
         continue-on-error: true
         uses: slsa-framework/slsa-github-generator/.github/actions/detect-workflow-js@main
+        with:
```
laurentsimon, line 136: same comment about the token being available

.github/actions/detect-workflow-js/src/detect.ts
```diff
@@ -134,3 +135,58 @@ export async function detectWorkflowFromContext(
...
+ uses: slsa-framewrok/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml'
+   secrets:
+     token: ${{ secrets.MY_TOKEN_WITH_EXTRA_PERM }}
+  ```
+*/
+export async function ensureOnlyGithubHostedRunners(
+  repoName: string,
+  token: string,
+): Promise<void> {
+  const [owner, repo] = repoName.split("/");
```
laurentsimon, line 156: do you know what permissions are needed to read this API? Does it require admin permissions?

laurentsimon (commented):
LGTM overall. Can we create a dedicated action for the detection of self-hosted runners? It will make the code more re-usable when we need it.

ramonpetgrave64:
> LGTM overall. Can we create a dedicated action for the detection of self-hosted runners? It will make the code more re-usable when we need it.

I think you're right. And it should be easier to write a e2e test for that way

.github/actions/detect-workflow-js/src/detect.ts
ramonpetgrave64: Yes, it needs a user-generated token with actions: read and administration:read against the repo. In the new action I'll write up the docs for it.

laurentsimon: I'm worried about this requirement `admin:read`. This is a hard ask. @ianlewis @kpk47 @pnacht @haydentherapper wdut?

laurentsimon:
> > We do this extra comparison because a user need not specify the target `runs-on` label `"self-hosted"` in order to actually run on their self-hosted runners: If they have a runner with a custom labels `"cloudtop"` or even `"ubuntu-latest"`, they need only specify `runs-on: "cloudtop"` or `rune-on: "ubuntu-latest"` to make their job run on their self-hosted runner.
> 
> Does this contradict [#1868 (comment)](https://github.com/slsa-framework/slsa-github-generator/issues/1868#issuecomment-1966251208)? Shall we create an issue on https://github.com/actions/runner to ask to update the doc?

Please cc me on the issue you create on their repo

.github/actions/detect-workflow-js/src/detect.ts
haydentherapper: This would give me pause if I saw I needed to generate an admin PAT. Which of these APIs needs this permission?

ramonpetgrave64: `admin: read` and `actions: read`, but the user can limit their token's permissions to a single repo.

pnacht: > Which of these APIs needs this permission?

`octokit.rest.actions.listSelfHostedRunnersForRepo` at least needs it.

But I agree, requiring this permission can be a hard sell...

---

Question: can't this permission be replaced by having a list of [known GitHub runner images](https://github.com/actions/runner-images?tab=readme-ov-file#available-images)?

Then, if `runs-on` contains only elements in that list, it's assumed to be a GitHub image.

I don't know if a self-hosted image with label `ubuntu-latest` will ever be chosen... But even if so, I assume it's highly unreliable (you never know if your runner will be chosen before the GH runner), unless GH gives preference to self-hosted runners whenever possible.

This can either be hardcoded into the generator (therefore requiring updates whenever GitHub releases a new image) or be dynamically parsed from the README (fragile!). I unfortunately couldn't find a REST API endpoint or something of the sort we can query to get all the images.

ramonpetgrave64: That's right. Github doesn't publish a full list, and if I assign my runner with the label "ubuntu-latest" or "macos-14-large", then my job can still run on my self-hosted runner. There's nothing stopping me from assigning my runner with the labels that github already uses.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 <!-- toc -->
 
 - [Unreleased](#unreleased)
+  - [Unreleased: Generic Generator](#unreleased-generic-generator)
```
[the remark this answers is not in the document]
ramonpetgrave64: added

.github/workflows/generator_container_slsa3.yml
ramonpetgrave64: it will be a different personal access token

.github/workflows/generator_generic_slsa3.yml
ramonpetgrave64: it will be a different personal access token

ramonpetgrave64:
> > LGTM overall. Can we create a dedicated action for the detection of self-hosted runners? It will make the code more re-usable when we need it.
> 
> I think you're right. And it should be easier to write a e2e test for that way

I've added it as a separate action. Still w work-in-progress, however, until we can get a good e2e story.

.github/actions/detect-workflow-js/src/detect.ts
laurentsimon: @steiza Any guidance how we can figure out whether a run uses self-hosted runner vs GitHub-hosted runners, and without admin permissions - https://github.com/slsa-framework/slsa-github-generator/issues/1868?

laurentsimon: @josepalafox @patrick-knight

ramonpetgrave64: I've also opened a ticket with github

- https://github.com/orgs/community/discussions/111347
- https://github.com/slsa-framework/slsa-github-generator/issues/1868#issuecomment-1985901738

ramonpetgrave64:
I found a way, plus a new `runner.environment` context we should probably use in all our workflows.
 - https://github.com/orgs/community/discussions/111347#discussioncomment-10490619
 - https://github.com/orgs/community/discussions/111347#discussioncomment-10490778

ianlewis:
> I found a way by scanning workflow logs, plus a new `runner.environment` context we should probably use in all our workflows.
> 
> * https://github.com/orgs/community/discussions/111347#discussioncomment-10490619
> * https://github.com/orgs/community/discussions/111347#discussioncomment-10490778

Yeah, I think `runner.environment` is really what we are looking for. I seem to remember that a job looking up its own logs was problematic for some reason but I might be mistaken as I can't remember specifics.

ramonpetgrave64:
After trying out the context, I see it only works at the step level, and not at the job level. This means we can prevent accidental uses of self-hosted runners. But to try to fail the job, I think will always require an execution of the ‘run’ on the runner, like ‘exit 1’. If the runner was truly malicious, then running that command is a self-certification. 

At best, I found that a separate job that performs the signing can examine the logs of the job that performs the build. And then the verifier can examine the certificate/oidc claims to assert that the signing job’s runner was github-hosted. This then implies that the build job’s runner was github-hosted.

ianlewis:
> After trying out the context, I see it only works at the step level, and not at the job level. This means we can prevent accidental uses of self-hosted runners. But to try to fail the job, I think will always require an execution of the ‘run’ on the runner, like ‘exit 1’. If the runner was truly malicious, then running that command is a self-certification.
> 
> At best, I found that a separate job that performs the signing can examine the logs of the job that performs the build. And then the verifier can examine the certificate/oidc claims to assert that the signing job’s runner was github-hosted. This then implies that the build job’s runner was github-hosted.

I wonder if we can use `if: {{ runner.environment }} == "github-hosted"` and check to see that the job has run? Or maybe we need to put that check on every job that we run?

Regardless I think we should add that check to the verifier too.

PR #3299, opened 2024-02-29
fix: Switch to newer DSSE rekor type

haydentherapper:
The intoto v001 type does not persist signatures of the DSSE envelope, as noted in https://github.com/sigstore/rekor/pull/973. We introduced an intoto v002 type shortly after to fix this, but since then, we've introduced another newer type, DSSE v001, which also does not persist the attestation in Rekor (as we discourage using Rekor as storage).

I also updated the verifier in https://github.com/slsa-framework/slsa-verifier/pull/742 to search for both Rekor entry types.

# Summary

...

## Testing Process

...

## Checklist

- [ ] Review the contributing [guidelines](./../CONTRIBUTING.md)


laurentsimon (commented):
Thanks @haydentherapper . I think this change works. Shall we consider it a breaking change or not? I think so. Wdut?

nit: Can you update the CHANGELOG.md file?

haydentherapper:
It's sort of a breaking change. It's not a breaking change for the generator itself, but it's a breaking change in that it'll require verifiers to be updated. When we made this change in Cosign, we didn't call it a breaking change since Cosign is both the signer and verifier, but given the verification tooling is separate here, probably would be good to have this as a minor rather than patch release.

haydentherapper:
Updated changelog

laurentsimon (approved):

laurentsimon:
There's a linter failing.

haydentherapper:
Fixed

haydentherapper:
Oh it is already. Hm..

laurentsimon:
pre-submit failing is not your fault. I'll send a PR to fix it.

laurentsimon:
PR is out from this morning https://github.com/slsa-framework/slsa-github-generator/pull/3454 :) Need to be merged first

PR #3313, opened 2024-03-08
fix: remove attestation-name input and output

### LINKED ISSUE #3031: [bug] Opaque failure in `final` step
**Describe the bug**

I'm attempting to publish a new version of the Python [`id`](https://github.com/di/id) package, using the package's GitHub Actions [release workflow](https://github.com/di/id/actions/workflows/release.yml).

This is what our current provenance generation step looks like:

```yaml
  generate-provenance:
    needs: [build]
    name: Generate build provenance
    permissions:
      actions: read   # To read the workflow path.
      id-token: write # To sign the provenance.
      contents: write # To add assets to a release.
    # Currently this action needs to be referred by tag. More details at:
    # https://github.com/slsa-framework/slsa-github-generator#verification-of-provenance
    uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.9.0
    with:
      attestation-name: provenance-id-${{ github.event.release.tag_name }}.intoto.jsonl
      base64-subjects: "${{ needs.build.outputs.hashes }}"
      compile-generator: true # Workaround for https://github.com/slsa-framework/slsa-github-generator/issues/1163
      upload-assets: true
```

When our release workflow is triggered, this step runs as expected, and the interior `detect-env` and `generator` steps appear to succeed. The `upload-assets` step is then skipped, and `final` fails opaquely (apparently checking some kind of success state, without an error message):

```
  echo "outcome=$([ "$SUCCESS" == "true" ] && echo "success" || echo "failure")" >> "$GITHUB_OUTPUT"
```

You can see that in the release workflow logs here: https://github.com/di/id/actions/runs/7184211112/job/19564897464

This error persists even when the job is re-run.

**To Reproduce**

I don't have an easy reproducer, unfortunately. I'm happy to perform debugging steps on the `di/id` repository, however.

**Expected behavior**

I expected the `generate-provenance` step to succeed, as it has for previous release runs.

### LINKED ISSUE #3072: [feature] `attestation-name` deprecation warning
We should emit a warning if users are using the deprecated `attestation-name` input.

Related #3031

ramonpetgrave64:
# Summary

- Fixes #3031
- Fixes #3072 
- Removes the `attestation-name` input and output from the `generator_generic_slsa3.yml`, which has been deprecated for `provenance-name`.

## Testing Process

We cannot properly test workflow changes without first merging to `main`, and revert if tests fail. After merging we will

- [ ] trigger a manual run of `pre-submit e2e generic default`
- [ ] wait for nightly e2e tests to report success

## Checklist

- [x] Add changes to the [CHANGELOG](./../CHANGELOG.md) if applicable.


.github/workflows/builder_gradle_slsa3.yml
```diff
@@ -40,7 +40,7 @@ on:
       provenance-name:
         description: "The file name of the attestation upload artifact."
         # NOTE: This is an "untrusted" value returned from the build.
-        value: "${{ fromJSON(jobs.slsa-run.outputs.build-artifacts-outputs).attestation-name }}"
```
laurentsimon, line 43: I'm not sure if we should update these. They are only used by BYOB integrations. If we want to update them, we'd also need to update our internal usage of it, like https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/gradle/create_attestation.sh, same for maven, etc.
I'm thinking we can skip them and keep this PR simpler

Let's keep it simple and only update the option for the generic generator.

laurentsimon (commented):
nit: can you edit the CHANGELOG

ramonpetgrave64:
@laurentsimon Yeah I did edit the changelog, but not as a breaking change since we're removing an already deprecated thing. 

- https://github.com/slsa-framework/slsa-github-generator/pull/3313/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR103

laurentsimon:
> @laurentsimon Yeah I did edit the changelog, but not as a breaking change since we're removing an already deprecated thing.
> 
> * https://github.com/slsa-framework/slsa-github-generator/pull/3313/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR103

Let's make it breaking, because some folks are still using it

ramonpetgrave64:
> Let's make it breaking, because some folks are still using it. Wdut?

Alright

laurentsimon (approved):

Issue #3393, opened 2024-03-20
tracking reverting breaking changes before new emergency release


ramonpetgrave64:
tracking revert PRs before the next emergency release

ramonpetgrave64:
getting some errors when trying to revert some PRs 
<img width="1118" alt="image" src="https://github.com/slsa-framework/slsa-github-generator/assets/32398091/d9e10f2d-6127-485e-8d9b-d39920f0fa4d">

kpk47:
Can we just create a new release branch from the last release date and apply the sigstore fix to that? We won't have the chore releases since then, but we'll at least have the sigstore fix.

ramonpetgrave64:
@kpk sounds good, we can bramch from 
https://github.com/slsa-framework/slsa-github-generator/commit/07e64b653f10a80b6510f4568f685f8b7b9ea830
and then apply the fixes from
https://github.com/slsa-framework/slsa-github-generator/pull/3355

But I think there may be an adversarial test for releasing from a non-main branch.

laurentsimon:
If we revert:

https://github.com/slsa-framework/slsa-github-generator/pull/3313 (I'll send a PR)
Need to update the CHANGELOG for Hayden's PR
example-package revert the v3->v4 action

I'm more confident doing this in the main branch than in a different branch. The e2e tests don't work on different branch. There's just one PR left to revert https://github.com/slsa-framework/slsa-github-generator/pull/3313

kpk47:
Okay, sounds good. Let's go with your plan.

I've got a release branch together already, so we've got that in case we need it for some reason.

PR #3394, opened 2024-03-20
chore: revert "feat: enhance the ko workflow"

ramonpetgrave64:
Reverts slsa-framework/slsa-github-generator#2709

#3393


laurentsimon:
This PR was only an enhancement so we need not revert it. Closing it. Please-re-open if you disagree.

PR #3416, opened 2024-03-21
chore: v1.9.1

kpk47:
#label:release v1.9.1


laurentsimon:
Should be v1.10.0, because we also have features in this release. I'll update the PR

PR #3570, opened 2024-04-15
fix: deadlock and improve debugging experience

### LINKED ISSUE #3569: [bug] docker generator should ouput logs as command is running
**Describe the bug**
The docker generator only outputs logs once the build command has returned, see https://github.com/slsa-framework/slsa-github-generator/blob/4534a0b24500dfdd11685f2950cba9a35086c4d2/internal/builders/docker/pkg/builder.go#L484. This creates a problem when attempting to debug a process that hangs in CI, never succeeding and never failing. 

**Expected behavior**
For verbose output, the generator should print logs as they are created by the build command.

### LINKED ISSUE #3571: [bug] slsa-framework/slsa-github-generator deadlocks if builder outputs a lot of data over stdio
**Describe the bug**
The slsa-framework/slsa-github-generator deadlocks if builder outputs a lot of data over both stdout and stderr. The reason is that the current implementation reads from stdout and stderr sequentially, which is a known cause of deadlocks: https://github.com/golang/go/issues/16787.

Because of #3569, it also does not print any logs in the process, the action just times out.

**To Reproduce**

1. Run the action in https://github.com/project-oak/oak/pull/5018, which happens to produce a lot of output over stdio. 

I'll create a simpler reproduction upon request. But the bug should be fairly obvious when looking at `internal/builders/docker/pkg/builder.go` and https://github.com/golang/go/issues/16787.

**Expected behavior**
No deadlock.

cc @tiziano88 @jblebrun

jul-sh:
# Summary

Fixes #3571. Fixes #3569. Shoutout to @jblebrun who wrote most of the code in this PR and debugged this issue together with me. 

## Testing Process

...

## Checklist

- [ ] Review the contributing [guidelines](./../CONTRIBUTING.md)

### DIFF
--- a/internal/builders/docker/pkg/builder.go
+++ b/internal/builders/docker/pkg/builder.go
@@ -23,6 +23,7 @@ package pkg
 // Docker image.
 
 import (
+	"bufio"
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
@@ -36,6 +37,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	intoto "github.com/in-toto/in-toto-golang/in_toto"
 	slsa1 "github.com/in-toto/in-toto-golang/in_toto/slsa_provenance/v1"
@@ -476,34 +478,87 @@ func (c *GitClient) checkoutGitCommit() error {
 	return nil
 }
 
-// saveToTempFile creates a tempfile in `/tmp` and writes the content of the
-// given readers to that file.
-func saveToTempFile(verbose bool, readers ...io.Reader) ([]string, error) {
-	var files []string
-	for _, reader := range readers {
-		bytes, err := io.ReadAll(reader)
-		if err != nil {
-			return files, err
-		}
+type tempFileResult struct {
+	File *os.File
+	Err  error
+}
+
+// A helper function used by saveToTempFile to process one individual file.
+// This should be called in a goroutine, and the channels passed in should be owned by the caller,
+// and remain open until the goroutine completes.
+func saveOneTempFile(verbose bool, reader io.Reader, fileChannel chan tempFileResult, printChannel chan string) {
+	var allBytes []byte
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		bytes := scanner.Bytes()
+		allBytes = append(allBytes, bytes...)
+		allBytes = append(allBytes, '\n')
 
 		if verbose {
-			if len(bytes) > 0 {
-				fmt.Print("\n\n>>>>>>>>>>>>>> output from command <<<<<<<<<<<<<<\n")
-				fmt.Printf("%s", bytes)
-				fmt.Print("=================================================\n\n\n")
-			}
+			printChannel <- string(bytes)
 		}
+	}
 
-		tmpfile, err := os.CreateTemp("", "log-*.txt")
-		if err != nil {
-			return files, fmt.Errorf("couldn't create tempfile: %v", err)
-		}
+	tmpfile, err := os.CreateTemp("", "log-*.txt")
+	if err != nil {
+		fileChannel <- tempFileResult{Err: err}
+		return
+	}
+	defer tmpfile.Close()
 
-		if _, err := tmpfile.Write(bytes); err != nil {
-			tmpfile.Close()
-			return files, fmt.Errorf("couldn't write bytes to tempfile: %v", err)
+	if _, err := tmpfile.Write(allBytes); err != nil {
+		fileChannel <- tempFileResult{Err: fmt.Errorf("couldn't write bytes to tempfile: %v", err)}
+	} else {
+		fileChannel <- tempFileResult{File: tmpfile}
+	}
+}
+
+// saveToTempFile creates a tempfile in `/tmp` and writes the content of the
+// given readers to that file.
+// It processes all provided readers concurrently.
+func saveToTempFile(verbose bool, readers ...io.Reader) ([]string, error) {
+	if verbose {
+		fmt.Print("\n\n>>>>>>>>>>>>>> output from command <<<<<<<<<<<<<<\n")
+	}
+	var wg sync.WaitGroup
+	// We need to make sure the fileChannel has enough buffere space to hold everything,
+	// since it won't be processed until the very end.
+	fileChannel := make(chan tempFileResult, len(readers))
+	printChannel := make(chan string)
+
+	// Start a goroutine to process each Reader concurrently.
+	for _, reader := range readers {
+		wg.Add(1)
+		go func(reader io.Reader) {
+			defer wg.Done()
+			saveOneTempFile(verbose, reader, fileChannel, printChannel)
+		}(reader)
+	}
+
+	// Close the channel once all goroutines have finished.
+	// We do the waiting in a goroutine so that we can move on
+	// to the lines below that handle printing the values passed on
+	// the print channel.
+	go func() {
+		wg.Wait()
+		close(printChannel)
+		close(fileChannel)
+	}()
+
+	// Print the output as it comes in.
+	// Once all of the provided readers have been closed, the WaitGroup
+	// will move to the Done state, completing this loop and allowing us to
+	// move to the filename collection below.
+	for line := range printChannel {
+		fmt.Println(line)
+	}
+
+	var files []string
+	for result := range fileChannel {
+		if result.Err != nil {
+			return nil, result.Err
 		}
-		files = append(files, tmpfile.Name())
+		files = append(files, result.File.Name())
 	}
 
 	return files, nil


internal/builders/docker/pkg/builder.go
```diff
@@ -481,25 +481,39 @@ func (c *GitClient) checkoutGitCommit() error {
...
-		if verbose {
-			if len(bytes) > 0 {
+		for _, reader := range readers {
+			if verbose {
 				fmt.Print("\n\n>>>>>>>>>>>>>> output from command <<<<<<<<<<<<<<\n")
-				fmt.Printf("%s", bytes)
+			}
+
+			scanner := bufio.NewScanner(reader)
+			for scanner.Scan() {
```
laurentsimon, line 492: is there a difference between using scanner instead of the original `fmt.Printf("%s", bytes)`?

jul-sh: the scanner incrementally returns individual lines of the reader, we then print each. the original solution read all bytes, and then printed them at once.

jul-sh: see https://pkg.go.dev/bufio#Scanner.Split

jul-sh:
cc @tiziano88 also

internal/builders/docker/pkg/builder.go
```diff
@@ -481,25 +481,39 @@ func (c *GitClient) checkoutGitCommit() error {
...
 	for _, reader := range readers {
-		bytes, err := io.ReadAll(reader)
-		if err != nil {
-			return files, err
-		}
+		var allBytes []byte
 
-		if verbose {
-			if len(bytes) > 0 {
+		for _, reader := range readers {
```
tiziano88, line 486: Is it intentional to have two nested loops over the same range?

jul-sh: unintentional, thx, fixed

internal/builders/docker/pkg/builder.go
```diff
@@ -481,25 +481,36 @@ func (c *GitClient) checkoutGitCommit() error {
...
+			fmt.Print("\n\n>>>>>>>>>>>>>> output from command <<<<<<<<<<<<<<\n")
 		}
 
-		if verbose {
-			if len(bytes) > 0 {
-				fmt.Print("\n\n>>>>>>>>>>>>>> output from command <<<<<<<<<<<<<<\n")
-				fmt.Printf("%s", bytes)
-				fmt.Print("=================================================\n\n\n")
+		scanner := bufio.NewScanner(reader)
+		for scanner.Scan() {
```
jblebrun, line 490: I think this will still try to process each provided Reader to completion before moving to the next. So we'd at least get the output from `stdout`, but if the process is terminated, we'd still never get to see `stderr`. 

One possible solution is to launch a goroutine to process each Reader concurrently. My golang is very out of date, so I'm not sure about the best practices around doing something like this these days.

internal/builders/docker/pkg/builder.go
laurentsimon: I guess my question is: what's wrong with the current printf? Is it that it may panic of there's too much data to print at once? Something else?

jblebrun: There are two downsides that I can think of:

1) (Less of an issue, just a "nice to have") - You can't see any output until the job completes. It's nice to be able to see progress as the task is running to ensure that things are progressing as expected.

2) (The reason we're exploring this) - If something happens to cause the task to hang or take a long time, the job is aborted. Because it's killed, there's never a chance for stdout or stderr to be printed to the output area. This makes it very challenging to understand what happened during the build to cause this. We have a task that's hanging and triggering a timeout. The behavior is only reproducible in this context, but we haven't been able to find a way to get any of the output.

internal/builders/docker/pkg/builder.go
jblebrun: (Updated this to try to concurrently print/process each provided file).

internal/builders/docker/pkg/builder.go
jblebrun: Update on this: We've been running our problematic job against this patched builder. The output streams along as the build progresses, which is really nice!

Strangely, the build just succeeds without hanging when we use this patch. I can't quite pin down why that would be the case.

internal/builders/docker/pkg/builder.go
```diff
@@ -476,34 +478,74 @@ func (c *GitClient) checkoutGitCommit() error {
...
 
-		if verbose {
-			if len(bytes) > 0 {
-				fmt.Print("\n\n>>>>>>>>>>>>>> output from command <<<<<<<<<<<<<<\n")
-				fmt.Printf("%s", bytes)
-				fmt.Print("=================================================\n\n\n")
-			}
-		}
+	// Close the channel once all goroutines have finished.
+	go func() {
```
laurentsimon, line 533: I think this should not be in a go routine?

jul-sh:
cc @timonvo also, I mentioned this PR in our sync earlier. It unblocks signed provenances for some of the remaining Oak components.

internal/builders/docker/pkg/builder.go
```diff
@@ -476,34 +478,74 @@ func (c *GitClient) checkoutGitCommit() error {
...
 }
 
+type tempFileResult struct {
+	File *os.File
+	Err  error
+}
+
+// A helper function used by saveToTempFile to process one individual file.
+func saveOneTempFile(verbose bool, reader io.Reader, fileChannel chan tempFileResult, printChannel chan string) {
+	var allBytes []byte
```
laurentsimon, line 488: I don't know if it's good practice or not, but you could:
```golang
defer fileChannel.Close()
defer printChannel.Close()
```

internal/builders/docker/pkg/builder.go
jblebrun: Either this needs to be in a goroutine, of lines 539-541 below need to be a in a goroutine, so that printing can occur while the reader processing goroutines are running.

internal/builders/docker/pkg/builder.go
jblebrun: We shouldn't close the channels here, because they are passed to more than one goroutine running this function. They should be only closed after the WaitGroup representing all active coroutines moves to the "Done" state.

internal/builders/docker/pkg/builder.go
```diff
@@ -476,34 +478,74 @@ func (c *GitClient) checkoutGitCommit() error {
...
+		}
+	}
+
+	tmpfile, err := os.CreateTemp("", "log-*.txt")
+	if err != nil {
+		fileChannel <- tempFileResult{Err: err}
+		return
+	}
+	if _, err := tmpfile.Write(allBytes); err != nil {
+		tmpfile.Close()
```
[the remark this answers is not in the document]
jblebrun: Yes, great idea!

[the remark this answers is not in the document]
jblebrun: I also fixed the issue that two values would be sent on the fileChannel in the error case!

internal/builders/docker/pkg/builder.go
laurentsimon: ho I see it now. That makes sense. Maybe add a comment

laurentsimon (approved):

internal/builders/docker/pkg/builder.go
jblebrun: I added a few other comments as well. Let me know if it's too much, or if you have other questions that we can capture in some comments here.

jblebrun:
I'm going to so some final testing with our pipeline after the changes, just to make sure I haven't made any mistakes with the changes.

internal/builders/docker/pkg/builder.go
jul-sh: Wouldn't it also be okay to just move this at the end of the parent function?

jblebrun:
@laurentsimon Thanks for the review! BTW, it looks like the workflows need to be approved again.

internal/builders/docker/pkg/builder.go
jblebrun: We need this to be running concurrently before we range over the channels below, because those range loops are terminated by the channel closing.

jul-sh: ah yes. My idea would've been to use the same routine, but only start waiting after those loops conclude. But you're right, in this case the loops would never conclude. Thanks for explaining.

laurentsimon (approved):

jul-sh:
@laurentsimon fixed the linter failure caught in the checks that ran after approval. Sorry, but I think you need to approve again.

jblebrun:
@laurentsimon Sorry, one more question!

We were trying to use a version including this fix... referencing a specific digest didn't work (detect-env fails), but even using a named ref like `main` results in a failure (the build completes but the artifact can't be found later).

After some digging I found https://github.com/slsa-framework/slsa-github-generator/blob/main/README.md#referencing-slsa-builders-and-generators which seems to indicate that everything only works if we use a specific tag formed like `vX.Y.Z`?

Is our only solution to wait for a properly tagged version to include this fix? Is there any other workaround?

laurentsimon:
You're correct, we need to release with the patch

jblebrun:
> You're correct, we need to release with the patch

Is there any specific time table that the org prefers to follow for releases? We could follow the process in RELEASES.md to get the next release moving, if that's invited.

laurentsimon:
@ramonpetgrave64 is looking into releasing. Unfortunately one needs to be a maintainer to go thru process [REALEASE](https://github.com/slsa-framework/slsa-github-generator/blob/main/RELEASE.md) (need to have certain permissions, etc)

ramonpetgrave64:
We have a pre-release that you can start using to build and generate provenances, but verifying won't yet work. https://github.com/slsa-framework/slsa-github-generator/releases/tag/v2.0.0-rc.0

ramonpetgrave64:
We've released v2.0.0 today.
https://github.com/slsa-framework/slsa-github-generator/releases/tag/v2.0.0

jul-sh:
> We've released v2.0.0 today. https://github.com/slsa-framework/slsa-github-generator/releases/tag/v2.0.0

thank you so much for the quick release! we updated our repo to v2 and it's working beautifully. cc @tiziano88 @jblebrun @tiziano88

PR #3572, opened 2024-04-16
fix: use @sigstore/cli in e2e.sign-attestations.schedule.yml

ramonpetgrave64:
# Summary

Addresses https://github.com/slsa-framework/slsa-github-generator/issues/3002

Fixes the .github/workflows/e2e.sign-attestations.schedule.yml workflow.
sigstore-js now has its cli tools in a separate package, to be installed with `install -g @sigstore/cli`.

## Testing Process

Invoked the workflow from my personal fork

 - https://github.com/ramonpetgrave64/slsa-github-generator/actions/runs/8757196289/job/24035331070#step:6:11

```
Verification succeeded
Verification succeeded
```

We can't add this to a pre-submit, because it requires token permissions that are not available to forks' PR runs.

## Checklist

### DIFF
--- a/.github/workflows/e2e.sign-attestations.schedule.yml
+++ b/.github/workflows/e2e.sign-attestations.schedule.yml
@@ -42,15 +42,15 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@5e21ff4d9bc1a8cf6de233a3057d20ec6b3fb69d # v3
         with:
-          node-version: 16
+          node-version: 20
       - name: install sigstore-js
-        run: npm install -g @sigstore/cli@^0.8.0
+        run: npm ci
       # TODO: Add subject verification. This only verifies the signature on the DSSE envelope.
       - name: run verify
         run: |
           set -euo pipefail
           for file in outputs/*; do
-            sigstore verify "$file"
+            npx sigstore verify "$file"
           done
 
   if-succeed:
--- a/.github/workflows/e2e.sign-attestations.schedule.yml
+++ b/.github/workflows/e2e.sign-attestations.schedule.yml
@@ -44,13 +44,13 @@ jobs:
         with:
           node-version: 16
       - name: install sigstore-js
-        run: npm ci
+        run: npm install -g @sigstore/cli
       # TODO: Add subject verification. This only verifies the signature on the DSSE envelope.
       - name: run verify
         run: |
           set -euo pipefail
           for file in outputs/*; do
-            ./node_modules/.bin/sigstore verify "$file"
+            sigstore verify "$file"
           done
 
   if-succeed:
--- a/.github/workflows/e2e.sign-attestations.schedule.yml
+++ b/.github/workflows/e2e.sign-attestations.schedule.yml
@@ -44,7 +44,7 @@ jobs:
         with:
           node-version: 16
       - name: install sigstore-js
-        run: npm install -g @sigstore/cli
+        run: npm install -g @sigstore/cli@^0.8.0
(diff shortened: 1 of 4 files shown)


ramonpetgrave64:
@laurentsimon @haydentherapper

.github/workflows/e2e.sign-attestations.schedule.yml
```diff
@@ -44,13 +44,13 @@ jobs:
         with:
           node-version: 16
       - name: install sigstore-js
-        run: npm ci
+        run: npm install -g @sigstore/cli@^0.8.0
```
haydentherapper, line 47: Does npm have a concept of `latest`? Then you don't have to worry about updating to the latest sigstore-js CLI

ramonpetgrave64: Yes we could always install the latest
```
npm install -g @sigstore/cli
```
, though we generally like to pin versions when we can. But I guess if the signer's version is already pinned, it could be okay to leave the verifier unpinned. @ianlewis wdyk?

laurentsimon: Let's keep pinning, this is useful for determinism.  A better way to do to use a lock file, so that dependabot / renovatebot sends us PR. renovatebot supports custom regex to let it know it should interpret this `^0.8.0` as a version to update. I think it's easier to use a lock file to pin the dep.

ianlewis: Any reason why we don't pin in the `package.json` in the project root? I managed the versions for dev dependencies there so that renovate could send PRs to update them etc.

https://github.com/slsa-framework/slsa-github-generator/blob/main/package.json

.github/workflows/e2e.sign-attestations.schedule.yml
```diff
@@ -44,13 +44,13 @@ jobs:
...
       # TODO: Add subject verification. This only verifies the signature on the DSSE envelope.
       - name: run verify
         run: |
           set -euo pipefail
           for file in outputs/*; do
-            ./node_modules/.bin/sigstore verify "$file"
+            sigstore verify "$file"
```
ianlewis, line 53: I think using `npm ci` above and then just running via npx would be better.
```suggestion
            npx sigstore verify "$file"
```

This is essentially what I was doing before but running via `npx` is probably nicer than referencing the executable directly in `node_modules/.bin`.

ramonpetgrave64: Thanks, that's a better place for this.

.github/workflows/e2e.sign-attestations.schedule.yml
ramonpetgrave64: Thanks, that's a better place for this.

laurentsimon (approved):

package.json
```diff
@@ -7,6 +7,7 @@
     "markdown-toc": "markdown-toc --bullets='-' -i"
   },
   "devDependencies": {
+    "@sigstore/cli": "^0.8.0",
```
ianlewis, line 10: Your call but I would just pin to a specific version since renovate will update for us anyway. Provides a bit more consistent behavior as `npm install` won't try to use later versions and update the `package-lock.json` (and dirty a checkout). I don't really think we want to update deps outside of renovate PRs anyway.

ianlewis (approved):

package.json
ramonpetgrave64: I think you're right. changed.

Issue #2951, opened 2023-11-07
[feature] workflow for publishing source code archives as release assets


junyer:
**Is your feature request related to a problem? Please describe.**
Bazel [recommends](https://blog.bazel.build/2023/02/15/github-archive-checksum.html) publishing source code archives as release assets – and Bazel Central Registry [verifies](https://github.com/bazelbuild/bazel-central-registry/blob/main/tools/verify_stable_archives.py) stability by checking for `…/releases/download/…` in GitHub URLs. Using `gh release download` and `gh release upload`, GitHub Actions can automate this trivially, but OpenSSF [punishes](https://github.com/ossf/scorecard/blob/main/docs/checks.md#signed-releases) projects whose release assets lack signature and provenance.

**Describe the solution you'd like**
SLSA should provide a workflow for publishing source code archives as release assets **_with signature and provenance_**. Ideally, any project's `release` workflow could include a job specifying **_only_** `permissions` and `uses` keys and get `.zip`, `.zip.intoto.jsonl`, `.tar.gz` and `.tar.gz.intoto.jsonl` files attached to the release.

**Describe alternatives you've considered**
Letting N different projects implement this themselves in approximately N different ways. ;)

**Additional context**
N/A

laurentsimon:
Hi, thanks for the issue. You can achieve this today using the generator https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/README.md and setting the `upload-asset: true`. Can you confirm this works for your use case?

junyer:
I'm confident that it would, but the idea here is to spare every project the `sha256sum … | base64 -w0` and `base64-subjects` dance and, moreover, to ensure that the `gh release download` and `gh release upload` dance is done by a trusted reusable workflow. It is – or it **_ought_** to be! – a common operation with approximately zero room for variations, so it should be made as convenient for every project as reasonably possible. Does that clarify the intention behind the request? :)

laurentsimon:
I understand the `sha256sum … | base64 -w0`.
I don't fully understand the `gh release download` and `gh release upload`. The generator I linked to _does_ do the release upload by setting the `upload-asset: true`. Can you clarify this point?

junyer:
[...]

PR #3587, opened 2024-04-24
feat: source archive builder

laurentsimon:
# Summary

Fixes https://github.com/slsa-framework/slsa-github-generator/issues/2951

## Testing Process

Locally. Need to merge to add e2e tests.

## Checklist

- [X] Review the contributing [guidelines](./../CONTRIBUTING.md)
- [X] Add a reference to related issues in the PR description.
- [X] Add unit tests if applicable.


internal/builders/archive/README.md
```diff
@@ -0,0 +1 @@
+TODO
```
laurentsimon, line 1: will do this later this week

.github/workflows/builder_source-archive_slsa3.yml
```diff
@@ -0,0 +1,71 @@
...
+    needs: [slsa-setup]
+    permissions:
+      id-token: write # For signing.
+      contents: read # For repo checkout of private repos.
+      actions: read # For getting workflow run on private repos.
+    uses: slsa-framework/slsa-github-generator/.github/workflows/delegator_generic_slsa3.yml@main
+    with:
+      slsa-token: ${{ needs.slsa-setup.outputs.slsa-token }}
+
+  # TODO(): upload attestation as release assets using needs.slsa-run.outputs.attestations-download-name and attestations-download-sha256
```
laurentsimon, line 71: Will add this later this week

.github/workflows/builder_source-archive_slsa3.yml
```diff
@@ -0,0 +1,72 @@
...
+          slsa-runner-label: "ubuntu-latest"
+          slsa-build-action-path: "./internal/builders/archive"
+          slsa-workflow-inputs: ${{ toJson(inputs) }}
+
+  slsa-run:
+    needs: [slsa-setup]
+    permissions:
+      id-token: write # For signing.
+      contents: write # To release assets.
+      packages: write # To publish to GitHub packages.
```
laurentsimon, line 66: is `packages:write` acceptable? It's not used by the builder but used by BYOB framework integration. 
@junyer wdut?

kpk47 (changes requested):
I meant to check "request changes" on that last round of comments.

.github/workflows/builder_source-archive_slsa3.yml
laurentsimon: Can't make them optional because reusable workflows don't support dynamic permissions. But we have 2 BYOB integrations, on of which requires only `token:write`, so I'll probably update to use the low perm BYOB integration

.github/workflows/builder_source-archive_slsa3.yml
```diff
@@ -0,0 +1,72 @@
...
+          Example: "zip tar.gz"
+        required: true
+        type: string
+      upload-assets:
+        description: >
+          If true, provenance is uploaded to a GitHub release for new tags.
+          The only value supported is 'true'. The workflow will fail if set to 'false'.
+        required: false
+        type: boolean
+        default: true
```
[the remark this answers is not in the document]
laurentsimon: I'm thinking in terms of backward compatibility, eg if we ever want to support non-release archive generation. This would require folks being able to set whether they want to upload  or not, for example. By default it could be that it does not upload, like in other workflows we have. Let me think a bit more, I'm less convinced than I was yesterday :)

internal/builders/archive/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+*/
+
+import * as github from "@actions/github";
+import * as core from "@actions/core";
+import * as inputs from "./inputs";
+import * as attestation from "./attestation";
+import * as crypto from "crypto";
+import * as tscommon from "tscommon";
+
+async function run(): Promise<void> {
```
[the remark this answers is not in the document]
laurentsimon: 1 and 2 only. +1 that the function is a bit long and should be split and documented

internal/builders/archive/src/index.ts
```diff
@@ -0,0 +1,152 @@
...
+      throw new Error("formats empty");
+    }
+    const formats = inputs.parseFormats(f);
+    const apiFormats = inputs.formatsToAPI(formats);
+
+    // Verify that upload-assets is set to true.
+    const uploadAssets = jsonInputs["upload-assets"].trim();
+    if (uploadAssets !== "true") {
+      throw new Error(`expected upload-assets to be true, got ${uploadAssets}`);
+    }
```
ramonpetgrave64, line 47: I'm guessing you mean for this to intentionally fail, but perhaps towards the end of the workflow

ramonpetgrave64, line 47: I'm guessing you mean for this to intentionally fail, but perhaps towards the end of the workflow

Issue #1555, opened 2023-01-19
[feature] Report variables in provenance


laurentsimon:
GitHub recently announced the GA of variables https://github.blog/2023-01-10-introducing-required-workflows-and-configuration-variables-to-github-actions/

They are available *by default* in the reusable workflow. We should report them as they are a source of user inputs.

Thanks @MarkLodato for pointing this out.

laurentsimon:
Thought on milestone for this feature? Shall we target BYOB?

laurentsimon:
That makes sense to me. The top-level workflow need not even be reported for builders since it's not an input and does *not* influence the builder. But because it's still useful for folks who want to enforce a policy on *which* workflow triggered their build, I think it's fine to keep it.

@ianlewis wdut?

ianlewis:
Ok, I think that makes sense. So from a "reproducibility" standpoint, for builders we only really care about what happens in the context of the reusable workflow. For generators we do care because presumably it's used and can affect the build.

ianlewis:
> Though since they aren't environment variables and would need to be passed to the script explicitly we could just not support that. Some builders might ultimately need that support though depending on user needs.

I think we should record them regardless for BYOB since a TRW could technically make use of them.

PR #3622, opened 2024-05-13
feat: Record the vars context.

ianlewis:
# Summary

Records the GitHub [vars context](https://docs.github.com/en/actions/learn-github-actions/contexts#vars-context) in the SLSA invocation in all generators and builders.

## generic generator, container generator, Go builder

The `vars` context is passed to the "builder" binary as a JSON blob via the `GITHUB_VARS` environment variable. The values are then recorded in the `invocation.parameters.vars` field of the provenance predicate. Masking of inputs or vars is not supported.

## BYOB

The `vars` context is converted to JSON and passed to `setup-generic` by the TRW in the same way that the `inputs` context is added. Vars are then recorded in the SLSA token. Individual `vars` can be masked from the provenance as well via the `slsa-masked-vars` field in the same way as inputs.

`verify-token` reads the vars from the SLSA token and includes them in the final provenance.

Note that changes to the TRW are necessary to record the vars context.

TODO:
- [x] Record vars in generic generator ([actions run](https://github.com/ianlewis/actions-test/actions/runs/9076037341) , [test release](https://github.com/ianlewis/actions-test/releases/tag/v0.1.164))
- [x] Record vars in container generator
- [x] Record vars in Go builder
- [x] Support recording vars in BYOB
- [ ] Record vars in npm builder
- [ ] Record vars in maven builder
- [ ] Record vars in gradle builder
- [ ] Record vars in bazel builder
- [ ] Document changes for TRWs

Updates #1555

## Testing Process

- [ ] Add how to test.

## Checklist

- [x] Update documentation if applicable (added #3627 to track)


ianlewis:
Probably can split this into two PRs. One for the generic generator, container generator, and Go builder. One for BYOB and BYOB workflows. Each PR wouldn't really depend on the other.

ianlewis:
Requiring changes to TRW inputs isn't ideal. Currently we need the TRW to pass the vars context to setup-generic because JavaScript actions don't have access to the vars context like they do for inputs (inputs are set to environment variables but vars aren't).

I wonder if I could create a wrap the current action in a composite action to get access to the vars context without needing it to be passed in an input.

ianlewis:
> I wonder if I could create a wrap the current action in a composite action to get access to the vars context without needing it to be passed in an input.

Looks like this won't work since the `vars` context isn't available to composite actions. I guess it's only available from workflows.

ianlewis:
Breaking up this PR into multiple PRs. First of which is #3633 for the generic generator, container generator, and Go builder.

ianlewis:
Recording vars for BYOB implemented in #3636. Closing this PR.

Issue #3627, opened 2024-05-14
[docs] Add inputs to provenance examples


ianlewis:
Current provenance examples (e.g. [generic generator](https://github.com/slsa-framework/slsa-github-generator/tree/main/internal/builders/generic#provenance-example)) do not include inputs. We should have an example that includes inputs (and vars - #1555).

PR #3633, opened 2024-05-14
feat: Record vars in SLSA generators

ianlewis:
# Summary

Records the GitHub [vars context](https://docs.github.com/en/actions/learn-github-actions/contexts#vars-context) in the SLSA invocation in the generic generator and container generator.

The `vars` context is passed to the "builder" binary as a JSON blob via the `GITHUB_VARS` environment variable. The values are then recorded in the `invocation.parameters.vars` field of the provenance predicate. Masking of inputs or vars is not supported.

Note that the `vars` context is set to the variables for the *repository that initiated the GitHub Actions run* and not the reusable workflow's repository.

Updates #1555

## Testing Process

- Set a variable on the test repo
  - Go to Settings -> Secrets & variables -> Actions and then click on the Variables tab.
  - Set some test variables.
- Generate provenance as normal using the generic generator or container generator
  - Note that the `compile-generator: true` input must be set.
- Examine the `invocation.parameters.vars` field in the resulting provenance. The vars should be recorded there.

## Checklist

- [x] Update documentation if applicable (added #3627 to track)


.github/workflows/builder_go_slsa3.yml
```diff
@@ -343,6 +343,7 @@ jobs:
           UNTRUSTED_ENV: "${{ needs.build-dry.outputs.go-env }}"
           UNTRUSTED_WORKING_DIR: "${{ needs.build-dry.outputs.go-working-dir }}"
           GITHUB_CONTEXT: "${{ toJSON(github) }}"
+          VARS_CONTEXT: "${{ toJSON(vars) }}"
```
ianlewis, line 346: I forgot for the Go builder, vars can't affect the build so we probably shouldn't record them in this case.

slsa/buildtype.go
```diff
@@ -53,39 +53,45 @@ type BuildType interface {
 
 // GithubActionsBuild is a basic build type for builders running in GitHub Actions.
 type GithubActionsBuild struct {
-	Context github.WorkflowContext
-	Clients ClientProvider
-	subject []intoto.Subject
+	Context  github.WorkflowContext
```
ramonpetgrave64, line 56: We should maybe rename this to `WorkflowContext`

ianlewis: I mostly agree and I don't think anyone's using it but since it's a public interface I'd rather not change it unless there's a better reason.

ramonpetgrave64: Then I think we should have comments here to explain each the properties

ramonpetgrave64 (approved):

slsa/buildtype.go
ianlewis: Done.

PR #3636, opened 2024-05-15
feat: Record vars in BYOB workflows

ianlewis:
# Summary

The `vars` context is converted to JSON and passed to `setup-generic` by the TRW in the same way that the `inputs` context is added. Vars are then recorded in the SLSA token. Individual `vars` can be masked from the provenance as well via the `slsa-masked-vars` field in the same way as inputs.

`verify-token` reads the vars from the SLSA token and includes them in the final provenance.

Note that changes to the TRW are necessary to record the vars context.

TODO:
- [x] Support recording vars in BYOB
- [ ] Record vars in npm builder
- [ ] Record vars in maven builder
- [ ] Record vars in gradle builder
- [ ] Record vars in bazel builder
- [ ] Document changes for TRWs

Updates #1555

## Testing Process

- [ ] Add how to test.

## Checklist

- [ ] Review the contributing [guidelines](https://github.com/slsa-framework/slsa-github-generator/blob/main/CONTRIBUTING.md)
- [ ] Update documentation if applicable (added #3627 to track)
- [ ] Add changes to the [CHANGELOG](https://github.com/slsa-framework/slsa-github-generator/blob/main/CHANGELOG.md) if applicable.


ianlewis:
> * [x] Record vars in Node.js builder
> * [x]  Record vars in maven builder
> * [x]  Record vars in gradle builder
> * [x]  Record vars in bazel builder

Though, I think BYOB itself should support setting the vars we perhaps shouldn't set them for these builders since they don't need to be used in the TRW.

ianlewis:
I'm kind of thinking we don't really care about this with BYOB for now since for builders the caller can just require users to pass vars as TRW inputs.

We will need this to support generators but we don't currently support generators anyway so it isn't urgent.

Issue #3750, opened 2024-07-29
[feature] Output provenance as Sigstore bundle format


haydentherapper:
**Is your feature request related to a problem? Please describe.**

There is no support for offline verification currently. While a certificate is persisted in the DSSE signature, the Rekor entry proof is not, meaning the SLSA verifier must do an online lookup for the entry in Rekor.

Additionally, since all Sigstore clients now support the bundle format, this would allow SLSA clients to directly use Sigstore libraries to handle verification of the attestation signature (reducing code and simplifying maintenance), with the SLSA client handling verification of the attestation contents.

**Describe the solution you'd like**

Sigstore has defined a [bundle format](https://github.com/sigstore/protobuf-specs/blob/main/protos/sigstore_bundle.proto) to store an attestation or signature along with verification metadata (certificate, signed timestamps, and Rekor log proofs). The generator should output a `.sigstore.json` bundle (the extension matches what other Sigstore clients output).

This was discussed in https://github.com/slsa-framework/slsa-github-generator/issues/716, but this issue was filed before the bundle format was created. This issue tracked persisting Rekor's inclusion proof as its own file.

**Describe alternatives you've considered**

An alternative would be to have a divergent format or save verification metadata as "detached" files, which complicates storage of that metadata for no benefit.

ramonpetgrave64:
The "BYOB" (bazel, maven, gradle, nodejs) builders will use the "delegator" workflows, which will use our `.github/actions/sign-attestations` action to produce Sigstore bundles.

 - https://github.com/slsa-framework/slsa-github-generator/blob/afa0f38ec2bbdf5230c68444c7e964bb19e2b64b/.github/workflows/builder_maven_slsa3.yml#L84
 - https://github.com/slsa-framework/slsa-github-generator/blob/afa0f38ec2bbdf5230c68444c7e964bb19e2b64b/.github/workflows/delegator_generic_slsa3.yml#L278
 - https://github.com/slsa-framework/slsa-github-generator/blob/afa0f38ec2bbdf5230c68444c7e964bb19e2b64b/.github/actions/sign-attestations/src/main.ts#L44
 - Example maven e2e workflow run - https://github.com/slsa-framework/example-package/actions/runs/10135785683

[...]

PR #3777, opened 2024-08-14
feat: produce sigstore Bundles for generic generator and go builder workflows

ramonpetgrave64:
# Summary

fixes https://github.com/slsa-framework/slsa-github-generator/issues/3750

pending https://github.com/slsa-framework/slsa-verifier/pull/799

Changes the internal go code to produce Sigstore Bundles, instead of only signed DSSE envelopes. This means that the generic generator and go builder workflows now produce Sigstore Bundles, just like the other BYOB-type workflows.

## Testing Process

Testing done on a previous commit with a test workflow. It's using a slightly modified slsa-verifier that respects sls-aw workflows from non-main branches.
 - https://github.com/slsa-framework/slsa-github-generator/actions/runs/10425271660

## Followup

[ ] Produce the provenance in v1 format, rather than the current v0.2 format.
[ ] fix initialism of `[build]invocationID` to `[build]invocationId` https://github.com/slsa-framework/slsa-github-generator/issues/3876

## Checklist


github/oidc.go
```diff
@@ -39,6 +39,9 @@ const (
 
 // OIDCToken represents the contents of a GitHub OIDC JWT token.
 type OIDCToken struct {
+	// Expiry is the expiration date of the token.
```
ramonpetgrave64, line 42: lint: reordering fields by type.

ramonpetgrave64:
@ianlewis @laurentsimon

internal/builders/docker/pkg/builder_test.go
```diff
@@ -53,7 +53,7 @@ func Test_CreateBuildDefinition(t *testing.T) {
 	}
 
 	if diff := cmp.Diff(got, want); diff != "" {
-		t.Errorf(diff)
+		t.Error(diff)
```
ramonpetgrave64, line 56: lint: change `t.Errorf` to `t.Error` when not actually using any string formatting.

signing/sigstore/bundle.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// get the oidc token.
+	oidcClient, err := github.NewOIDCClient()
+	if err != nil {
+		return nil, err
+	}
+	TokenStruct, err := oidcClient.Token(ctx, []string{"sigstore"})
```
haydentherapper, line 76: nit: `tokenStruct`

signing/sigstore/bundle.go
```diff
@@ -0,0 +1,146 @@
...
+	innerBundle, err := sigstoreSign.Bundle(content, keypair, *bundleOpts)
+	if err != nil {
+		return nil, err
+	}
+
+	// print the logIndex.
+	// Bundle will have already verified that the TLog entries are signed.
+	logIndex := innerBundle.GetVerificationMaterial().GetTlogEntries()[0].GetLogIndex()
+	fmt.Printf("Signed attestation is in rekor with UUID %d.\n", logIndex)
+	fmt.Printf("You could use rekor-cli to view the log entry details:\n\n"+
```
haydentherapper, line 98: Minor thing, but I would avoid printing the info about how to look up the entry given these look ups aren't verifying inclusion. In Cosign we've always struggled with what is useful output, and I don't have a good answer there - maybe a list of steps that the client took (requested token, fetched cert, uploaded entry to log, etc)?

signing/sigstore/bundle.go
haydentherapper, line 1: Could we add any tests for this? Possibly e2e tests with a similar setup like Colleen added in Cosign? https://github.com/sigstore/cosign/blob/main/test/README.md

haydentherapper (commented):
This is really great!

signing/sigstore/bundle.go
ramonpetgrave64: It's more of a UX thing, since I'm deleting another print statement that showed the log UUID. 

I now see I need to update that text to say Log Index, not UUID.

signing/sigstore/bundle.go
ramonpetgrave64: There are e2e tests in slsa-framework/example-package that will produce and verify. We can't have the e2e tests in PRs because PRs don't have the needed token permissions to sign.
  - https://github.com/slsa-framework/example-package/blob/b771344a4ceeb4224c0e683910fc60360f91827c/.github/workflows/e2e.generic.release.main.default.slsa3.yml#L84

signing/sigstore/bundle.go
ramonpetgrave64: fixed.

signing/sigstore/bundle.go
ramonpetgrave64: fixed.

signing/sigstore/bundle.go
haydentherapper: Can you use the identity token workflow we set up for conformance testing that publishes a valid ID token? Here's an example of gitsign using it - https://github.com/sigstore/gitsign/pull/549/files

haydentherapper (approved):

PR #3940, opened 2024-10-11
chore: go1.23.1 golangci lint 1.6.1

ramonpetgrave64:
# Summary

Followup to https://github.com/slsa-framework/slsa-verifier/pull/810

- upgrades to go 1.23.1 and golangci-lint 1.6.1
- makes necessary changes for the linter
  - as of go 1.22, no need to make a copy of the loop variable
  - change non-string use of `fmt.Errorf(err)` to `fmt.Error(err)`

## Testing Process

unit tests continue to pass, as well as the linter

## Checklist

### DIFF
--- a/.github/workflows/pre-submit.lint.yml
+++ b/.github/workflows/pre-submit.lint.yml
@@ -87,8 +87,8 @@ jobs:
         with:
           go-version-file: "go.mod"
       - env:
-          GOLANGCI_LINT_VERSION: "1.57.0"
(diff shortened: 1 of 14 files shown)


ramonpetgrave64:
@jku @loosebazooka

jku (approved):
Looks good to me.

The golangci-lint job looks like it should be replaced with `golangci/golangci-lint-action`  but i'm not very familiar with this (and it's not really related to the chore here) so I'm not strongly suggesting it.

ramonpetgrave64:
@jku @loosebazooka, A while ago @ianlewis and @laurentsimon and I decided it was better to keep it this way, since we get to verify the golangci-lint binary's hash. Their official action I believe still does not verify the binary of the hash, even if downloading directly from their GitHub repo's releases.

PR #4120, opened 2025-02-19
chore: update refs to v2.1.0-rc.1

ramonpetgrave64:
#label:release v2.1.0-rc.1

# Summary

Retrying the release, under a new tag, not rc.0.

The [original release steps](https://github.com/slsa-framework/slsa-github-generator/blob/v2.0.0/RELEASE.md#verify-rc-version-references) required doing a force push to update the tag in order to trigger workflow to produce the binaries as release artifacts. But now, repo settings make tags immutable, and I don't have permission to change the settings, so we'll try something different:

1. Use the changes in this PR's branch (not main) to cut a new pre-release under a new tag. This should produce the binaries and associate it with the release.
    - https://github.com/slsa-framework/slsa-github-generator/releases/tag/v2.1.0-rc.1
1. Merge this PR, after doing step 1.
1. The release and main should now be at the same commit sha

## Testing Process

presubmits should pass

## Checklist

### DIFF
--- a/.github/actions/generate-builder/action.yml
+++ b/.github/actions/generate-builder/action.yml
@@ -62,7 +62,7 @@ runs:
   using: "composite"
   steps:
     - name: Checkout builder repository
-      uses: slsa-framework/slsa-github-generator/.github/actions/secure-builder-checkout@v2.1.0-rc.0
+      uses: slsa-framework/slsa-github-generator/.github/actions/secure-builder-checkout@v2.1.0-rc.1
       with:
         repository: ${{ inputs.repository }}
         ref: ${{ inputs.ref }}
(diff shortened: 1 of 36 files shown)


haydentherapper (approved):

ramonpetgrave64:
The effect was not as expected. Main has a different commit hash

ramonpetgrave64:
The only way may be to
1. create a new rc.2, merge it wihout the tests passing. Most tests should fail because the tag would not have existed yet.
2. create the new tag from main

haydentherapper:
I'm not following what is broken. Can we just tag a release from main and test against that?

ramonpetgrave64:
> I'm not following what is broken. Can we just tag a release from main and test against that?

We need a tagged release, where the released code's workflows also references workflows and actions at the same tag: the the code within v2.1.0-rc.1 must reference call v2.1.0-rc.1, not main. Under normal situations, you can't do this without overwriting a tag. But we don't have permissions to do that anymore.

PR #4441, opened 2025-10-13
fix: provide path to go.sum for caching

HastD:
This prevents warnings about a missing dependency file.


HastD:
@ianlewis Just following up on this—should I go ahead and change this PR to set `cache: false` to disable caching?

